PR #59, opened 2021-08-30
migrate diagram CSS styling from jss -> emotion

bpowers:
Emotion is [now the default for material-ui](https://next.material-ui.com/getting-started/installation/), and JSS both increases bundle size and needs some hacks.


src/app/Home.tsx
```diff
@@ -17,262 +18,241 @@ import MenuItem from '@material-ui/core/MenuItem';
...
+        anchorEl: event.currentTarget,
+      });
+    };
+
+    handleProjectCreated = (project: Project) => {
+      window.location.pathname = '/' + project.id;
+    };
+
+    getGridListCols = () => {
+      // TODO: this should be 1 on small screens, but useMediaQuery doesn't
```
bpowers, line 93: is there a way to fix this with 5?

PR #69, opened 2022-05-22
Publically -> Publicly

evanp:
https://en.wiktionary.org/wiki/publically

"This spelling is omitted from many dictionaries. The suffix -ally is normally used in this context (added onto words that end in -ic), but public is an exception, and publicly is generally considered more correct."


bpowers:
ah great find, thanks!

PR #70, opened 2022-07-30
support for index expressions and `SUM` builtin

bpowers:
two major pieces here: expanding the lexer + parser to support the full array indexing syntax.  Second is supporting the array version of `min` and `max` builtins, along with the array-only `rank`, `size`, `stddev`, and `sum`.


bpowers:
this was done in a different way

PR #107, opened 2025-08-25
Implement public projects

### LINKED ISSUE #84: publicly visible does not work?
I just created a new project `this_is_a_public_project` under `gsverhoeven`, based on another project that I had previously exported as STMX file.
I made sure to tick the "publicly visible" button.

However, using my smartphone as a not logged user, and browsing to https://app.simlin.com/gsverhoeven/this_is_a_public_project
the project is not showing, instead Simlin asks for login credentials.

Is it already possible to create publicly visible projects? Or is this still WIP?

Regards
Gertjan

filipesmedeiros:
Hello!

This PR implements the public projects feature, where, if a project is made public, everyone (with or without login) can open it and play with it.

The current state of the PR is *draft*: there are no tests, and, although functionally it sort of works, it uses some hacks (like ignoring auth errors). As of opening the PR, it shows the minimum viable changes to have a barely-working version of the feature.

## Design decision

* What can people do on public models? Should they be able to edit them (of course only locally, with no changes saved)? Or only seeing results — and, in the near future when we have multi-run working, running the model?

### Additional note

I noticed that the architecture around routing/authentication on the client side could be simplified, probably by letting each page take care of its own authentication logic.

Closes #84

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -228,7 +228,16 @@ const InnerApp = styled(
 
     editor = (props: RouteComponentProps<EditorMatchParams>) => {
       const { username, projectName } = props.params;
-      return <HostedWebEditor username={username} projectName={projectName} baseURL={this.getBaseURL()} />;
+      const user = this.state.user;
+      const readOnlyMode = !user || user.id !== username;
+      return (
+        <HostedWebEditor
+          username={username}
+          projectName={projectName}
+          baseURL={this.getBaseURL()}
+          readOnlyMode={readOnlyMode}
+        />
+      );
     };
 
     home = (_props: RouteComponentProps) => {
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -267,12 +267,14 @@ const InnerApp = styled(
         );
       }
 
-      if (!this.state.user) {
-        return <Login disabled={this.state.authUnknown} auth={this.state.auth} />;
-      }
+      if (!/\/.*\/.*/.test(window.location.pathname)) {
+        if (!this.state.user) {
+          return <Login disabled={this.state.authUnknown} auth={this.state.auth} />;
+        }
 
-      if (this.state.isNewUser) {
-        return <NewUser user={defined(this.state.user)} onUsernameChanged={this.handleUsernameChanged} />;
+        if (this.state.isNewUser) {
+          return <NewUser user={defined(this.state.user)} onUsernameChanged={this.handleUsernameChanged} />;
+        }
       }
 
       return (
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -13,7 +13,7 @@ import {
   User as FirebaseUser,
 } from 'firebase/auth';
 
-import { useLocation, Route, RouteComponentProps, Switch } from 'wouter';
+import { useLocation, Route, RouteComponentProps, Switch, Redirect } from 'wouter';
 import { styled } from '@mui/material/styles';
 import { createTheme, ThemeProvider } from '@mui/material/styles';
 import CssBaseline from '@mui/material/CssBaseline';
@@ -230,6 +230,7 @@ const InnerApp = styled(
       const { username, projectName } = props.params;
       const user = this.state.user;
       const readOnlyMode = !user || user.id !== username;
+
       return (
         <HostedWebEditor
           username={username}
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -277,7 +277,11 @@ const InnerApp = styled(
         );
       }
 
-      // if a user is navigating to a model,
+      const urlParams = new URLSearchParams(window.location.search);
+      const projectParam = urlParams.get('project');
+      if (projectParam) return <Redirect to={projectParam} />;
+
+      // if a user is navigating to a project,
       // skip the high level auth check, to enable public models
       if (!/\/.*\/.*/.test(window.location.pathname)) {
         if (!this.state.user) {
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -276,6 +276,8 @@ const InnerApp = styled(
         );
       }
 
+      // if a user is navigating to a model,
+      // skip the high level auth check, to enable public models
       if (!/\/.*\/.*/.test(window.location.pathname)) {
         if (!this.state.user) {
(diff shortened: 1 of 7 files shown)


bpowers:
thanks I will review this today!

src/server/app.ts
```diff
@@ -224,19 +224,30 @@ class App {
 
     this.app.get(
       '/:username/:projectName',
-      userAuthz,
+      async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
bpowers, line 227: I think this route (`GET /:username/:projectName` is the _only_ route using userAuthz.  can we either "fix" that with the right logic, or delete that and just use this lambda?

src/diagram/HostedWebEditor.tsx
```diff
@@ -25,6 +25,7 @@ interface HostedWebEditorProps {
   projectName: string;
   embedded?: boolean;
   baseURL?: string;
+  readOnlyMode?: boolean;
```
bpowers, line 28: should this _also_ make it so the editing tools don't pop up, or equation isn't editable?

bpowers (changes requested):
I had mostly small requests, conceptually looks reasonable!

src/server/app.ts
filipesmedeiros: I will do that! :)

src/app/App.tsx
```diff
@@ -267,12 +267,14 @@ const InnerApp = styled(
         );
       }
 
-      if (!this.state.user) {
-        return <Login disabled={this.state.authUnknown} auth={this.state.auth} />;
-      }
-
-      if (this.state.isNewUser) {
-        return <NewUser user={defined(this.state.user)} onUsernameChanged={this.handleUsernameChanged} />;
+      if (!/\/.*\/.*/.test(window.location.pathname)) {
```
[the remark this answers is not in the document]
filipesmedeiros: Sure! 💯

src/diagram/HostedWebEditor.tsx
filipesmedeiros: It's a design choice we need to make! InsightMaker makes the model _truly_ read only (no edits whatsoever). I think it's a mistake, because you then need an account to fork if you want to change even the tiniest thing.

I believe it's a fantastic design choice to let the "visitor" mess with the model as much as they want, given that nothing will be saved, and that they are not interacting with the server in any way. This also gives a "soft feature" of anonymous/"account-less" modelling (which you might not want ahah).

What do you think?

src/app/App.tsx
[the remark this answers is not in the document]
filipesmedeiros: Done :)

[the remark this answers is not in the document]
bpowers: yeah agree - my comment wasnt' as clear as I meant

src/diagram/HostedWebEditor.tsx
filipesmedeiros: Agreed! I'll give it a check

filipesmedeiros: Done! I think it looks pretty good for now :)

bpowers (approved):
I think the build + public -> build change needs to be fixed before this can be re-deployed to app.simlin.com, but I ALSO think I broke that when I changed the build system a few weeks ago.  So I'm going to merge as is and potentially clean that part of it up in the next few days.  thanks!

src/server/app.ts
bpowers: sorry I wrote a long reply but apparently got distracted and never sent it.

I agree the architecture for development is somewhat convoluted -- the same problem doesn't exist when deployed to app.simlin.com (or anywhere else): 
* `yarn deploy` in the top level of the project does:
* `export NODE_ENV=production && yarn clean && yarn build && yarn workspace @system-dynamics/app deploy`
  * (runs a clean build, comverting all typescript -> JS and building the bundled assets for use by the frontend)
* and deploy in app/ does: `cp -r build/* public && cp -r build-component/static/js/* ./public/static/js/ && cp -r build-component/static/wasm/* ./public/static/wasm/ && rm public ../server/public ../server/default_projects` 
  * (copies the relevant parts of the build into public/ in the root level - it also does 2 additional things: invoking the google app engine deploy command and running deploy-clean to clean up the public/ folder back to what was last committed)
* and then I think by default app engine runs `start`: `node src/server/lib` (starting the compiled JS version of our 

so: when deployed to "production" there is a _single_ server/service running, and it is pretty dumb: it doesn't know how to do typescript compilation, it doesn't know about live-reloading, it doesn't do server-side partial react rendering (for now at least) -- it _just_ serves some of the static assets (app engine is also configured to directly serve those transparently from a CDN in app.yaml, which reduces load/cost), handles a few APIs (login/logout/create model/get model/save model), and what you modified here: serving index.html for known `/:username/:projectname` paths.

I hope that helps: when deployed we do option 1.  I also know that this is somewhat different than how development works, but when I set it up I didn't see an easy better way to do it.  happy if you have ideas to make this simpler/more robust!

filipesmedeiros: Hey! Yeah sorry I should have checked that stuff. But from what you are saying, that actually seems much more like 2. (the same server is serving everything), whereas 1. Would be having two servers running (one to serve the static frontend, just an HTTP server, and another for everything else)

Given that we have option 2. going (I reiterate, as far as I can tell), I would say that setting up a React framework like Next is even easier. That would simplify the dev environment and tooling but, as you stated, not the production environment, because that is incredibly simple already. It's more about how many pieces we need to think about when developing, rather than how many pieces are actually running in the end.

I have started a prototype for Next, just to show you how it would look like. Do you have any interest in that? Again, I keep on mentioning Next simply because it's the most mature, but I am not tied to it, and it only serves as an example for the overall architecture.

filipesmedeiros:
That is very fair. I know making this type of change, even with the best of intentions, can have unintended consequences in the long term that are hard to predict. But what I can assure you is that I do not want to make the deployment harder at all. I want to keep it as easy for you or ideally even simpler, and also, if everything goes well, allow for easier deployment on other platforms as well.

I can commit to making this prototype, and discussing it with you at your pace, and only merging it if you see the benefit it would bring for yourself, the project in the long term, and also system dynamics at large!

PR #128, opened 2025-12-19
engine: add subdimension support for StarRange subscripts

bpowers:
## Summary

This PR adds support for subdimension relationships in the simulation engine, enabling the `*:SubDim` (StarRange) syntax for sparse iteration over subdimension elements. For example, `SUM(z[*:SubA])` now correctly sums only elements A2 and A3 when SubA is defined as a subset of DimA containing those elements. The implementation spans the type system (DimensionsContext), expression lowering (Expr2), compilation (compiler), and runtime (interpreter with ArrayView sparse iteration).

The core abstraction is `SubdimensionRelation`, which maps each subdimension element index to its offset in the parent dimension. A `RelationshipCache` with Mutex-based interior mutability provides O(1) lookups after the first computation, caching both positive matches and negative results (when one dimension is not a subdimension of another).

## Design Decisions and Tradeoffs

**Thread-safe caching with Mutex**: The cache uses `Mutex<HashMap>` rather than `RefCell` to ensure thread-safety if the engine is used from multiple threads. The cache stores `Option<SubdimensionRelation>` to distinguish "not computed yet" from "computed and determined not a subdimension."

**Multiple sparse dimensions via Vec<SparseInfo>**: Rather than supporting only one sparse dimension per subscript, `ArrayView` uses `Vec<SparseInfo>` to handle expressions like `matrix[*:SubRow, *:SubCol]`. The interpreter iterates the Cartesian product of sparse offset lists, translating result indices to parent coordinates.

**Contiguous optimization**: When subdimension offsets happen to be sequential (e.g., SubA=[A2,A3] maps to offsets [1,2]), the compiler can use existing range logic instead of sparse iteration. The `is_contiguous()` and `start_offset()` methods on `SubdimensionRelation` enable this optimization path.

**Named dimensions only (initially)**: Indexed subdimensions are deferred because the datamodel lacks metadata to express which range of parent indices a child dimension maps to. The code returns `None` for indexed dimension relationships with a TODO comment explaining why.

**Parent dimension from context**: The implementation does not include a `find_parent_dimension` API because the parent is always known from the variable's dimension metadata when processing subscripts. This simplifies the API and avoids ambiguity when an element appears in multiple dimensions.

**StarRange focus, dimension-as-subscript deferred**: This PR implements the `*:SubA` syntax parsed as `IndexExpr2::StarRange`. The related `a[SubA]` pattern (using a dimension name directly as a subscript) is parsed differently and deferred to a follow-up PR.

### DIFF
--- a/src/simlin-engine/src/array_tests.rs
+++ b/src/simlin-engine/src/array_tests.rs
@@ -1677,6 +1677,42 @@ mod star_range_subdimension_tests {
         project.assert_interpreter_result("slice", &[12.0, 13.0, 22.0, 23.0, 32.0, 33.0]);
     }
 
+    #[test]
+    fn star_to_subdimension_non_contiguous() {
+        // Test star range with non-contiguous subdimension (exercises sparse iteration)
+        let project = TestProject::new("star_to_subdim_non_contiguous")
+            .named_dimension("DimA", &["A1", "A2", "A3", "A4"])
+            .named_dimension("SubA", &["A1", "A3"]) // Non-contiguous: offsets [0, 2]
+            .array_with_ranges(
+                "values[DimA]",
+                vec![("A1", "10"), ("A2", "20"), ("A3", "30"), ("A4", "40")],
+            )
+            .array_aux("result[SubA]", "values[*:SubA]");
+
+        project.assert_compiles();
+        project.assert_sim_builds();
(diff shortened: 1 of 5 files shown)


bpowers:
@codex review

PR #137, opened 2025-12-30
engine: populate dimension metadata in ByteCodeContext

bpowers:
Previously, the ByteCodeContext dimension tables (dimensions, subdim_relations,
and names) were initialized as empty during compilation. This was a critical
gap that could cause runtime issues with star ranges and dynamic slicing that
rely on dimension metadata.

This commit:
1. Adds dimensions and dimensions_ctx fields to the Module struct to carry
   dimension information from project parsing through to compilation
2. Implements populate_dimension_metadata() in the Compiler which:
   - Interns dimension names and element names into the names table
   - Creates DimensionInfo entries for both named and indexed dimensions
   - Detects subdimension relationships between named dimensions and
     populates SubdimensionRelation entries with parent offsets
3. Implements PartialEq for DimensionsContext (comparing only dimensions,
   not the cache) to satisfy Module's derive requirements
4. Adds comprehensive tests for dimension metadata population including
   named dimensions, indexed dimensions, and subdimension relations

### DIFF
--- a/src/simlin-engine/src/compiler.rs
+++ b/src/simlin-engine/src/compiler.rs
@@ -3367,9 +3367,10 @@ impl<'module> Compiler<'module> {
     /// This populates:
     /// - `names`: interned dimension and element names
     /// - `dimensions`: DimensionInfo for each dimension
-    /// - `subdim_relations`: SubdimensionRelation for each subdimension pair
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #138, opened 2026-01-01
engine: add dynamic range support for array subscripts

bpowers:
Implement support for dynamic ranges in array subscripts where the
range bounds are runtime expressions rather than compile-time constants.
This enables expressions like SUM(arr[start_idx:end_idx]) where
start_idx and end_idx are variables evaluated at runtime.

## Problem

The existing array infrastructure supported static ranges (compile-time
constant bounds) but not dynamic ranges where bounds come from variables.
This blocked real-world models that compute partial sums over variable
array slices, such as cumulative calculations or sliding window analysis.

## Solution

The implementation distinguishes between static and dynamic ranges at
the AST level, keeping the efficient static path while adding runtime
evaluation for dynamic bounds.

### AST Changes (expr3.rs)
- Add IndexExpr3::StaticRange for compile-time constant ranges
- Keep IndexExpr3::Range for dynamic runtime ranges that need evaluation

### Compiler Changes (compiler.rs)
- Add SubscriptIndex enum with Single(Expr) and Range(Expr, Expr) variants
  to represent subscript indices in the lowered IR
- Update lower_index_expr3 to detect constant vs dynamic ranges
- Add walk_expr_as_view support for Expr::Subscript with dynamic indices

### Bytecode Changes (bytecode.rs, vm.rs)
- Add PushVarViewDirect opcode for pushing views with explicit dimension
  sizes (needed when we have bounds but not dim_ids from static analysis)
- Add ViewRangeDynamic opcode for applying runtime-computed ranges to
  views on the view stack
- Implement VM execution for both new opcodes

### Interpreter Changes (interpreter.rs)
- Update iter_array_elements to handle Expr::Subscript with Range indices
- Update find_array_dims to correctly identify preserved dimensions
- Enable reduction operations (SUM, etc.) to work with dynamic subscripts

### Dependency Tracking Fix (variable.rs)
- Fix IdentifierSetVisitor.walk_index to walk Range expressions
- Extract is_dimension_or_element helper for filtering dimension names
  and named dimension elements (e.g., Boston, LA) from dependencies
- Ensures proper topological ordering when range bounds are variables

## Testing

Added two new tests demonstrating the feature:
- sum_with_dynamic_range: 1D array with variable range bounds
- sum_with_dynamic_range_2d: 2D array with mixed static/dynamic indices

All 428 engine tests pass.

### DIFF
--- a/src/simlin-engine/src/array_tests.rs
+++ b/src/simlin-engine/src/array_tests.rs
@@ -2018,6 +2018,22 @@ mod structural_lowering_tests {
     }
 
     #[test]
+    fn size_with_empty_dynamic_range() {
+        // Test SIZE with empty range (start > end) returns 0, not 1
+        let project = TestProject::new("size_empty_range")
+            .indexed_dimension("Index", 10)
+            .array_aux("data[Index]", "Index")
+            .scalar_const("start_idx", 7.0)
+            .scalar_const("end_idx", 3.0) // Reversed: start > end = empty range
+            // SIZE(data[7:3]) should be 0 (empty range)
+            .scalar_aux("result", "SIZE(data[start_idx:end_idx])");
+
+        project.assert_compiles();
+        project.assert_sim_builds();
+        project.assert_scalar_result("result", 0.0);
+    }
+
+    #[test]
     fn sum_with_dim_name_subscript_preserves_other_dimension() {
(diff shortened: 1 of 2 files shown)


bpowers:
@codex review

PR #142, opened 2026-01-02
engine: implement indexed dimension broadcasting support

bpowers:
This implements support for broadcasting arrays with different indexed
dimensions that have the same size. For example, sales[Products] +
costs[Regions] now works when Products and Regions are both indexed
dimensions with matching sizes.

Key changes:

1. Extended unify_dims_with_names in ast/expr2.rs to support:
   - Indexed dimension matching by size (when names don't match)
   - UNION broadcasting: combining dimensions from both arrays when
     neither is a subset of the other (e.g., a[X] + b[Y] → result[X,Y])
   - Named dimensions still require exact name matching (semantic safety)

2. Added is_indexed_dimension method to Expr2Context trait to distinguish
   indexed dimensions from named dimensions during type checking.

3. Added dimension matching infrastructure in compiler.rs:
   - find_target_for_source: name-first then size-based matching for indexed dims
   - DimensionMapping, match_dimensions, broadcast_view: scaffolding for future optimizations

4. Added comprehensive tests:
   - Enabled 6 previously ignored broadcasting tests
   - Added 6 new N-dimensional tests (3D/4D broadcasting)

The implementation correctly distinguishes between indexed dimensions
(which can match by size) and named dimensions (which must match by name).
This prevents semantic errors like adding sales[Cities] + prices[Products]
just because both dimensions happen to have the same size.

### DIFF
--- a/src/simlin-engine/src/ast/expr2.rs
+++ b/src/simlin-engine/src/ast/expr2.rs
@@ -950,7 +950,10 @@ mod tests {
         }
 
         fn is_indexed_dimension(&self, _name: &str) -> bool {
-            // For tests, assume all dimensions are indexed
+            // For tests, treat all dimensions as named (not indexed).
+            // This is more conservative - dimensions must match by name, not size.
+            // Tests that need indexed dimension behavior should use TestProject
+            // which has proper dimension context.
             false
         }
     }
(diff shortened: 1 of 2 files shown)


bpowers:
@codex review

PR #145, opened 2026-01-02
engine: implement SIZE(DimensionName) support

bpowers:
This adds support for passing a bare dimension name to the SIZE builtin
function, which returns the element count of that dimension. This is
the semantic equivalent of Vensim's ELMCOUNT function, which xmutil
converts to SIZE(DimensionName) when translating models from Vensim
to XMILE format.

The implementation detects when SIZE's argument is a dimension name
(not a variable) during the Expr1 -> Expr2 conversion phase and
converts it to a constant expression with the dimension's element count.
This avoids the "dimension in scalar context" error that would otherwise
occur when referencing a dimension name outside of an array context.

Unit tests verify this works for both named dimensions (like "DimA" with
elements ["A1", "A2", "A3"]) and indexed dimensions (like "Items(5)").
The elmcount SDEverywhere test model is now enabled as an integration
test, verifying end-to-end compatibility with Vensim models that use
ELMCOUNT.

### DIFF
--- a/src/simlin-engine/src/ast/expr2.rs
+++ b/src/simlin-engine/src/ast/expr2.rs
@@ -630,8 +630,15 @@ impl Expr2 {
                         .transpose()?,
                     ),
                     Size(e) => {
-                        // Special case: SIZE(DimensionName) returns the element count of the dimension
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #146, opened 2026-01-02
engine: support subscripted lookup tables

bpowers:
## Summary
- Add support for `LOOKUP(c[DimA], i)` where `c` is an arrayed variable with per-element graphical functions
- Change `BuiltinFn::Lookup` signature to accept any expression as the first argument
- Implement unified table model where scalar tables are size-1 arrays

### Phase 1: Per-Element GF Parsing
- Add `gf` field to `VarElement` in XMILE parsing
- Extend `Equation::Arrayed` to carry per-element graphical functions as 4-tuples

### Phase 2: Per-Element Table Storage
- Change `Variable::Var.tables` from `Option<Table>` to `Vec<Table>`
- Update `parse_var` to build per-element tables from `Equation::Arrayed`

### Phase 3: Lookup Signature Change
- Change `BuiltinFn::Lookup` from `(String, Box<Expr>, Loc)` to `(Box<Expr>, Box<Expr>, Loc)`
- Update all AST transformation phases and downstream consumers

### Phase 4 & 5: Compiler and Runtime
- Pre-populate `graphical_functions` with all tables at compilation start
- Track `table_base_ids` mapping table variables to their base index
- Update `Lookup` opcode: stack expects `[element_offset, lookup_index]`
- Handle both `StaticSubscript` and `Subscript` expressions for table references

### Phase 6: Integration Test
- Add `test/lookup_arrayed/` with a model demonstrating per-element graphical functions
- Test verifies `LOOKUP(tables[A], input)` and `LOOKUP(tables[B], input)` work correctly

## Test plan
- [x] All 483 simlin-engine tests pass
- [x] All 24 simlin-compat simulation tests pass
- [x] New `simulates_lookup_arrayed` test passes

### DIFF
--- a/src/simlin-engine/src/compiler.rs
+++ b/src/simlin-engine/src/compiler.rs
@@ -3945,6 +3945,13 @@ impl<'module> Compiler<'module> {
                         }
                         Expr::StaticSubscript(off, view, loc) => {
                             // Static subscript - element offset is precomputed in the ArrayView
+                            // Reject ranges/wildcards - only single element selection is valid
+                            if view.size() > 1 {
+                                return sim_err!(
+                                    BadTable,
+                                    "range subscripts not supported in lookup tables".to_string()
+                                );
+                            }
                             (*off, Expr::Const(view.offset as f64, *loc))
                         }
                         Expr::Subscript(off, subscript_indices, dim_sizes, _loc) => {


bpowers:
@codex review

src/simlin-engine/src/units_infer.rs
```diff
@@ -111,10 +111,16 @@ impl UnitInferer<'_> {
...
                     Ok(Units::Explicit(UnitMap::new()))
                 }
-                BuiltinFn::Lookup(ident, _, _loc) => {
+                BuiltinFn::Lookup(table_expr, _, _loc) => {
                     // lookups have the units specified on the table
-                    let units: UnitMap =
-                        [(format!("@{prefix}{ident}"), 1)].iter().cloned().collect();
+                    let table_name = match table_expr.as_ref() {
+                        Expr2::Var(name, _, _) => name.as_str(),
+                        _ => return Ok(Units::Constant), // subscripted lookup tables not yet supported
```
bpowers: why not?  can't you "just" add another match arm for `Expr2::Subscript`?  we care about the type/units of that, the units of the expression that indexes into it doesn't matter

src/simlin-engine/src/json.rs
```diff
@@ -1073,11 +1080,13 @@ impl From<datamodel::Aux> for Auxiliary {
             datamodel::Equation::Arrayed(dims, elems) => {
                 let ees = elems
                     .into_iter()
-                    .map(|(subscript, equation, initial_equation)| ElementEquation {
-                        subscript,
-                        equation,
-                        initial_equation: initial_equation.unwrap_or_default(),
-                    })
+                    .map(
+                        |(subscript, equation, initial_equation, _gf)| ElementEquation {
```
bpowers: we don't want `_gf` to be unused -- we want full parity in our JSON representation.

src/simlin-engine/src/json.rs
```diff
@@ -449,6 +449,7 @@ impl From<Stock> for datamodel::Stock {
                                     } else {
                                         Some(ee.initial_equation)
                                     },
+                                    None, // per-element gf not supported in JSON format
```
bpowers: this needs to be fixed before we can merge

src/simlin-engine/src/bytecode.rs
```diff
@@ -563,8 +563,12 @@ pub(crate) enum Opcode {
     Apply {
         func: BuiltinId,
     },
+    /// Lookup a value in a graphical function table.
+    /// Stack: [..., element_offset, lookup_index] -> [..., result]
+    /// The actual table used is graphical_functions[base_gf + element_offset].
+    /// For scalar tables, element_offset is always 0.
     Lookup {
-        gf: GraphicalFunctionId,
+        base_gf: GraphicalFunctionId,
```
bpowers: I think this needs a bounds that indicates the max gf ID, so that if the expression evaluates to <= 0 or > max we properly handle the error (e.g. return `NaN`) instead of runtime panics or silent errors (using an adjacent table)

src/simlin-engine/src/compiler.rs
```diff
@@ -3914,12 +3936,113 @@ impl<'module> Compiler<'module> {
...
+                                        Box::new(term),
+                                        *_loc,
+                                    ),
+                                });
+
+                                // Update stride for next dimension
+                                stride *= dim_sizes.get(i).copied().unwrap_or(1);
+                            }
+
+                            (*off, offset_expr.unwrap_or(Expr::Const(0.0, *_loc)))
```
bpowers: I'm struggling to understand why we want to do `unwrap_or` here -- wouldn't it be better to either return an error, or if we _always_ expect to have an `offset_expr` to use a bare `unwrap()`?

bpowers:
@codex review

PR #149, opened 2026-01-04
engine: support dimension mappings for cross-dimension subscript resolution

bpowers:
In Vensim, dimensions can be mapped to other dimensions using syntax like
`DimA: A1, A2, A3 -> DimB`, which establishes a positional correspondence
between elements. This mapping was being lost during Vensim-to-XMILE
conversion, causing compilation failures when variables indexed by one
dimension reference variables indexed by a mapped dimension.

This change implements full dimension mapping support:

1. Updated the protobuf schema with a maps_to field on Dimension
2. Changed datamodel::Dimension from enum to struct with maps_to support
3. Updated xmutil C++ (XMILEGenerator.cpp) to extract dimension mappings
   from ExpressionSymbolList::Map() and emit <isee:maps_to> elements
4. Updated XMILE parser to deserialize maps_to (using local name since
   quick-xml strips namespace prefixes during deserialization)
5. Added maps_to tracking to NamedDimension and DimensionsContext with
   helper methods for translation between mapped dimensions
6. Updated compiler to use dimension mappings during subscript resolution
7. Updated builtins_visitor to pass DimensionsContext for dimension
   mapping during synthetic variable substitution in SMOOTH/DELAY builtins
8. Regenerated smooth.xmile test with dimension mapping information

The test/sdeverywhere/models/smooth model now simulates correctly with
its dimension mappings (e.g., DimA -> DimB, SubA -> DimB).

### DIFF
--- a/src/simlin-engine/src/compiler.rs
+++ b/src/simlin-engine/src/compiler.rs
@@ -1914,13 +1914,6 @@ impl Context<'_> {
                                     // the maps_to dimension (e.g., SubB is a subdimension of DimB,
                                     // and source_dim maps to DimB).
                                     let maps_to = self.dimensions_ctx.get_maps_to(source_dim_name);
-                                    eprintln!(
-                                        "DEBUG: source_dim={}, target_dim={}, maps_to={:?}, subscript={}",
-                                        source_dim_name.as_str(),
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #150, opened 2026-01-05
engine: support cross-dimension broadcasting in array reduction builtins

bpowers:
In Vensim, expressions like SUM(a[DimA!] + h[DimC!]) where DimA and DimC
are different dimensions produce a cross-product: the sum iterates over
all combinations (DimA x DimC). This is now supported.

The fix adds an allow_dimension_union flag to the Expr2Context trait.
When processing array reduction builtins (SUM, STDDEV, SIZE), the flag
is set to true before type-checking the inner expression. In
unify_dims_with_names(), when this flag is set, named dimensions are
allowed to combine into a union (cross-product) rather than requiring
exact match.

For example: a[DimA]=[1,2,3], h[DimC]=[10,20,30]
SUM(a[*]+h[*]) = (1+10)+(1+20)+(1+30)+(2+10)+(2+20)+(2+30)+(3+10)+(3+20)+(3+30) = 198

This enables the sum.xmile integration test from SDEverywhere.

### DIFF
--- a/src/simlin-engine/src/array_tests.rs
+++ b/src/simlin-engine/src/array_tests.rs
@@ -3205,4 +3205,87 @@ mod cross_dimension_reduction_tests {
         // 2 time steps: t=0 and t=1
         project.assert_interpreter_result("result", &[344.0, 344.0]);
     }
+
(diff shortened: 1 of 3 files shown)


bpowers:
@codex review

PR #151, opened 2026-01-05
build: auto-install protoc compiler in cloud-init.sh

bpowers:
The cloud-init.sh script now automatically downloads and installs the
protobuf compiler (protoc) from GitHub releases if it is not already
available on the system. This ensures that cargo build succeeds out of
the box after running the initialization script.

The implementation:
- Pins protoc version 33.2 (latest stable as of January 2026)
- Detects OS (Linux/macOS) and architecture (x86_64/aarch64)
- Downloads the appropriate binary from GitHub releases
- Installs to ~/.local/bin with retry logic for network failures
- Adds ~/.local/bin to PATH for the current session
- Falls back to manual installation instructions if auto-install fails

This enables a fully working development environment with just:
  ./scripts/cloud-init.sh
  cargo build

### DIFF
--- a/scripts/cloud-init.sh
+++ b/scripts/cloud-init.sh
@@ -28,90 +28,42 @@ cd "$REPO_ROOT"
 # Function to install protoc from GitHub releases
 install_protoc() {
     local version="$1"
-    local install_dir="$HOME/.local"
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #153, opened 2026-01-05
engine: optimize iteration view handling with LoadIterViewAt opcode

bpowers:
## Summary

Add new `LoadIterViewAt` opcode that allows accessing views by stack offset during iteration, avoiding the need to push/pop views on every iteration. This reduces view cloning from O(N*V) to O(V) where N is the iteration count and V is the number of source views.

## Changes

- **New `LoadIterViewAt` opcode** (`bytecode.rs`) - Loads from a view at a specific stack offset during iteration
- **VM handler** (`vm.rs`) - Implements the opcode with full broadcasting support
- **Compiler optimization** (`compiler.rs`) - Hoists source view pushes before the loop, uses `LoadIterViewAt` inside the loop body
- **Criterion benchmarks** (`benches/array_ops.rs`) - Measures iteration, reduction, and broadcasting performance

## Key Implementation Details

The compiler now:
1. Collects all source views referenced in the RHS expression
2. Pushes the output view, then calls `BeginIter` (which captures `view_stack.last()` as iteration bounds)
3. Pushes source views AFTER `BeginIter` - critical because `BeginIter` uses the top of stack
4. Inside the loop, `LoadIterViewAt` accesses pre-pushed views by offset instead of push/load/pop per view per iteration

## Benchmark Results

| Benchmark | Improvement |
|-----------|-------------|
| array_sum/100 | **-12%** |
| array_sum/500 | -4% |
| array_sum/1000 | -2% |
| elementwise_add/100 | **-7%** |
| same_array_refs | -2-3% |

Larger improvements for smaller arrays make sense - the overhead reduction from hoisting view pushes is a fixed cost per loop, so it's more prominent when iteration counts are lower.

## Test plan

- [x] All 533 existing tests pass
- [x] Benchmarks show expected performance improvements
- [x] Pre-commit checks pass (fmt, clippy, tests, lint)

### DIFF
--- a/src/simlin-engine/src/compiler.rs
+++ b/src/simlin-engine/src/compiler.rs
@@ -4070,17 +4070,13 @@ impl<'module> Compiler<'module> {
                     // In iteration context with optimized view hoisting
                     let static_view = self.array_view_to_static(*off, view);
 
-                    if let Some(offset) = self.find_iter_view_offset(&static_view) {
-                        // View was pre-pushed - use LoadIterViewAt with the offset
-                        self.push(Opcode::LoadIterViewAt { offset });
-                    } else {
-                        // Fallback: view not in pre-pushed set (shouldn't happen normally)
-                        // This can occur if the expression has views that weren't collected
-                        let view_id = self.add_static_view(static_view);
-                        self.push(Opcode::PushStaticView { view_id });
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #157, opened 2026-01-11
engine: improve unit inference and fix related bugs

bpowers:
Several improvements to the unit inference and checking system:

1. Fixed unit name canonicalization in build_unit_components to handle
   case-insensitive unit comparisons (e.g., "Month" vs "month").

2. Fixed subscripted expressions to inherit units from their base array.
   Previously `delay_a[A1]` was treated as dimensionless instead of
   having the same units as `delay_a`.

3. Updated the inference system to distinguish between actual unit
   mismatches (conflicting concrete units) and under-constrained models
   (unresolved metavariables). Only actual mismatches return errors.

4. Fixed multiple unit checking tests that were using incorrect time
   units or missing required unit declarations on delay parameters.

5. Added `array_const_with_units` helper to TestProject for creating
   arrayed constants with explicit units.

The inference system now correctly handles arrayed module parameters and
properly propagates units through stdlib function expansions. Inference
errors are not currently surfaced to maintain backwards compatibility
with models that have incomplete unit annotations.

### DIFF
--- a/src/simlin-engine/src/project.rs
+++ b/src/simlin-engine/src/project.rs
@@ -69,8 +69,17 @@ impl From<datamodel::Project> for Project {
             // with models that have incomplete or incorrect unit annotations. The check_units
             // call below will still catch mismatches between explicitly declared units.
             // TODO: Consider surfacing UnitMismatch errors once more test models have correct units.
-            let inferred_units =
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

PR #158, opened 2026-01-12
engine: add location information to unit inference errors

bpowers:
Thread constraint source information (variable identifier + optional equation location) through the unit inference constraint system. This enables more informative error messages that can identify which variable(s) are involved in a unit mismatch and point to specific positions within their equations.

The key insight is that unlike the compiler (which parses one equation at a time), unit inference mixes constraints from multiple variables. An error might involve a mismatch within one variable's equation, a conflict between two variables' declarations, or a structural relationship (stock/flow, module input) that has no equation location.

Changes:
- Add ConstraintSource and LocatedConstraint types to track constraint origins with variable names and optional Loc within that variable's equation
- Add InferenceError variant to UnitError with sources list
- Update constraint generation, substitution, and unification to preserve and merge constraint sources
- Update all consumers of UnitError (engine, libsimlin) to handle the new variant

Inference errors are still swallowed in project.rs to maintain backwards compatibility with models that have incomplete unit annotations, but the location information is now available for future use when we decide to surface these errors to users.

### DIFF
--- a/src/libsimlin/src/errors.rs
+++ b/src/libsimlin/src/errors.rs
@@ -353,4 +353,62 @@ mod tests {
             .contains("units error in model 'main' variable 'bad_units': unit_mismatch"));
         assert!(lines.next().is_none());
     }
+
+    #[test]
+    fn inference_error_formats_correctly() {
+        use simlin_engine::builtins::Loc;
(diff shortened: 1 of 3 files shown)


bpowers:
@codex review

PR #160, opened 2026-01-12
engine: surface unit mismatch errors during inference

bpowers:
## Summary

- Surface unit mismatch errors to users (previously silently suppressed)
- Only show errors for models with declared units to avoid false positives
- Add built-in singular/plural unit aliases (person/people, minute/minutes, etc.)
- Fix unit inconsistencies in several test models that were hidden by suppressed errors
- Enable previously ignored `test_smth1_unit_mismatch_initial` test

## Test plan

- [x] All Rust tests pass (561 passed)
- [x] All pysimlin tests pass
- [x] Pre-commit hooks pass (clippy, fmt, tsc, lint)


bpowers:
@codex review

PR #170, opened 2026-01-14
pysimlin: remove Python protobuf dependency

bpowers:
## Summary

- Remove the `protobuf` Python library dependency from pysimlin
- Delete the auto-generated `pb.py` module and all imports/exports of it
- Remove the `test_round_trip_protobuf` test (the only place that used Python protobuf to parse output)
- Remove the `protos:` Makefile target that regenerated `pb.py`

## Context

The Python protobuf library was only needed to *parse* protobuf output in one test. The `serialize_protobuf()` method works via C FFI (Rust handles serialization), so it doesn't require Python protobuf at runtime.

**What still works:**
- Loading `.pb` files (passes bytes directly to Rust engine)
- `serialize_protobuf()` method (returns bytes serialized by Rust; users can parse with their own protobuf dependency if needed)

## Test plan

- [x] All 259 pysimlin tests pass
- [x] `serialize_protobuf()` still returns valid bytes
- [x] Loading `.pb` files still works


bpowers:
@codex review

PR #177, opened 2026-01-16
engine: fix VM error handling and separate unit warnings

bpowers:
The VM creation error was being silently discarded in simlin_sim_new via
.ok(), causing users to see a generic "simulation has not been initialised
with a VM (code 32)" error instead of the actual compilation error.

This change:
1. Separates unit mismatch errors from blocking compilation errors. Unit
   mismatches are now stored in a new `unit_warnings` field on models
   instead of `errors`. This allows simulations to run even when models
   have unit inconsistencies (common in real-world models).

2. Stores VM creation errors in SimState so they can be reported when
   run_to/run_to_end is called, matching the pattern used in the WASM
   engine.

3. Updates pysimlin FFI bindings to include ErrorKind and UnitErrorKind
   fields that were previously missing from the SimlinErrorDetail struct.

The approach follows the existing WASM engine pattern where both the
vm and any creation error are stored, allowing proper error reporting
while still enabling simulation when only warnings (not errors) exist.

### DIFF
--- a/src/libsimlin/src/errors.rs
+++ b/src/libsimlin/src/errors.rs
@@ -53,8 +53,18 @@ pub struct FormattedErrors {
     pub has_variable_errors: bool,
 }
 
-/// Format all static errors for a compiled project, matching the CLI style.
-pub fn collect_formatted_errors(project: &engine::Project) -> FormattedErrors {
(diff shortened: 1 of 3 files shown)


bpowers:
@codex review

PR #224, opened 2026-02-02
engine: replace LALRPOP parser with hand-written recursive descent

bpowers:
## Summary

- Replace the LALRPOP-based equation parser (`equation.lalrpop`) with a hand-written recursive descent parser
- Remove lalrpop and lalrpop-util dependencies
- Add comprehensive parser tests (99 tests) including equivalence tests

## Details

The new parser in `src/parser/mod.rs` follows the same two-layer architecture (lexer + parser), reusing the existing lexer and producing identical AST output. All 740 existing tests pass.

Key implementation details:
- TokenKind enum for efficient peek comparisons without payload matching
- Proper precedence handling matching the original LALRPOP grammar
- Same composability restrictions preserved (function calls and subscripts only on identifiers, transpose only via postfix, if only at top level)
- Identical error codes and span positions
- Safe division (`//`) desugars to `safediv()` function call
- Zero-arity builtins still reified via `Expr0::reify_0_arity_builtins()`

## Test plan

- [x] All 740 existing tests pass
- [x] 99 new parser-specific tests added
- [x] Equivalence tests verify new parser produces identical AST to old parser
- [x] Clippy passes with no warnings

### DIFF
--- a/src/simlin-engine/src/parser/tests.rs
+++ b/src/simlin-engine/src/parser/tests.rs
@@ -1077,3 +1077,496 @@ fn test_loc_span_transpose() {
     assert_eq!(loc.start, 0);
     assert_eq!(loc.end, 2);
 }
+
+// ============================================================================
+// Additional coverage tests
+// ============================================================================
+
+#[test]
+fn test_error_extra_token() {
+    // Valid expression followed by extra content should fail with ExtraToken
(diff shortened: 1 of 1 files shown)


bpowers:
@codex review

bpowers:
@codex review

PR #236, opened 2026-02-05
diagram: fix group drag behavior for clouds and click-deselect

bpowers:
## Summary

- Fix click-deselect bug: clicking an already-selected element no longer dissolves the group selection. Uses deferred selection pattern (standard Figma/Illustrator behavior) where selection change is deferred to mouseUp and only applies if no drag occurred.
- Fix cloud drag selection: clouds were explicitly excluded from marquee/lasso selection, so inline chains (cloud-flow-stock-flow-cloud) couldn't be fully selected. Now includes clouds with center-in-rectangle containment.
- Fix cloud re-attachment: entering flow re-attachment mode when clicking a cloud now only activates when the cloud is the sole selection, not when part of a multi-selection.
- Consolidate duplicate selection logic in handleSetSelection into a single call to a new pure `computeMouseDownSelection` function with full test coverage.

## Test plan

- [x] New `selection-logic.test.ts` with 11 tests covering all selection state transitions
- [x] New cloud drag-selection tests in `group-selection.test.ts` (6 tests)
- [x] New inline chain group-movement tests in `group-movement.test.ts` (2 tests)
- [x] All existing tests pass (394 total)
- [ ] Manual: drag-select stock+flow+cloud -> all highlight -> drag group -> all move as unit
- [ ] Manual: drag-select inline chain -> drag -> all translate uniformly
- [ ] Manual: drag-select multiple elements -> click one -> drag -> entire group moves
- [ ] Manual: click single cloud -> drag onto stock -> flow re-attaches

### DIFF
--- a/src/diagram/drawing/Canvas.tsx
+++ b/src/diagram/drawing/Canvas.tsx
@@ -52,6 +52,7 @@ import { shouldShowVariableDetails } from './pointer-utils';
 import {
   computeMouseDownSelection,
   computeMouseUpSelection,
+  pointerStateReset,
   resolveSelectionForReattachment,
 } from '../selection-logic';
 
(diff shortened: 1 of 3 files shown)


bpowers:
## Review Cycle Summary

During the review cycle, one issue was identified and fixed: the deferred-selection early-return path for non-named elements (like clouds) was missing pointer state cleanup. When a user clicked on an already-selected cloud in a group, the `moveDelta` and movement flags were left in a stale state, causing unnecessary `applyGroupMovement` calls on subsequent renders.

The fix extracts a `pointerStateReset()` function into `selection-logic.ts` that provides the canonical set of state fields to clear when pointer interactions end. This is now used by both `clearPointerState` and the deferred-click early-return, ensuring consistent cleanup across all pointer-end code paths. The function is covered by 4 new tests.

PR #240, opened 2026-02-07
engine: replace protobuf patch types with native Rust types, add module support

bpowers:
The apply_patch API previously converted JSON input to protobuf
(project_io::ProjectPatch) as an intermediate format before applying
operations to the datamodel. This was a leftover from when the API was
protobuf-based.

This change defines native Rust patch types (ProjectPatch, ModelPatch,
ProjectOperation, ModelOperation) that use datamodel types directly,
eliminating the protobuf round-trip. The JSON conversion in libsimlin
now goes directly from JSON -> native patch types -> datamodel, which
is simpler and more maintainable.

Additionally, this adds:
- AddModel project operation for creating new models within a patch
  transaction, enabling atomic "add model + add module referencing it"
- Comprehensive test coverage for module upsert, delete, replace,
  module references, and combined AddModel+UpsertModule operations
- Test for patch rollback behavior on error
- Fix pre-existing TS type error in core/datamodel.ts (List<string>)

https://claude.ai/code/session_011dCiV2VEYEPkUekXvBayyh

### DIFF
--- a/src/libsimlin/src/lib.rs
+++ b/src/libsimlin/src/lib.rs
@@ -3120,7 +3120,7 @@ unsafe fn apply_project_patch_internal(
         project_locked.datamodel.clone()
     };
 
-    if let Err(err) = engine::apply_patch(&mut staged_datamodel, &patch) {
+    if let Err(err) = engine::apply_patch(&mut staged_datamodel, patch) {
         store_error(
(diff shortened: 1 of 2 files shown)


bpowers:
@codex review

PR #241, opened 2026-02-07
engine: bytecode VM optimizations

bpowers:
## Summary

A series of bytecode VM optimizations that improve simulation performance by ~45% on the slider_interaction benchmark (130ms -> 73ms):

- Replace Vec-based arithmetic stack with fixed-size array and unsafe unchecked access (~17% speedup)
- Group 11-14 mutable eval arguments into EvalState struct to reduce register spilling (~32% speedup from here on)
- Move inline dim arrays from opcodes to a side table, shrinking Opcode from 12 to 8 bytes for better icache density
- Optimize run_to loop: pre-fill constants once instead of per-step, simplify buffer access patterns
- Add peephole optimizer that fuses common opcode sequences into superinstructions (LoadConstant+Assign, BinOp+Assign)

## Test plan

- 77 new unit tests covering stack operations, superinstruction execution, peephole optimizer patterns, dim list side table, reset/run_to behavior
- All existing tests pass (cargo test)
- 90.6% coverage on changed lines

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "packageManager": "pnpm@10.6.0",
   "devDependencies": {
     "@playwright/test": "^1.58.1",
-    "npm-run-all": "^4.1.5",
     "prettier": "^3.8.1",
     "ts-protoc-gen": "^0.15.0",
     "typescript": "^5.9.3"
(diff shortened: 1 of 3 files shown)


bpowers:
## Review cycle summary

Addressed feedback from 7 iterations of automated code review (Codex + Claude), resulting in 5 additional commits that strengthen the safety argument and documentation:

**Compile-time stack depth validation**: Added a static analysis pass in `ByteCodeBuilder::finish()` that computes the maximum arithmetic stack depth of compiled bytecode by walking the opcode stream and tracking each instruction's stack effect. This makes the VM's unsafe unchecked stack access provably safe rather than relying on informal reasoning about SD expression depth. The validation uses `checked_sub` (not `saturating_sub`) to eagerly catch bugs in the `stack_effect()` metadata.

**Defensive assertions and documentation**: Converted the peephole optimizer's jump target bounds check from a silent skip to a hard assertion. Added `debug_assert!(did_initials)` to `get_value_now()` to enforce the documented precondition. Added thorough SAFETY comments on both `push()` and `pop()` referencing the compile-time proof. Documented design decisions on `get_dim_list` panicking, `reset()` not zeroing data, `PushSubscriptIndex`/`LoadSubscript` stack effects, `EvalModule` result routing, and the peephole optimizer's jump target fusion policy -- all to ensure these deliberate choices aren't mistaken for oversights in future reviews.

PR #243, opened 2026-02-08
Simplify VM overrides to only support simple constants

bpowers:
## Summary

- Replaces the evaluate-then-patch override system with inline override checking during `AssignConstCurr` opcode execution
- Narrows the API: `set_value()` now only accepts simple constants (variables whose equation is a literal number), rejecting stocks, flows, and computed auxiliaries with an error
- Fixes a correctness issue where overrides only applied during the initials phase but not during flow computation in simulation steps
- Uses `is_empty()` fast-path so there's zero overhead when no overrides are set
- Renames API: `set_override` -> `set_value`, `clear_overrides` -> `clear_values`
- Removes `simlin_sim_set_override`, `simlin_sim_set_override_by_offset`, `simlin_sim_clear_overrides` FFI functions

## Test plan

- [x] All 1656 engine Rust tests pass
- [x] All libsimlin FFI tests pass
- [x] All TypeScript tests pass (updated to use constant variables)
- [x] All Python tests pass (updated to use constant variables)
- [x] Pre-commit hook passes all checks

### DIFF
--- a/src/libsimlin/src/simulation.rs
+++ b/src/libsimlin/src/simulation.rs
@@ -405,8 +405,7 @@ pub unsafe extern "C" fn simlin_sim_set_value(
     let mut state = sim_ref.state.lock().unwrap();
     if let Some(ref mut vm) = state.vm {
         match vm.set_value(&canon_name, val) {
-            Ok(()) => {
-                let off = vm.get_offset(&canon_name).unwrap();
+            Ok(off) => {
                 state.overrides.insert(off, val);
             }
             Err(err) => {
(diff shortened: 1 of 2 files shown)


bpowers:
## Review cycle summary

Ran two iterations of code review (Codex + Claude). The first iteration identified two improvements that were addressed in commit e57b9525:

1. **Eliminated redundant clone in `apply_override`**: The method was cloning `constant_info[&off]` twice -- once for saving originals and once for writing new literals. Restructured to clone once and reuse the local variable for both purposes.

2. **`set_value` now returns the data-buffer offset**: Previously returned `Result<()>`, which forced the FFI layer in libsimlin to do a redundant `get_offset` lookup (with an `unwrap()`) immediately after a successful `set_value` call. The offset is now returned directly from `set_value`, and the FFI callsite uses it without any unsafe unwrap. A new test (`test_set_value_returns_correct_offset`) verifies the returned offset matches expectations.

The second iteration confirmed both reviewers found no remaining actionable issues.

PR #246, opened 2026-02-14
pysimlin: modernize tooling and add free-threaded Python safety

bpowers:
Replace black with ruff format, expand ruff lint rules (SIM, TCH, RUF,
PT, C4, UP, B), and modernize type annotations to use PEP 604/585
syntax with `from __future__ import annotations` across all source
files.

Add per-object `threading.Lock` to Project, Model, and Sim wrapper
classes so that each instance is safe to use from multiple threads.
The underlying Rust FFI layer already uses per-object Mutexes for the
C-level state; the Python locks protect Python-level invariants (_ptr,
cached properties, _ran flag) which matter under free-threaded Python
(PEP 703 / 3.13t+ / 3.14) where the GIL no longer serialises access.

Model uses double-checked locking for cached properties and releases
its lock before calling Project methods to avoid lock-ordering
deadlocks.  Module-level `_finalizer_refs` (WeakValueDictionary) is
protected by `_refs_lock` in _ffi.py.

New test_thread_safety.py exercises concurrent loads, shared-object
access, context-manager cleanup races, and finalizer registry safety.

https://claude.ai/code/session_01D29RcQWGT6oKwKE9fMWoVM


bpowers:
@codex review

PR #252, opened 2026-02-15
engine: return Cow<str> from canonicalize to avoid defensive allocations

bpowers:
## Summary

- Changed `canonicalize()` to return `Cow<'_, str>` instead of `Ident<Canonical>`, avoiding allocation when the input is already in canonical form (lowercase, no whitespace, no quotes, no periods needing conversion)
- Added `is_canonical()` fast-path check and `Ident::new()` constructor for sites that need an owned `Ident`
- HashMap lookups now use `&*canonicalize(x)` to leverage `Borrow<str>` on `Ident` keys, avoiding allocation entirely for already-canonical names

## Test plan

- [x] All 1788 Rust unit tests pass
- [x] Clippy clean
- [x] TypeScript type check and tests pass
- [x] WASM build succeeds
- [x] pysimlin tests pass
- [x] Full pre-commit hook passes

### DIFF
--- a/src/simlin-engine/src/common.rs
+++ b/src/simlin-engine/src/common.rs
@@ -276,8 +276,11 @@ fn is_canonical(name: &str) -> bool {
                     return false;
                 }
             }
-            // Uppercase letters get lowercased
-            c if c.is_uppercase() => return false,
+            // Reject any character that to_lowercase() would change.
+            // This covers both uppercase (Lu) and titlecase (Lt) Unicode
+            // categories -- titlecase letters like ǅ are NOT uppercase but
+            // to_lowercase() still maps them to a different character (ǆ).
+            c if c.to_lowercase().ne(std::iter::once(c)) => return false,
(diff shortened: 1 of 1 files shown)


bpowers:
This PR went through 3 iterations of automated code review (codex + claude). The review process surfaced several improvements that were addressed in follow-up commits:

The first review round identified a whitespace handling gap where tab characters weren't treated consistently between the `is_canonical` fast path and the `replace_whitespace_with_underscore` slow path. Both paths now handle tabs, and a proptest was added to verify the fast-path/slow-path invariant holds for arbitrary input strings. The review also caught unused API surface (`from_canonical_cow`, `Ident::from_raw`, and orphan-adjacent `PartialEq` impls) which was removed, and identified that `sort_by_key` with `canonicalize().into_owned()` should use `sort_by_cached_key` to avoid calling the closure O(n log n) times instead of once per element.

PR #261, opened 2026-02-18
engine: add AST deps, LTM metadata, and layout metadata API

bpowers:
## Summary

Overhauls the layout engine's dependency extraction, feedback loop analysis, and annealing pipeline. The public API now accepts a project + model name instead of a bare model, enabling compilation-based analysis.

**AST-based dependency extraction**: replaces the string-heuristic approach with `identifier_set()` and `resolve_non_private_dependencies()` from the compiler. Falls back to the heuristic per-variable when compilation fails. Eliminates false positives from builtin name collisions (e.g. `TIME`) and substring matching (e.g. `birth` inside `birthday`). Phantom deps from module-internal refs (e.g. `m.out`) are filtered against the set of rendered model variables so they don't become invisible force-directed nodes.

**LTM integration**: `compute_metadata` now runs the full Loops That Matter pipeline -- loop detection, LTM-instrumented simulation, and per-loop importance time series extraction -- falling back to persisted `loop_metadata` UIDs when any step fails. Non-finite importance values are replaced with `0.0` to preserve series alignment with simulation timesteps. Loops are sorted descending by average importance.

**Dominant period calculation**: new `DominantPeriod` struct with `start`/`end` time bounds, score-ranked `dominant_loops`, and `combined_score` averaged across all timesteps in the period. At each timestep, loops of the same polarity as the leader are accumulated greedily until they explain >= 50% of behavior. Consecutive timesteps with the same dominant set are grouped; zero-score gaps force splits.

**Annealing improvements**: per-node displacement limits (tight 25-unit limit for chain nodes, 200-unit for auxiliaries) and strongest-neighbor coupled motion at 50% displacement. SFDP now interleaves annealing rounds via an iteration callback, tracking the best-seen crossing count and layout so that SFDP drift after a good annealing round doesn't regress the result. Config fields are validated with minimum clamping.

**Other changes**:
- chain importance scoring (`stocks*10 + flows*5`) with descending sort
- label-aware view bounds calculation using `estimate_label_bounds`
- `compute_layout_metadata()` entry point for analysis without full layout generation
- `resolve_model_name` helper so the `"main"` alias works in compilation and LTM paths
- `create_connectors` gracefully skips phantom deps instead of failing layout generation
- `create_missing_auxiliary_elements` returns `Result` instead of silently defaulting

## Test plan

- [x] Existing layout integration tests updated for project-scoped API
- [x] `test_ltm_populates_loop_importance` -- LTM produces non-empty importance series
- [x] `test_ltm_detects_polarity` -- loops get Reinforcing/Balancing polarity
- [x] `test_loops_sorted_by_average_importance` -- descending sort invariant
- [x] `test_compute_layout_metadata_has_dominant_periods` -- end-to-end dominant period validation
- [x] `test_compute_layout_metadata_chains_sorted` -- chain importance descending sort
- [x] `test_generate_layout_with_zero_reheat_does_not_panic` -- config validation clamping

### DIFF
--- a/src/simlin-engine/src/layout/mod.rs
+++ b/src/simlin-engine/src/layout/mod.rs
@@ -1964,8 +1964,7 @@ fn try_detect_ltm_loops(
 
             // Extract the relative loop score time series
             let var_name = format!("$⁚ltm⁚rel_loop_score⁚{}", loop_item.id);
-            let var_ident =
-                Ident::<Canonical>::from_str_unchecked(&Ident::new(&var_name).to_source_repr());
+            let var_ident = Ident::new(&var_name);
             let importance_series = vm
                 .get_series(&var_ident)
                 .unwrap_or_default()


bpowers:
@codex review

bpowers:
@codex review

PR #262, opened 2026-02-18
engine: move active_initial from Equation enum to Compat struct

### LINKED ISSUE #248: Clean up initialEquation
We added initialEquation to our datamodel to support Vensims active initial concept.  This is important for comparability, but is not a suitably "central" concept to SD to put it on the same level/literally next to "equation" in our datamodel and JSON schemas.  We should introduce an optional `compat` field on stock, flow, aux and module and shove what we currently call initialEquation on that as activeInitial.

bpowers:
## Summary

- Moves Vensim's ACTIVE INITIAL concept from Option<String> in Equation::Scalar/ApplyToAll variants to a new Compat struct on Stock/Flow/Aux variables
- Separates compatibility features from core SD equation concepts
- Protobuf retains backward compat: reads old initial_equation from ScalarEquation/ApplyToAllEquation, writes only compat.active_initial
- All other formats (JSON, TypeScript, Python) fully migrated with no shims

Fixes #248

## Test plan

- [x] Existing engine tests pass
- [x] Integration tests with Vensim models using ACTIVE INITIAL pass
- [x] Protobuf round-trip backward compatibility verified

### DIFF
--- a/src/pysimlin/simlin/json_converter.py
+++ b/src/pysimlin/simlin/json_converter.py
@@ -503,10 +503,14 @@ def _create_converter() -> cattrs.Converter:
         gf = None
         if d.get("graphicalFunction"):
             gf = conv.structure(d["graphicalFunction"], GraphicalFunction)
+        compat = None
+        compat_dict = d.get("compat")
+        if compat_dict:
+            compat = Compat(active_initial=compat_dict.get("activeInitial"))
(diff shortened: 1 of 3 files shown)


bpowers:
@codex review

bpowers:
@codex review

PR #280, opened 2026-02-21
build: parallelize pre-commit hook and enable build caching

bpowers:
## Summary

- Restructure pre-commit hook into three concurrent pipelines (Rust, TypeScript, Python) that run in parallel, with further parallelism within each pipeline (clippy & test, tsc & test)
- Stop destroying incremental caches in `engine/build.sh` (`rm -rf lib lib.browser core *.tsbuildinfo` removed; WASM only copied when changed)
- Enable ESLint `--cache` across all packages and run lint/tsc/test in parallel via pnpm
- Optimize pysimlin tests: skip CFFI rebuild when libsimlin.a unchanged, add pytest-xdist for parallel test execution, skip wheel build locally

**Result: pre-commit wall clock drops from ~90s to ~26s (3.5x faster) with warm caches.**

## Test plan

- [x] `time bash scripts/pre-commit` passes and completes in ~26s (was ~90s)
- [x] Introducing Rust format error is detected and reported
- [x] Introducing TypeScript type error is detected and reported
- [x] Exit code is non-zero on failure
- [x] `pnpm lint`, `pnpm tsc`, `pnpm test` all work standalone
- [x] Engine build caches correctly (second run is fast)
- [x] ESLint cache populates and subsequent runs are faster

### DIFF
--- a/.claude/skills/address-feedback-github/SKILL.md
+++ b/.claude/skills/address-feedback-github/SKILL.md
@@ -3,7 +3,17 @@ name: address-feedback-github
 description: Improve the current branch's changes by getting code reviews from GitHub-based reviewers in a loop until all feedback is addressed
 ---
 
-You are performing iterative code review and improvement using GitHub-based reviewers. This skill takes no arguments (if any were given, ignore). You are operating on the local checkout.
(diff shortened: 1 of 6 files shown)


bpowers:
@codex review

Issue #297, opened 2026-02-23
libsimlin: sync_state TOCTOU race in apply_project_patch_internal


bpowers:
## Problem

In `src/libsimlin/src/patch.rs`, the `apply_project_patch_internal` function has a critical time-of-check to time-of-use (TOCTOU) race condition involving `sync_state`. The problem occurs between taking `sync_state` at line 441 and restoring it at lines 518-520:

1. Line 441: `sync_state.lock().unwrap().take()` sets sync_state to None
2. Lines 442-446: Incremental sync to staged state
3. Lines 501-523: If patch is rejected or dry_run, attempt to restore prev_state

Between steps 1 and 3, another thread calling `apply_project_patch_internal` can:
- Observe `sync_state == None` (because it was taken in step 1)
- Call `sync_from_datamodel_incremental` with `prev_state = None`
- Resync the DB from scratch based on the original datamodel
- Hit the reject/dry-run branch (lines 501-523)
- Execute line 520: `*sync_state.lock().unwrap() = None`
- Return without restoring the original sync state

Result: The salsa DB and sync handles become inconsistent with the committed project state.

## Why It Matters

This is a correctness issue affecting multi-threaded callers (Python bindings, any concurrent patch calls). While the WASM/TS path is single-threaded, the API is documented as thread-safe (lines 544-547). A concurrent patch call during validation/revert can corrupt the DB state permanently, causing silent failures in subsequent simulations.

## Affected Component

- `src/libsimlin/src/patch.rs` - `apply_project_patch_internal` function (lines 441, 501-523)
- Root cause: Taking vs. restoring sync_state is not atomic with respect to concurrent patch calls

## Possible Solutions

1. **Clone instead of take** (safest, minimal change): Clone prev_state instead of taking it at line 441, only update sync_state on commit (line 529). Concurrent readers would observe a consistent state.

2. **Lock-based atomicity**: Hold sync_state lock for the entire operation (lines 441-530), preventing other threads from observing intermediate None states.

3. **Read-write lock**: Use RwLock instead of Mutex to allow concurrent reads but exclusive writes during patch application.

## Context

Identified during code review of PR #289 (incremental compilation implementation).

PR #300, opened 2026-02-23
engine: incremental compilation followup fixes

### LINKED ISSUE #302: Box::leak in build_submodel_metadata causes unbounded memory growth
## Problem

`build_submodel_metadata` in `src/simlin-engine/src/db.rs` (line 3007) uses `Box::leak(Box::new(stub))` to create `&'a Variable` references for sub-model variable stubs. This leaks memory on every call and is never reclaimed.

```rust
let stub: &'a crate::variable::Variable = Box::leak(Box::new(stub));
```

The comment in the surrounding code suggests salsa caching limits invocations, but `build_submodel_metadata` is a plain function -- not a `#[salsa::tracked]` function -- so salsa does not cache or deduplicate its calls. Its callers (`compile_implicit_var_fragment`, `compile_var_fragment`) are also non-tracked. Each recompilation leaks O(M) `Variable` stubs where M is the number of sub-model variables.

Individual stubs are small, but in a long-running editor session with frequent recompilation (e.g., the user typing equations), this grows without bound.

## Why It Matters

- **Correctness**: Unbounded memory growth in a long-running process (the editor) will eventually degrade performance or exhaust memory.
- **Developer experience**: The `Box::leak` pattern here is surprising and misleading -- the comment's claim about salsa caching is inaccurate for a non-tracked function.

## Component(s) Affected

- `src/simlin-engine/src/db.rs` -- `build_submodel_metadata` function (around line 3007)

## Possible Approaches

1. **Arena allocation**: Use a typed arena (e.g., `typed_arena::Arena` or `bumpalo`) scoped to the compilation unit, so all stubs are freed together when compilation completes.
2. **Salsa interning**: Make the stub variables salsa-interned or tracked values so they participate in salsa's lifecycle management.
3. **Refactor `VariableMetadata`**: Change `VariableMetadata` to own its `Variable` (or use `Arc<Variable>`) instead of requiring a borrowed reference with lifetime `'a`, eliminating the need for leaked references.

## Context

Identified during PR #300 code review.

### LINKED ISSUE #303: libsimlin: sync_state TOCTOU race in simlin_project_add_model
## Problem

In `src/libsimlin/src/project.rs`, the `simlin_project_add_model` function has the same class of sync_state TOCTOU race as #297 (which covers `apply_project_patch_internal` in `patch.rs`).

At lines 362-369:

```rust
let prev_state = proj.sync_state.lock().unwrap().take();  // line 362: sets sync_state to None
let new_state = engine::db::sync_from_datamodel_incremental(  // lines 363-367: rebuild DB
    &mut db,
    &project_locked.datamodel,
    prev_state.as_ref(),
);
drop(db);                                                    // line 368: release DB lock
*proj.sync_state.lock().unwrap() = Some(new_state);          // line 369: restore sync_state
```

Between `drop(db)` on line 368 and restoring sync_state on line 369, another thread calling `simlin_sim_new` can acquire the DB lock, observe `sync_state` as `None`, and receive a `NotSimulatable` error because `sync_from_datamodel_incremental` is called with `prev_state = None`, triggering a full resync from scratch rather than using the incremental path.

## Why It Matters

This is a correctness issue for multi-threaded callers (Python bindings, concurrent FFI calls). The API is documented as thread-safe, so concurrent `add_model` + `sim_new` calls should not produce spurious errors. The impact is a transient `NotSimulatable` failure that is non-deterministic and difficult to reproduce, making it a particularly confusing bug for users.

## Components Affected

- `src/libsimlin/src/project.rs` - `simlin_project_add_model` function (lines 362-369)

## Possible Solutions

The fix should restructure the lock protocol so sync_state is never in a `None` state visible to concurrent readers. Two straightforward approaches:

1. **Hold the DB lock until after sync_state is restored**: Move `drop(db)` after the sync_state assignment on line 369, so no concurrent reader can observe the intermediate state.

2. **Clone instead of take**: Replace `.take()` with `.clone()` (as suggested for #297), so sync_state always holds a valid value and concurrent readers see the previous consistent state.

This does NOT warrant restoring the monolithic compilation fallback -- the incremental-only path is correct; the issue is purely about lock ordering.

## Context

- Identified during Codex review of PR #300
- Same class of bug as #297 (sync_state TOCTOU in `apply_project_patch_internal`), but in a different function
- Related to #298 (staged DB state visibility during patch validation)

bpowers:
## Summary

Incremental compilation correctness fixes, plus two issues identified during code review:

- **Incremental compilation fixes**: error handling now surfaces real compile failures instead of generic messages; array table placeholders preserved for sparse per-element graphical functions; implicit MODULE variables get correct sub-model slot sizes in flattened offset calculation
- **Module refs consistency**: strip module ident prefix from implicit var module_refs, matching all other construction sites
- **TOCTOU race fix** (#303): `simlin_project_add_model` used `.take()` on `sync_state` and dropped the db lock before restoring it, allowing concurrent `simlin_sim_new` to see `None` and fail. Now uses `.clone()` and holds the db lock through the restore, matching the #297 fix in `patch.rs`
- **Memory leak fix** (#302): `build_submodel_metadata` used `Box::leak` to create stub variable references, leaking memory on every recompilation. Replaced with `bumpalo::Bump` arena allocation scoped to each compilation function, so stubs are freed when compilation completes

Fixes #302
Fixes #303

## Test plan
- [x] `cargo test -p simlin-engine` -- 2139 tests
- [x] `cargo test --features file_io --test simulate` -- 32 integration tests
- [x] `cargo test -p libsimlin` -- 151 tests including new concurrency regression test
- [x] Full pre-commit hooks (clippy, fmt, WASM build, TS lint/build/test, pysimlin)

### DIFF
--- a/src/simlin-engine/Cargo.toml
+++ b/src/simlin-engine/Cargo.toml
@@ -38,6 +38,7 @@ csv = { version = "1", optional = true }
 resvg = { version = "0.47", optional = true }
 xmutil = { version = "1", path = "../xmutil", optional = true }
 
+bumpalo = "3"
 salsa = "0.26"
 
 rand = { version = "0.9", default-features = false, features = ["std_rng"] }
--- a/src/simlin-engine/src/db.rs
+++ b/src/simlin-engine/src/db.rs
@@ -2967,13 +2967,14 @@ fn build_stub_variable(
 /// Populate sub-model metadata in `all_metadata` for module variable compilation.
 /// Mirrors the monolithic `build_metadata` but works with salsa SourceModel/SourceVariable.
 /// Recursively populates metadata for nested modules.
-fn build_submodel_metadata<'a>(
-    db: &'a dyn Db,
+fn build_submodel_metadata<'arena>(
+    arena: &'arena bumpalo::Bump,
+    db: &dyn Db,
     sub_model: SourceModel,
     project: SourceProject,
     all_metadata: &mut HashMap<
         Ident<Canonical>,
-        HashMap<Ident<Canonical>, crate::compiler::VariableMetadata<'a>>,
+        HashMap<Ident<Canonical>, crate::compiler::VariableMetadata<'arena>>,
     >,
 ) {
     let sub_model_name: Ident<Canonical> = Ident::new(sub_model.name(db));
@@ -2986,10 +2987,7 @@ fn build_submodel_metadata<'a>(
     let source_vars = sub_model.variables(db);
     let project_models = project.models(db);
 
-    // We need to store parsed/lowered variables so their references live
-    // long enough. Use a leaked Box to make them 'static (acceptable since
-    // these are cached by salsa and only created once per variable).
-    let mut sub_metadata: HashMap<Ident<Canonical>, crate::compiler::VariableMetadata<'a>> =
+    let mut sub_metadata: HashMap<Ident<Canonical>, crate::compiler::VariableMetadata<'arena>> =
         HashMap::new();
 
     let mut sorted_names: Vec<&String> = source_vars.keys().collect();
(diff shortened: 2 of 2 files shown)


bpowers:
@codex review

PR #332, opened 2026-02-25
engine: fix mdl roundtrip edge cases

### LINKED ISSUE #325: MDL writer: format_f64 emits "inf"/"-inf" for infinity values
## Description

`format_f64()` in `src/simlin-engine/src/mdl/writer.rs` (line 602-608) handles NaN correctly (falls through to Rust's default `Display` formatting which produces `NaN`) but has no special-case handling for positive or negative infinity. When `f64::INFINITY` or `f64::NEG_INFINITY` is passed in, Rust's `Display` trait emits `inf` and `-inf` respectively.

Vensim does not recognize `inf`/`-inf` as valid numeric literals. MDL files typically represent these cases using `:NA:` or very large sentinel numbers (e.g., `1e+38`).

Current code:

```rust
fn format_f64(v: f64) -> String {
    if v == v.trunc() && v.abs() < 1e15 {
        format!("{}", v as i64)
    } else {
        format!("{}", v)
    }
}
```

The `v.abs() < 1e15` guard prevents the integer fast path for infinity (since `f64::INFINITY.abs()` is infinity, not less than 1e15), so it falls through to the else branch which emits `inf`.

## Why it matters

- **Vensim compatibility**: MDL files containing `inf` would fail to load in Vensim, breaking the roundtrip contract of the MDL writer.
- **Correctness**: Even though infinity values are unlikely in practice (they could arise from division by zero in graphical function scales or other edge cases), the writer should produce valid MDL output for all possible floating-point inputs.

## Component(s) affected

- `simlin-engine` -- specifically `src/simlin-engine/src/mdl/writer.rs`, the `format_f64()` function

## Possible approaches

1. Add explicit checks for `v.is_infinite()` before the existing logic, emitting `:NA:` (or a Vensim-compatible large number like `1e+38`) for positive infinity and `-1e+38` (or `:NA:`) for negative infinity.
2. Investigate what Vensim actually writes in these cases by examining real `.mdl` files and match that behavior.

## Context

Identified during review of PR #324 (MDL writer implementation). The NaN case is already handled implicitly (Rust emits `NaN` which may also need attention), but infinity is the more immediately concerning case since it produces output Vensim cannot parse.

### LINKED ISSUE #326: datamodel::StockFlow lacks a name/title field
## Problem

`datamodel::StockFlow` does not carry a view title/name field. As a result, `write_stock_flow_view()` in the MDL writer (`src/simlin-engine/src/mdl/writer.rs`) hardcodes `"View 1"` as the sketch view title. This means the original view name from a parsed MDL file is lost on roundtrip -- reading a model with a view named e.g. `"Overview"` and writing it back out will produce `"View 1"` instead.

## Why it matters

- **Correctness**: View names carry semantic meaning in Vensim models. Losing them on roundtrip is a fidelity regression.
- **Roundtrip fidelity**: MDL read-write-read cycles should preserve all metadata that the engine understands. This is one of the remaining gaps.

## Components affected

- `src/simlin-engine/src/datamodel.rs` -- `StockFlow` struct needs a `name: Option<String>` (or similar) field
- `src/simlin-engine/src/mdl/writer.rs` -- `write_stock_flow_view()` should use the field instead of hardcoding `"View 1"`
- `src/simlin-engine/src/mdl/parser.rs` (or convert layer) -- needs to populate the new field when parsing view headers
- XMILE reader/writer may also need updates for consistency

## Possible approach

1. Add `pub name: Option<String>` to `datamodel::StockFlow`.
2. Populate it during MDL parsing from the view header line (e.g. `*View 1`).
3. Use it in the MDL writer, falling back to `"View 1"` when `None`.
4. Update XMILE serde if `<views><view name="...">` is not already handled.
5. Add a roundtrip test that verifies view name preservation.

## Context

Identified during review of PR #324 (MDL writer implementation). The hardcoded value was an intentional pragmatic choice to keep the PR scope manageable.

### LINKED ISSUE #328: MDL writer: write_stock_variable drops initial value for arrayed stocks
## Problem

`write_stock_variable()` in `src/simlin-engine/src/mdl/writer.rs` (lines 704-707) only handles `Equation::Scalar` when extracting the initial value for the INTEG equation:

```rust
let initial = match &stock.equation {
    Equation::Scalar(eqn) => equation_to_mdl(eqn),
    _ => String::new(),
};
```

For stocks with `Equation::ApplyToAll` or `Equation::Arrayed`, this produces an empty initial value, emitting malformed MDL like `INTEG(flow, )`. In Vensim MDL, subscripted stocks need per-element INTEG equations or a single INTEG with the correct initial expression for each subscript combination.

## Why it matters

- **Correctness**: MDL roundtrip for any subscripted stock model will lose initial values, producing invalid or semantically wrong output.
- **Maintainability**: The `_ => String::new()` catch-all silently degrades rather than signaling an unhandled case.

## Components affected

- `src/simlin-engine/src/mdl/writer.rs`, specifically `write_stock_variable()` (line 688)

## Possible approaches

1. Handle `Equation::ApplyToAll` by extracting the inner equation string and calling `equation_to_mdl` on it (same as Scalar, since it applies uniformly).
2. Handle `Equation::Arrayed` by reconstructing per-subscript INTEG equations in the MDL format that Vensim expects (e.g., separate equations with subscript qualifiers, or a single equation referencing the appropriate subscripted initial values).
3. Add integration test models with subscripted stocks to exercise the roundtrip path.

## Context

Identified during review of PR #324 (MDL writer). No current test model exercises this path -- the existing MDL roundtrip tests use scalar-only models.

### LINKED ISSUE #329: MDL writer: write_flow_element omits pipe connector geometry (type 1 lines)
## Problem

`write_flow_element()` in `src/simlin-engine/src/mdl/writer.rs` (line 909) emits only two lines per flow:

- **Type 11** (valve): position and visual properties
- **Type 10** (attached variable): name and label position

It does **not** use `flow.points` to emit the **type 1 connector lines** that represent the pipe geometry -- the line segments connecting stocks/clouds to the valve. This means stock/cloud attachments at flow endpoints are lost during MDL roundtrip. The MDL reader correctly parses these type 1 lines (see `view/elements.rs:201`), so the data is available on the `Flow` struct's `points` field, but the writer never consumes it.

## Impact

- **Visual fidelity**: Flow pipe paths (start/end connections, intermediate waypoints) are lost in written MDL sketches. Flows appear as isolated valves with no visible connections.
- **Roundtrip correctness**: MDL -> datamodel -> MDL loses diagram connectivity information.
- **Equation semantics**: Not affected. Flow equations and stock linkages are preserved correctly through the equation section of the MDL file.

## Component(s) affected

- `src/simlin-engine/src/mdl/writer.rs` -- `write_flow_element()` function (line 909)

## Possible approach

For each flow element, in addition to the existing type 11 and type 10 lines, emit type 1 connector lines derived from `flow.points`. Each point pair defines a pipe segment. The from/to UIDs for the first and last segments should reference the connected stock/cloud UIDs (available via `flow.points`), and intermediate segments reference the valve UID. The connector line format is documented in `view/elements.rs:201-203`.

## Context

Identified during review of PR #324 (MDL writer).

bpowers:
## Summary
- preserve MDL view titles by adding `name: Option<String>` to `datamodel::StockFlow` and wiring it through MDL/XMILE/JSON/protobuf conversion paths
- fix `format_f64` to emit finite MDL-compatible sentinels for `+/-infinity` instead of Rust's `inf`/`-inf`
- refactor stock MDL emission to handle `Equation::Scalar`, `Equation::ApplyToAll`, and `Equation::Arrayed` so INTEG initial values are never dropped for arrayed stocks
- emit type-1 flow pipe connector geometry from `flow.points`, including endpoint attachment connectors and intermediate waypoints
- add targeted regression tests covering infinity formatting, view title roundtrip, arrayed stock INTEG reconstruction, and flow pipe connector serialization

## Testing
- `cargo test -p simlin-engine mdl::writer::tests::`
- `cargo test -p simlin-engine mdl::view::convert::tests::test_build_single_view`
- `cargo test -p simlin-engine json_proptest::`
- `cargo test -p simlin-engine --features file_io --test simulate simulates_models_correctly`
- full pre-commit hook suite via `git commit`

Fixes #325
Fixes #326
Fixes #328
Fixes #329

### DIFF
--- a/src/simlin-engine/src/mdl/view/convert.rs
+++ b/src/simlin-engine/src/mdl/view/convert.rs
@@ -11,8 +11,9 @@ use crate::datamodel::{self, View, ViewElement, view_element};
 use std::collections::HashSet;
 
 use super::processing::{
-    EffectiveGhosts, PrimaryMap, angle_from_points, associate_variables, compose_views,
-    is_cloud_endpoint, xmile_angle_to_canvas,
+    EffectiveGhosts, PrimaryMap, angle_from_points, associate_variables,
+    build_attached_valve_flow_maps, compose_views, is_cloud_endpoint, resolve_flow_uid_for_valve,
+    xmile_angle_to_canvas,
 };
 use super::types::{VensimComment, VensimElement, VensimVariable, VensimView};
 
@@ -286,11 +287,12 @@ fn convert_view(
 ) -> Option<View> {
     let mut elements = Vec::new();
     let uid_offset = view.uid_offset;
+    let (valve_to_flow, flow_to_valve) = build_attached_valve_flow_maps(view);
 
     // Track which comments are clouds (flow endpoints)
     let mut cloud_comments: HashMap<i32, i32> = HashMap::new(); // comment_uid -> flow_uid
     for (uid, _elem) in view.iter_with_uids() {
-        if let Some(flow_uid) = is_cloud_endpoint(uid, view) {
+        if let Some(flow_uid) = is_cloud_endpoint(uid, view, &valve_to_flow) {
             cloud_comments.insert(uid, flow_uid);
         }
     }
@@ -324,6 +326,7 @@ fn convert_view(
                     view_idx,
                     uid_offset,
                     view_offsets,
+                    &flow_to_valve,
                 ) {
                     if matches!(&view_elem, ViewElement::Flow(_)) {
                         emitted_flow_uids.insert(uid);
@@ -342,7 +345,9 @@ fn convert_view(
                 // Non-cloud comments are ignored
             }
             VensimElement::Connector(conn) => {
-                if let Some(link) = convert_connector(conn, uid, view, uid_offset, symbols) {
+                if let Some(link) =
+                    convert_connector(conn, uid, view, uid_offset, symbols, &valve_to_flow)
+                {
                     elements.push(link);
                 }
             }
@@ -407,6 +412,7 @@ fn convert_variable(
     view_idx: usize,
     uid_offset: i32,
     view_offsets: &[i32],
+    flow_to_valve: &HashMap<i32, i32>,
 ) -> Option<ViewElement> {
     let canonical = to_lower_space(&var.name);
 
@@ -470,7 +476,8 @@ fn convert_variable(
         })),
         VariableType::Flow => {
             // For flows, find the associated valve and compute flow points
-            let (flow_x, flow_y, points) = compute_flow_data(var, view, uid_offset, symbols);
+            let (flow_x, flow_y, points) =
+                compute_flow_data(var, view, uid_offset, symbols, flow_to_valve);
 
             Some(ViewElement::Flow(view_element::Flow {
                 name: xmile_name,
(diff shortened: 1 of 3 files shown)


bpowers:
Addressed the latest Codex feedback around MDL valve/flow roundtripping by removing the strict `valve_uid + 1` assumption during import. The MDL view conversion path now builds a per-view attached valve↔flow mapping and reuses it consistently for cloud endpoint detection, connector valve indirection, and flow endpoint reconstruction. This keeps legacy adjacency behavior as a fallback while correctly handling writer-emitted non-adjacent valve IDs.

I also added regression coverage at two levels: an end-to-end `project_to_mdl -> parse_mdl` roundtrip test in `mdl/writer.rs` and a conversion-level test in `mdl/view/convert.rs` for non-adjacent attached valves. Review iteration count for this local feedback cycle: 1.

Issue #364, opened 2026-03-01
engine: array-producing builtins nested inside other BuiltinFn calls produce incorrect results in interpreter


bpowers:
## Problem

Array-producing builtins (VECTOR ELM MAP, VECTOR SORT ORDER, ALLOCATE AVAILABLE) produce incorrect results in the interpreter when nested inside other BuiltinFn calls. For example:

- `MAX(VECTOR ELM MAP(source[*], offsets[*]), 15)` returns 15 for element 0 instead of the correct value 30 (when VEM should produce 30 for that element).
- `SUM(VECTOR ELM MAP(...))` returns NaN.

The root cause appears to be how array-producing builtins interact with the per-element evaluation context when wrapped inside another BuiltinFn. The outer builtin evaluates the inner array-producing builtin in a context where the per-element iteration state is not correctly propagated, causing the inner builtin to return wrong values or NaN rather than the correctly mapped element.

This is a pre-existing limitation, not introduced by PR #361 (VM vector operations).

## Why it matters

- **Correctness**: Models that compose array-producing builtins inside scalar builtins (a common Vensim/Stella pattern) will silently produce wrong simulation results.
- **Developer experience**: The failure mode is silent -- wrong values rather than errors -- making it difficult to diagnose.
- **Model compatibility**: Real-world SD models frequently nest these operations (e.g., `MAX(VEM(...), threshold)`, `SUM(VEM(...))`, `MIN(ALLOCATE_AVAILABLE(...), cap)`).

## Component(s) affected

- `src/simlin-engine/src/interpreter.rs` (per-element evaluation context for array-producing builtins)
- Potentially also relevant to VM implementation once #355 is resolved, since the VM will need to handle the same composition patterns.

## Possible approaches

1. Ensure array-producing builtins fully materialize their result array before the outer builtin evaluates, so the outer builtin sees the correct per-element value rather than an incomplete or context-dependent intermediate.
2. Propagate the per-element iteration index through nested builtin calls so inner array-producing builtins know which element they should compute.
3. Review how the interpreter's `calc_eval` / element iteration interacts with `BuiltinFn::Vem`, `BuiltinFn::Vso`, and `BuiltinFn::AllocateAvailable` when they appear as arguments to another `BuiltinFn`.

## Context

Identified during code review of PR #361 (VM vector operations). The issue is pre-existing in the interpreter and was surfaced by testing nested compositions of array-producing builtins.

PR #368, opened 2026-03-01
engine: fix nested array-producing builtin reductions in A2A

bpowers:
## Summary
This PR fixes incorrect interpreter behavior when array-producing builtins are nested inside other builtins in A2A contexts.

The root cause was nested hoisting always rewriting array-producing builtins to `TempArrayElement`, even when the parent builtin expected an array argument (for example `SUM(...)`). That collapsed full-array semantics into element-local reads.

## Changes
- Made nested builtin rewriting argument-mode aware (`ScalarElement` vs `ArrayValue`).
- Keep `TempArray` for array-consuming argument positions and `TempArrayElement` for scalar-consuming positions.
- Added regression tests for nested:
  - `SUM(VECTOR ELM MAP(...))` in array context
  - `SUM(VECTOR SORT ORDER(...))` in array context
  - `SUM(ALLOCATE AVAILABLE(...))` in array context

## Validation
- `cargo test -p simlin-engine --test compiler_vector`

### DIFF
--- a/src/simlin-engine/src/compiler/mod.rs
+++ b/src/simlin-engine/src/compiler/mod.rs
@@ -995,6 +995,15 @@ fn builtin_contains_array_producing(builtin: &BuiltinFn) -> bool {
 /// handling the case where nested builtins operate on different dimensions.
 /// On the first call (element 0), collects the hoisted AssignTemp expressions.
 /// On subsequent calls, only performs the replacement using the same temp IDs.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum NestedBuiltinArgMode {
+    /// Expression is consumed as a scalar for the current A2A element.
+    ScalarElement,
+    /// Expression is consumed as an array value (e.g., SUM arg) and must keep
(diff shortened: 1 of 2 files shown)


bpowers:
This change set fixes the root cause behind issue #364 by preserving full-array semantics when nested array-producing builtins are hoisted in A2A lowering. Previously, nested replacements always became element reads, which caused reductions like `SUM(...)` over `VECTOR ELM MAP`, `VECTOR SORT ORDER`, and `ALLOCATE AVAILABLE` to collapse incorrectly to per-element values. The compiler now distinguishes scalar-consuming vs array-consuming argument positions so array reducers receive full temp arrays.

I also added focused regression coverage in `tests/compiler_vector.rs` for these nested reduction patterns in array context, plus sanity checks for nested max/sum behavior. After the update, the `simlin-engine` compiler-vector test suite is green (`cargo test -p simlin-engine --test compiler_vector`). This review cycle completed in 1 Codex iteration (Claude sub-review intentionally skipped per request).

Issue #366, opened 2026-03-01
engine: SUM(VECTOR ELM MAP(...)) in A2A context evaluates per-element rather than reducing the full array


bpowers:
## Problem

When an array-producing builtin like `VECTOR ELM MAP` is wrapped in a reducer like `SUM()` and the entire expression appears in an A2A (apply-to-all) context, the hoisting logic treats the inner array-producing builtin as the primary expression shape. This causes the outer `SUM()` to evaluate per-element rather than first materializing the full array and then reducing it.

For example, with `result[DimA] = SUM(VECTOR ELM MAP(source[*], offsets[*]))`:
- Expected: each element of `result` gets the same scalar value (the sum of all mapped elements)
- Actual: the A2A hoisting detects the inner VEM and hoists it, but the outer SUM is applied per-element rather than over the full materialized array

## Relationship to other issues

- Overlaps with #364 (array-producing builtins nested inside other BuiltinFn calls in interpreter), but this issue is specifically about the A2A expansion path in the compiler
- Related to #365 (preserve_wildcards conflation), since the same flag controls both reducer and vector builtin dimension preservation

## Context

Identified during code review of PR #361 (VM vector operations). Referenced in the PR summary as a deferred pre-existing issue but not previously filed. The issue predates the PR and affects both interpreter and VM paths.

## Component(s) affected

- `src/simlin-engine/src/compiler/mod.rs` (A2A hoisting detection and expansion)
- `src/simlin-engine/src/compiler/context.rs` (preserve_wildcards_for_iteration semantics)

bpowers:
Fixed in commit bb3e9771 (PR #368) — A2A hoisting now correctly distinguishes array-value and scalar-element arguments for reducer-wrapped array-producing builtins.

PR #369, opened 2026-03-01
engine: hoist nested vector builtins for scalar reducers

### LINKED ISSUE #367: engine(vm): scalar SUM(VECTOR ELM MAP(...)) fails VM compilation
## Problem
A scalar reduction over an array-producing builtin fails VM compilation:

`result = SUM(VECTOR ELM MAP(source[*], offsets[*]))`

Interpreter evaluation succeeds, but VM compilation fails with:

`Cannot push view for expression type Discriminant(7) - expected array expression`

## Repro
Use a simple model with:
- `source[D] = [10,20,30]`
- `offsets[D] = [2,0,1]`
- `result = SUM(VECTOR ELM MAP(source[*], offsets[*]))`

Then run VM compilation (`Simulation::compile()` path).

## Expected
VM should compile and produce `result = 60` (matching interpreter).

## Actual
VM compilation errors before simulation starts.

## Notes
Discovered while working on issue #364 (nested array-producing builtins in A2A context). This appears to be a separate VM/codegen limitation for scalar reductions over array-producing builtin calls.

bpowers:
## Summary
- hoist nested array-producing builtins in scalar equations before assignment
- rewrite scalar nested vector-builtins using `AssignTemp + TempArray` so reducers consume full arrays
- add VM regression tests for reducer wrappers around `vector_elm_map`, `vector_sort_order`, and `allocate_available`

## Root Cause
Scalar lowering (`Ast::Scalar`) did not run nested array-builtin hoisting. That left expressions like `sum(vector_elm_map(...))` as direct builtin apps inside reducer args, and VM codegen cannot push views for those app nodes.

A2A nested-hoisting logic already handled array-value vs scalar-element argument modes, but test coverage only exercised interpreter behavior for reducer-wrapped builtins in array contexts.

## Validation
- `cargo test -p simlin-engine --test compiler_vector -- --nocapture`
- `cargo test -p simlin-engine --lib array_tests::vector_elm_map_tests -- --nocapture`
- pre-commit hook (fmt, clippy, rust tests, ts checks, pysimlin tests)

Closes #367
Related to #366

### DIFF
--- a/src/simlin-engine/src/compiler/mod.rs
+++ b/src/simlin-engine/src/compiler/mod.rs
@@ -752,10 +752,10 @@ impl Var {
     }
 }
 
-/// For scalar equations, rewrite nested array-producing builtins into
-/// AssignTemp + TempArray so reducers (SUM/MEAN/etc.) consume full arrays.
-/// This mirrors A2A nested hoisting semantics without introducing element
-/// projection in non-A2A contexts.
+/// For scalar equations, hoist nested array-producing builtins only where the
+/// parent expects an array value (reducers/vector array args). Scalar-argument
+/// positions are left unchanged so existing structured compile errors are
+/// preserved instead of forcing scalar-element rewrites.
 fn hoist_nested_array_builtins_in_scalar(main_expr: Expr, exprs: &mut Vec<Expr>) -> Expr {
     if is_array_producing_builtin(&main_expr) || !contains_array_producing_builtin(&main_expr) {
         return main_expr;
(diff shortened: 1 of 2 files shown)


bpowers:
Completed one local automated review iteration with codex after opening this PR, and no additional actionable issues were identified relative to `origin/main`.

This branch keeps the compiler change focused on one root cause: scalar equations now hoist nested array-producing builtins into `AssignTemp + TempArray`, so reducer builtins consume full materialized arrays in VM codegen. The PR also adds VM regression coverage for reducer-wrapped `vector_elm_map`, `vector_sort_order`, and `allocate_available`, including the A2A shape tied to #366, so this behavior is guarded against future regressions.

PR #389, opened 2026-03-11
engine: make previous fully intrinsic

bpowers:
## Summary

- Replace the remaining stdlib-backed PREVIOUS behavior with a single intrinsic two-argument path
- Unary PREVIOUS now desugars to PREVIOUS(x, 0), module-backed and expression arguments rewrite through scalar helper auxes
- LTM first-step guards use INITIAL_TIME semantics instead of an implicit previous model
- Removes stdlib/previous.stmx and its generated registration
- Updates dependency classification so PREVIOUS fallback expressions behave like init-only dependencies

## Test plan

- [ ] All existing engine tests pass (cargo test)
- [ ] LTM tests updated and passing
- [ ] Interpreter/VM behavior aligned across direct, arrayed, SELF, module-backed, and nonzero-start cases

### DIFF
--- a/src/simlin-engine/src/db_ltm.rs
+++ b/src/simlin-engine/src/db_ltm.rs
@@ -788,9 +788,9 @@ pub(super) fn compile_ltm_equation_fragment(
     let inputs = BTreeSet::new();
     let mut module_models = model_module_map(db, model, project).clone();
 
-    // Merge LTM implicit module references from LTM equation parsing into
-    // the module_models map so the compiler
-    // context can resolve module_var_name -> sub_model_name lookups.
+    // Merge LTM implicit module references from LTM equation parsing into the
+    // module_models map so the compiler context can resolve module_var_name ->
+    // sub_model_name lookups.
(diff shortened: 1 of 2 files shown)


bpowers:
## Review cycle summary

Three review iterations with both codex and Claude reviewers. The initial commit was clean and well-structured; reviews caught two real bugs and prompted several improvements:

**Bugs fixed:** LoadPrev stack effect metadata was (0,1) instead of (1,1) since the opcode now pops a fallback. Unit inference for PREVIOUS was returning units from the fallback argument instead of the lagged argument, causing variables using unary PREVIOUS(x) to lose their inferred units. Both fixed with regression tests.

**Improvements:** Added equality constraint between PREVIOUS's lagged argument and fallback in unit inference (matching Max/Min handling), so incompatible fallback units are caught during inference rather than only at check time. Removed the dead `_model_name` parameter from `classify_module_for_ltm`. Added explanatory comments for the interpreter's non-Var fallback invariant and the dual desugaring in expr1.rs.

PR #435, opened 2026-04-03
engine: add RK2 and RK4 integration methods

bpowers:
## Summary

- Adds RK2 (Heun's method) and RK4 (classical Runge-Kutta) integration to the simlin-engine VM, as required by the XMILE spec
- Reuses existing Euler stock bytecode: derivative extraction via `next[off] - curr[off]` works at any trial point, requiring zero compiler changes
- Scratch buffers allocated once in `Vm::new()` (2*N_stocks floats), split via `split_at_mut` in the loop -- zero per-timestep allocation
- Stock offsets collected by scanning stock-phase bytecode, recursing into submodule `EvalModule` opcodes (covers SMOOTH/DELAY internals)
- After RK stages, flows are re-evaluated with restored state for correct output values and PREVIOUS() snapshot timing

## Test plan

- [x] Harmonic oscillator: Euler energy explodes (>5e9), RK4 conserves (<1.01)
- [x] Exponential decay: RK4 relative error <0.001% at dt=1.0
- [x] Convergence order: Euler ~2x, RK2 ~4x, RK4 ~16x when halving dt
- [x] PREVIOUS() and INIT() correct with RK methods
- [x] reset() produces bitwise-identical results
- [x] save_step != dt works correctly
- [x] Partial run_to + continue matches full run
- [x] All 2890 existing tests pass (zero regression)

### DIFF
--- a/src/simlin-engine/src/rk_integration_tests.rs
+++ b/src/simlin-engine/src/rk_integration_tests.rs
@@ -502,6 +502,35 @@ fn rk4_with_init_builtin() {
 // ── PREVIOUS fallback during initial timestep ─────────────────────
 
 #[test]
+fn euler_previous_fallback_at_initial_timestep() {
+    // Verify Euler still uses the fallback at t=0 after the LoadPrev
+    // refactor (use_prev_fallback flag replaces TIME == INITIAL_TIME check).
+    let tp = TestProject::new("euler_prev_fallback")
+        .with_sim_time(0.0, 2.0, 1.0)
+        .with_sim_method(datamodel::SimMethod::Euler)
+        .flow("inflow", "PREVIOUS(s, 99)", None)
+        .stock("s", "0", &["inflow"], &[], None);
+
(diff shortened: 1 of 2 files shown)


bpowers:
## Review cycle summary

Two review iterations were needed. The first iteration caught a P1 correctness bug: `PREVIOUS(x, fallback)` returned 0 instead of the explicit fallback during RK intermediate stages of the initial timestep. The root cause was that RK stages advance `curr[TIME]` to trial points, causing `LoadPrev`'s `TIME == INITIAL_TIME` guard to fail before `prev_values` was populated. Fixed by adding a `use_prev_fallback` flag tracked in `Vm::prev_values_valid` across segmented `run_to()` calls.

The first iteration also added a SMOOTH submodule stock test (with time-varying input to create real dynamics), defensive sort+dedup on `collect_stock_offsets`, and minor cleanups. The second iteration had both reviewers report no actionable issues.

PR #451, opened 2026-04-09
engine: include referenced stdlib models in project datamodel

bpowers:
## Summary

- Fixes two bugs in the diagram editor where modules referencing stdlib models (e.g. in the hiring systems model) couldn't display their model name or be drilled into
- Adds `Project::ensure_referenced_stdlib_models()` to the Rust engine which scans module variables for stdlib references and enriches the project datamodel with current definitions
- Updates TypeScript `getAvailableModels` to properly categorize stdlib models in the selector dropdown
- Comprehensive test coverage across Rust unit tests, FFI integration tests, and TypeScript unit tests

## Test plan

- [x] Rust unit tests for enrichment function (add, idempotent, skip user models, dedup, multiple)
- [x] Rust integration test loading hiring model and verifying stdlib models in serialized JSON
- [x] TypeScript tests for stdlib categorization and wiring panel access
- [x] All pre-commit checks pass (fmt, clippy, cargo test, pnpm lint, pnpm test, pysimlin)

### DIFF
--- a/src/diagram/Editor.tsx
+++ b/src/diagram/Editor.tsx
@@ -1609,8 +1609,6 @@ export class Editor extends React.PureComponent<EditorProps, EditorState> {
     if (!view) {
       return;
     }
-    // Don't drill into models that aren't materialized in the project
-    // (e.g. stdlib models that are internal to the engine).
     const project = this.project();
     if (!project || !project.models.has(targetModelName)) {
       return;
--- a/src/diagram/Editor.tsx
+++ b/src/diagram/Editor.tsx
@@ -1613,8 +1613,8 @@ export class Editor extends React.PureComponent<EditorProps, EditorState> {
       return;
     }
     // Guard: don't push a nonexistent model onto the navigation stack.
-    // Stdlib models are included in project.models by the engine's
-    // ensure_referenced_stdlib_models() enrichment.
+    // Stdlib models are included in project.models because the editor
+    // passes includeStdlib=true when calling serializeJson().
     const project = this.project();
     if (!project || !project.models.has(targetModelName)) {
       return;
--- a/src/diagram/Editor.tsx
+++ b/src/diagram/Editor.tsx
@@ -2537,7 +2537,7 @@ export class Editor extends React.PureComponent<EditorProps, EditorState> {
 
     const serializedProject = await engine.serializeProtobuf();
 
-    const json = JSON.parse(await engine.serializeJson()) as JsonProject;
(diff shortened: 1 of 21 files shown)


bpowers:
## Review Summary

This PR went through 4 iterations of automated review (codex + claude).

The initial implementation added stdlib model enrichment to the engine datamodel and fixed the TypeScript module details and drill-in handling. Reviews identified three areas for improvement: (1) the enrichment function was unconditionally replacing existing models, which would silently overwrite user-defined models that shadow stdlib names -- fixed by switching to add-only with an explicit preservation test; (2) the stdlib model selector was empty for first-time insertion because it only drew from materialized project models -- fixed by populating from the full STDLIB_MODEL_NAMES registry; (3) unreferenced stdlib models were never cleaned up after module deletion -- fixed by adding a prune step that removes stale definitions while preserving user models.

Additional polish from the reviews: consolidated the duplicated `STDLIB_PREFIX` constant, added cycle checking for stdlib shadow models in the selector, documented the drill-in guard and prune design decisions, and expanded test coverage to 9 Rust unit tests + 1 FFI integration test + 6 TypeScript tests covering all identified edge cases.

bpowers:
## Review Summary (iterations 5-8)

This review cycle addressed two significant correctness issues and several refinements identified by automated reviewers.

The most important fix was moving stdlib model enrichment from the in-memory datamodel into the JSON serialization path with an explicit `include_stdlib` flag. The original approach enriched the datamodel at project-open time, which meant stdlib models would be persisted to Firestore/disk on save. The new approach threads an `include_stdlib` boolean through the full FFI stack (Rust -> C header -> WASM -> TypeScript backends -> Python CFFI) so the Editor can request enriched JSON for display while all save paths get clean output. Additionally, the `includeStdlib=true` flag is now passed in all three editor JSON deserialization paths (updateProject, openInitialProject, and openEngineProject) so stdlib models are available on first render and after undo/redo, not just after the first edit.

A second fix addressed bare-name misclassification: the `isStdlibModel()` predicate matches both prefixed names like `stdlib⁚delay1` and bare names like `delay1`, which would incorrectly classify user-created models with common names. The module selector and details panel now use `startsWith(STDLIB_PREFIX)` for classification, ensuring only engine-owned stdlib models (with the Unicode TWO DOT PUNCTUATION prefix) are treated as read-only. The search bar's variable details panel is also now suppressed when inside a stdlib model to prevent patch errors against the non-materialized engine datamodel.

Issue #23, opened 2020-09-19
lower `xmile` structs into an intermediate representation, use that everywhere


bpowers:
I believe we want to end up with 3 wasm-ified libraries:
- *system-dynamics-compat:* import (and eventually export) projects to XMILE + Vensim mdl formats.  These will output a protobuf representation of models
- *system-dynamics-engine:* load, check, and edit projects.  Works on protobuf inputs (no dependencies on e.g. XML library)
- *system-dynamics-simulation:* simulate models, runs in a web worker.  We can keep the core engine on the main thread to simplify the concurrency story for editing, and the engine can generate bytecode and spit it over to the `simulation` piece to run it in a worker.

bpowers:
done

PR #459, opened 2026-04-18
engine: reduce LTM circuit-enumeration memory and time on dense graphs

bpowers:
## Summary

Five-commit stack that takes wrld3-03 LTM circuit enumeration from 8.45 GiB / 27 s to 491 MiB / 1.2 s (17x smaller, 22x faster) while preserving all existing LTM semantics. Production `MAX_LTM_CIRCUITS=100_000` path drops from 322 MiB / 569 ms to 33 MiB / ~29 ms.

| Commit | What it does | wrld3-03 uncapped impact |
|---|---|---|
| 7303af3d | SCC-restricted indexed (u32) DFS, skip trivial SCCs | 8.45 GiB → 5.87 GiB, 26.8 s → 2.87 s |
| 2fd4a8f5 | Indexed `LoopCircuitsResult` (names table + `Vec<Vec<u32>>`) | 5.87 GiB → 914 MiB, 2.87 s → 1.73 s |
| 7334e1dd | Hash-fingerprint dedup keys | 914 MiB → 491 MiB, 1.73 s → 1.49 s |
| aa56c5a1 | Johnson 1975 with blocked-set unblocking | 491 MiB (unchanged), 1.49 s → 1.23 s |
| c0e9aa96 | Tech-debt items from LTM critical review | -- |

## Test plan

- [x] `cargo test -p simlin-engine --lib` (3016 passed)
- [x] `cargo test -p simlin-engine --features file_io --test simulate_ltm` (45 passed)
- [x] `cargo test -p simlin-engine --test layout` (includes `test_layout_arms_race`, 31 passed)
- [x] `cargo test -p simlin-engine --test wrld3_ltm_panic` (60 s budget, passes comfortably)
- [x] Johnson-vs-Tiernan equivalence test: fixture corpus + 256-case proptest (both agree)
- [x] Peak RSS measured via `ltm_mem_bench` example on wrld3-03 at budget 100K and unbounded

### DIFF
--- a/docs/design-plans/2026-04-17-ltm-mem-reduction.md
+++ b/docs/design-plans/2026-04-17-ltm-mem-reduction.md
@@ -1,0 +1,115 @@
+# LTM Circuit-Enumeration Memory Reduction
+
+Status: in progress on branch `reduce-ltm-mem`.
+Baseline captured 2026-04-17.
+
+## Problem
+
+`CausalGraph::find_circuit_node_lists` enumerates elementary circuits via
+lexicographic-small-start DFS over the causal graph. On dense SD models this
+blows up in both time and memory:
+
+* test/metasd/WRLD3-03/wrld3-03.mdl has **1,863,803** deduplicated elementary
+  circuits (mean length 47 nodes, max 80).
+* Full enumeration consumes **8.45 GiB peak RSS** and **27 s wall-clock** on
+  reduce-ltm-mem@d1870d2e (baseline commit for this work). WASM's 4 GiB linear
+  memory limit forces the existing `MAX_LTM_CIRCUITS = 100_000` budget, which
+  itself already consumes ~322 MiB.
+* Per-circuit cost: ~4.5 KiB (Vec<Ident<Canonical>> with ~47 heap-allocated
+  String clones per circuit, plus the dedup HashSet<String> key which is
+  string-joined node names).
+
+## Graph Characterization (wrld3-03)
+
+* 302 nodes with outgoing edges, 313 nodes overall
+* 483 total out-edges
+* 1 weakly-connected component (size 313)
+* **148 strongly-connected components. Largest = 166 nodes.**
+* **Exactly 1 non-trivial SCC (size > 1).** 147 trivial SCCs carry zero cycles.
+* 15 stocks, all in the single non-trivial SCC.
+
+The `166`-node SCC is where every loop lives. Enumeration across all 302
+start-nodes spends the majority of its work inside this SCC; the 147 nodes
+outside it never close a cycle but are still explored when their successor
+edges lead into the SCC.
+
+## System-Dynamics Domain Notes
+
+* Directed edges: flow→stock (structural), stock→flow / aux→aux / stock→aux
+  / aux→flow (from equation dependencies). Stock init-value dependencies are
+  intentionally excluded from the cycle graph.
+* A well-formed SD cycle contains at least one stock: stocks provide the
+  DT-delay that makes the feedback solvable. A pure auxiliary cycle would
+  be an unsolvable algebraic loop and is rejected elsewhere.
+* The DT delay is a semantic property of how the simulator integrates the
+  stock, not a structural property of the graph. For the purposes of cycle
+  enumeration the directed graph is the complete ground truth; Johnson's
+  algorithm on it is both necessary and sufficient.
+* SD models tend to have long cycles (many auxes between stocks). wrld3's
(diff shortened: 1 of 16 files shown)


bpowers:
## Review-cycle summary

Six automated-review iterations over this branch; both reviewers (codex-review and the /review skill) landed clean in the same iteration at `df2f240c`.

Post-review improvements driven by reviewer feedback, beyond the original stack:

- **Hash determinism and correctness.** The initial dedup used `std::hash::RandomState` (process-seeded), which meant the `LoopCircuitsResult` salsa cache value could silently reshuffle after the rare 2^64 collision. Ported a narrow rapidhash V3 slice (HashMicro variant) to `src/simlin-engine/src/rapidhash.rs`, cross-validated byte-for-byte against the Go reference port, with a compile-error guard against big-endian targets. All dedup fingerprints are now deterministic across runs; 3.5x faster throughput at LTM's 188-byte mean input size than the interim FNV-1a.

- **Dedup moved into per-SCC enumeration.** Previously dedup ran as a post-hoc pass over all emitted circuits, which meant peak transient memory on dense multidigraph inputs was proportional to raw cycle count rather than unique output. Folding the dedup into `johnson_circuit` keeps `state.circuits` proportional to unique node-sets.

- **Name table trimmed to cycle-participants only.** `LoopCircuitsResult.names` used to include every graph node, so renaming an acyclic feeder invalidated the salsa cache. Now compacted to only the nodes that appear in at least one circuit; `find_indexed_circuits_trims_names_to_cycle_participants` regression-tests the invariant.

- **Budget correctly charges on every raw cycle.** The intermediate implementation only decremented `max_circuits` for unique-node-set survivors, which let dense multidigraphs run DFS past the cap (K9: 125k raw cycles but only 502 unique node-sets). New `johnson_budget_charges_duplicate_raw_cycles` test pins K4 as a minimal fixture.

- **Canonical multidigraph representative pinned.** The pre-refactor code iterated successors in HashMap-insertion order, so polarity on multidigraph cycles was non-deterministic across processes. Sorting successors lex-ascending converts that into a canonical "lex-smallest sequence" guarantee; `johnson_canonical_multidigraph_representative` documents and pins the invariant with an explicit K3 assertion.

- **Small hygiene.** Dropped the unused `_is_affecting_stock` flag in `ltm_augment.rs`, tightened the bench's root-model lookup to be deterministic on multi-model projects, removed the now-redundant Loop-level `deduplicate_loops` call (replaced with a `debug_assert!` invariant check), and fixed several stale doc comments that referenced removed helpers or the wrong tech-debt item.

Net wrld3-03 impact unchanged from the opening numbers (17x memory, 22x wall-time at uncapped; the production 100K-cap path at ~33 MiB / ~29 ms). The six review iterations tightened correctness, determinism, and observability around those numbers without changing them.

bpowers:
## Address-feedback iteration 6b: strip incidental determinism

Triaged every comment/review on this PR.  The P3 nits (updated `state.seen` rustdoc, FNV-1a phrasing, `find_indexed_circuits_with_limit` debug_assert, tech-debt #23 resolution note, `from_edges` non-determinism note) are addressed in `e771dc27`.

The same commit also takes the user's explicit "embrace non-determinism unless correctness demands it" guidance and rolls back the successor-list sort in `from_edges`, the inner-SCC sort in `tarjan_scc`, and the outer SCC-iteration sort in `enumerate_indexed_circuits`.  All three were pure incidental determinism: no production consumer requires the specific order they pinned, and only one test (`johnson_canonical_multidigraph_representative`, which this PR itself added) relied on them.  That test is replaced with `johnson_multidigraph_dedups_to_single_node_set`, which asserts the correctness-relevant property (K3's two 3-cycles fold into one circuit with node-set {a,b,c}) without pinning a rotation.

Within-process determinism is still guaranteed (HashMap iteration is stable for a given instance), so `LoopCircuitsResult` salsa caching continues to work.  Cross-process rotation on multidigraph cycles now matches the pre-refactor behaviour -- which was already non-deterministic -- but the new code is organized to make the non-determinism explicit in code comments rather than incidental.

The one deferred item -- sizing `JohnsonState.blocked` / `b_list` to the current SCC's node count rather than the whole graph's -- was filed as issue #460 for a follow-up.  It's a real allocation win on hypothetical graphs with many small SCCs in a large node space, but wrld3-03's single 166-node SCC makes it a no-op in practice.

wrld3-03 bench unchanged within noise: 100K-cap ~32 MiB / ~50 ms, uncapped ~467 MiB / ~1.47 s.

PR #461, opened 2026-04-19
engine: enable LTM always-on via auto-flip + post-sim rel_loop_score

bpowers:
## Summary

- Remove the `MAX_LTM_CIRCUITS = 100_000` safety cap and replace it with a principled structural gate: `MAX_LTM_SCC_NODES = 50` triggers auto-flip to discovery mode when any element-level SCC exceeds the threshold, with a Warning diagnostic surfaced through `model_all_diagnostics`.
- Move `rel_loop_score` from compile-time synthetic equations to post-simulation computation (`ltm_post::compute_rel_loop_scores`). Eliminates the O(P^2) equation-text blowup that previously projected to ~140 TB on WRLD3.
- WRLD3 (1,863,803 elementary circuits, one 166-node SCC) now compiles with LTM enabled in ~1.6s at 467 MiB VmPeak, down from OOM/SIGKILL in >15 GiB.

## Why

Simlin's WASM target has a 4 GiB linear-memory ceiling. Running LTM on WRLD3 with the previous pipeline needed on the order of 15-60+ GiB of intermediate state:
- `build_element_level_loops` materialized 1.86M Loop structs (mean 47 Link entries each) for ~17 GiB by itself.
- `rel_loop_score` equation text was O(P^2) per partition — 75 MB per equation times 1.86M equations ~= 140 TB projected.

The 100k circuit cap hid this by returning empty on budget exhaustion, so raising it exposed the downstream cliffs. See `docs/design-plans/2026-04-18-ltm-cap-lift-diagnosis.md` for the measured trace and `docs/design-plans/2026-04-18-ltm-cap-lift-options.md` for the design rationale.

## How

1. **Auto-flip gate** (`src/simlin-engine/src/db_ltm.rs`): compute the largest element-level SCC before calling `build_element_level_loops`. If it exceeds `MAX_LTM_SCC_NODES = 50`, take the `is_discovery` branch internally and emit a Warning through the `CompilationDiagnostic` accumulator. The paper's prescribed two-tier strategy; Simlin had both modes but no automatic switch.
2. **Post-sim rel_loop_score** (`src/simlin-engine/src/ltm_post.rs`): `compute_rel_loop_scores(results, loop_partitions)` normalizes `loop_score` timeseries against the sum-of-absolutes per partition. `LtmVariablesResult::loop_partitions` caches the loop-to-partition mapping. Two consumers migrated: `libsimlin::simlin_analyze_get_relative_loop_score` FFI and `layout/mod.rs` feedback-loop metadata.
3. **Cap removal** (`src/simlin-engine/src/ltm.rs`): `MAX_LTM_CIRCUITS`, `RUNTIME_LTM_CIRCUIT_BUDGET`, `set_max_ltm_circuits`, `default_max_ltm_circuits`, `current_ltm_circuit_budget`, and the three default-budget wrappers deleted. `TruncatedByBudget` and `*_with_limit` kept as the general form.
4. **Diagnostic visibility fix**: `model_all_diagnostics` did not invoke `model_ltm_variables`, so the auto-flip Warning was invisible to FFI callers. Fixed with a gated call + regression test.
5. **Corner tests** (adversarial): disjoint-SCC keying (largest, not sum), user-discovery suppression (no double-warning), element-level SCC sizing for A2A loops, NaN propagation contract in `compute_rel_loop_scores`.

## Test plan

- [x] `cargo test -p simlin-engine --release` -- 3045 engine tests pass
- [x] `cargo test --release --features file_io --test simulate_ltm` -- 45/45 including bit-for-bit preservation of the `test/logistic_growth_ltm/ltm_results.tsv` golden TSV
- [x] `cargo test --release --features file_io --test wrld3_ltm_panic` -- both tests pass in 1.63s total (was SIGKILL/OOM pre-fix)
- [x] `pnpm build` + `pnpm test` -- green via pre-commit hook on every commit
- [x] pysimlin tests pass (pre-commit hook)
- [x] WRLD3 end-to-end at cap=100k / 2M / usize::MAX -- all three produce identical output (483 link scores, auto-flip fires), peak 467 MiB, wall ~1.6s
- [x] `arrayed_population.stmx` (6 circuits, SCC small) stays exhaustive, 3.8ms / 10.5 MiB
- [x] `cross_element.stmx` (12 circuits, SCC small) stays exhaustive, 5.0ms / 10.5 MiB


bpowers:
Completed iterative review cycle with `/address-feedback`. The branch has now passed through iterations 16-20 with both codex and claude reviewers; each iteration's feedback was addressed in a single atomic commit (8957b146, f55ba73d, 3f6304a6, d46ba7fb).

High-level improvements layered on top of the original auto-flip + post-sim cap-lift work:

1. **Structural vs synthesis cap separation** -- `MAX_LTM_DETECTED_LOOPS = 100_000` gates structural consumers (layout, `simlin_analyze_get_loops`, `simlin-cli --ltm`) independently of the stricter `MAX_LTM_TOTAL_CIRCUITS = 10_000` that governs LTM equation synthesis.  Structural callers no longer lose topology information just because LTM synthesis would auto-flip.  The node-count safety cap was raised from 10M to 100M to keep the distinct-circuit cap as the primary trigger on WRLD3-shape models, with the node cap remaining as the backstop for long-loop pathologies.

2. **Bounded FFI memory** -- `simlin_analyze_get_relative_loop_score` previously materialized the full `loop_count * step_count` rel_score matrix on first call (~800 MB at the 10k-loop cap).  Replaced with per-cycle-partition denominator caching that bounds peak memory at `O(num_partitions * step_count)`, while still amortizing denominator work across the iterate-every-loop usage pattern.  The same streaming pattern now applies in `layout::try_detect_ltm_loops_incremental`, cutting transient layout memory by an equivalent factor.

3. **Persistent LTM diagnostics** -- `simlin_sim_new(enable_ltm=true)` resets the `ltm_enabled` flag after compile to avoid leaking LTM synthesis into later patch validation; that reset invalidated the salsa accumulator's LTM warnings and made `simlin_project_get_errors()` silent about models that auto-flipped.  LTM-specific diagnostics are now captured on `SimlinProject::pending_ltm_diagnostics` at sim_new time and replayed through `get_errors`, with per-(model, message) dedup so multi-model projects do not lose one model's warning to another's.  Cleared on `apply_patch` since patches can change which models trip the threshold.

4. **Determinism and test strength** -- The post-sim rel_loop_score helpers now sort loop IDs before float summation in both the full-pass and streamed paths, keeping denominator bits stable across salsa cache invalidations.  A new property-style test hashes with 16 fresh `RandomState` seeds to reliably catch a missing sort (verified to fail ~93% of the time with 8 trials, >99% with 16).

5. **Docstring and test hygiene** -- Refreshed stale rustdoc in `db_analysis.rs`, renamed stale cap references in test assertion messages, consolidated duplicated comment blocks in the FFI, added a node-cap unit test for `find_loops_if_under_limit_ex`, added FFI integration tests covering both the auto-flip warning surfacing path and its dedup semantics, and fixed `ltm_full_bench` to mark truncation explicitly on dense models instead of reporting `circuits=0`.

Tracked as tech debt for future follow-up: dedicated integration coverage for the `detected.truncated` fallback branch (#467), and preserving the removed compile-time emitter's exact summation order so the post-sim path is bit-for-bit identical rather than only deterministic (#468, sub-ULP drift below all existing test tolerances).

PR #489, opened 2026-05-07
engine: bound module pathway depth by graph node count

### LINKED ISSUE #312: ltm: fixed max depth 20 truncates module pathway enumeration
## Summary
Module pathway enumeration uses a hardcoded maximum depth of 20 nodes when finding input-port → output paths.

## Why this is a problem
Composite link scoring should consider all relevant simple pathways through a dynamic module. A fixed depth cap can truncate valid longer pathways, causing incomplete composite selection and incorrect boundary link scores.

Section 6 (composite link score) relies on evaluating pathway scores across the module’s internal causal pathways and selecting max-abs each timestep. Depth-based truncation is an implementation limit that can violate this for larger internal structures.

`enumerate_module_pathways()` calls `find_all_simple_paths(input_port, output_name, 20)`.

Location:
- `src/simlin-engine/src/ltm.rs` (`enumerate_module_pathways`)

## Impact
- Under-counted pathways in deeper modules.
- Potentially wrong composite scores and loop attribution for module-crossing loops.

## Success criteria
1. Remove hardcoded depth cap or derive it safely from graph size/topology (while preventing infinite recursion on cycles).
2. Ensure all simple input→output paths are considered up to a clearly documented, principled limit.
3. Add tests with an internal module path length >20 verifying those pathways are included.
4. Document any remaining practical limits and their behavioral consequences.

## Notes
This issue is unrelated to the known flow-to-stock timing offset.

bpowers:
## Summary

- `enumerate_module_pathways` previously called `find_all_simple_paths(input_port, output_name, 20)`, silently dropping internal pathways longer than 20 nodes in deeper user modules and producing incomplete composite link scores.
- Compute the node count of the module graph (union of adjacency-list keys and edge targets) once per call and pass it as the DFS depth cap.
- Added a unit test that builds a 22-node `input -> n0 -> ... -> n19 -> output` chain and verifies the pathway is enumerated end to end. The test fails on the previous code and passes with the fix.

## Why N is the correct bound

A simple path can visit each distinct node at most once, so the number of nodes in the graph is the principled upper bound on simple-path length. The visited set inside `dfs_simple_paths` already enforces simple-path semantics; the cap's only remaining job is recursion-stack safety, which N satisfies exactly.

This branch also includes a cherry-pick of `serve: clear inherited GIT_* env in git probe and tests` (already on `ltm/480-array-reducer-polarity`) so the project pre-commit hook can run `cargo test` without recursive invocation triggered by the `simlin-serve` git probe tests. Without this prep commit, the pre-commit hook fails on a test isolation issue unrelated to the LTM fix.

Fixes #312

### DIFF
--- a/src/simlin-engine/src/ltm.rs
+++ b/src/simlin-engine/src/ltm.rs
@@ -1547,6 +1547,21 @@ impl CausalGraph {
             }
         }
 
+        // The DFS depth cap is the principled upper bound on simple-path
+        // length: a simple path can visit each distinct node at most
+        // once, so N nodes -> at most N entries on the path stack. The
+        // visited set inside `dfs_simple_paths` already enforces simple-
+        // path semantics; the cap's only remaining job is recursion-stack
+        // safety, which N satisfies. The total node set is the union of
+        // the adjacency-list keys (sources) and `has_incoming` (targets),
+        // matching the convention in `detect_output_ports`.
+        let node_count = self
+            .edges
+            .keys()
+            .filter(|node| !has_incoming.contains(*node))
+            .count()
+            + has_incoming.len();
+
         for input_port in self.edges.keys() {
             if input_port == output_name {
                 continue;
@@ -1556,7 +1571,7 @@ impl CausalGraph {
                 continue;
             }
 
-            let raw_paths = self.find_all_simple_paths(input_port, output_name, 20);
+            let raw_paths = self.find_all_simple_paths(input_port, output_name, node_count);
             if raw_paths.is_empty() {
                 continue;
             }
@@ -4721,6 +4736,61 @@ mod tests {
             "Should find pathways with standard 'output' name"
         );
         assert!(pathways.contains_key(&Ident::new("input")));
+    }
+
+    #[test]
+    fn test_enumerate_module_pathways_deeper_than_legacy_cap() {
+        // Build a strictly-linear module graph deeper than the legacy
+        // hard-coded depth=20 truncation: input -> n0 -> n1 -> ... -> n19
+        // -> output, which is 22 distinct nodes and 21 edges.  A simple
+        // path can visit at most N distinct nodes (the visited set
+        // already enforces that), so the principled DFS cap is the
+        // module-graph node count.  A 21-element chain is enough to
+        // prove the old cap is gone without blowing past the per-test
+        // time budget.
+        const INTERMEDIATE_NODES: usize = 20;
+
+        let mut edges: HashMap<Ident<Canonical>, Vec<Ident<Canonical>>> = HashMap::new();
+
+        // input -> n0
+        edges.insert(Ident::new("input"), vec![Ident::new("n0")]);
+        // n_i -> n_{i+1}
+        for i in 0..INTERMEDIATE_NODES - 1 {
+            edges.insert(
+                Ident::new(&format!("n{i}")),
+                vec![Ident::new(&format!("n{}", i + 1))],
+            );
+        }
+        // n_{LAST} -> output
+        edges.insert(
+            Ident::new(&format!("n{}", INTERMEDIATE_NODES - 1)),
+            vec![Ident::new("output")],
+        );
+
+        let graph = CausalGraph {
+            edges,
+            stocks: HashSet::new(),
+            variables: HashMap::new(),
+            module_graphs: HashMap::new(),
+        };
+
+        let pathways = graph.enumerate_module_pathways(&Ident::new("output"));
+        let from_input = pathways.get(&Ident::new("input")).expect(
+            "input -> output pathway should be enumerated regardless of \
+             chain depth; a hardcoded depth-20 cap silently drops chains \
+             this long",
+        );
+        assert_eq!(
+            from_input.len(),
+            1,
+            "expected exactly one simple input -> output pathway"
+        );
+        // 22 nodes -> 21 directed links along the chain.
+        assert_eq!(
+            from_input[0].len(),
+            INTERMEDIATE_NODES + 1,
+            "pathway should traverse every link in the chain"
+        );
     }
 
     /// Construct a three-node reinforcing cycle (A → B → C → A) for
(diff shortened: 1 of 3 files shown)


bpowers (commented):
Adversarial review of the issue #312 fix. Verified locally on `ltm/312-module-pathway-depth` (HEAD `566c0f2e`).

## Verification evidence

- `cargo build --workspace` -> clean.
- `cargo clippy -p simlin-engine --no-deps` -> clean.
- `cargo test -p simlin-engine` -> 3156 unit tests + integration tests pass; new `ltm::tests::test_enumerate_module_pathways_deeper_than_legacy_cap` passes in <0.01s.
- Test reverse-validation: locally restored the legacy `find_all_simple_paths(input_port, output_name, 20)` literal and re-ran the new test. It panics on `expect()` with the documented diagnostic, confirming the assertion is meaningful and not a tautology.
- Cherry-pick `323744d7` byte-identical to original `e656d0a9` on `ltm/480-array-reducer-polarity` (verified by diffing `git show` output for the two commits; only the SHA line differs).

## Correctness of the bound (Critical: none)

The new bound is correct and sufficient. Walking through `dfs_simple_paths` (line 1503): `path` is initialized to `[from]`, the visited set excludes `target`, and the recursion check `path.len() > max_depth` runs at function entry. The deepest reachable `path.len()` before adding the target is `N - 1` (every non-target node), so any `max_depth >= N - 1` avoids truncation. The implementer chose `N`, which is one frame looser than the tightest possible bound but trivially safe and easier to defend in the comment ("a simple path can visit each distinct node at most once"). Not a concern.

The node-count formula `|edges.keys() \ has_incoming| + |has_incoming|` is mathematically equal to `|sources union targets|`, the size of the actual node set. Reusing the already-built `has_incoming` set avoids a second `HashSet` allocation, so this is slightly leaner than the `chain(...).collect::<HashSet<_>>().len()` style used in `detect_output_ports`. The comment claims it "matches the convention in `detect_output_ports`" -- the *result* matches but the *implementation* differs; the wording is slightly imprecise but not misleading.

I confirmed the bound works on the test fixture: 21 sources, 21 targets, sources \\ targets = {input}, so node_count = 1 + 21 = 22. With max_depth=22 the deepest dfs entry has path.len()=21 (`21 > 22` false), so the chain is enumerated end to end. With the legacy max_depth=20, the dfs entry at path.len()=21 trips `21 > 20` and returns before reaching `n19`'s outgoing edge to `output`. Confirmed by the locally-reverted-then-rerun test (panics at the `expect()`).

## Test quality (Important: 0, Minor: 1)

- The test exists in `mod tests` clustered with related pathway tests. Reasonable placement.
- Asserts both pathway count (1) and link count (`INTERMEDIATE_NODES + 1` = 21), so a partial-truncation regression that emitted a shortened path would still fail. Strong assertion.
- Runs in well under one second. No flakiness risk (deterministic graph construction; HashMap iteration order doesn't affect the result because the chain is single-source single-sink).
- **Minor**: the inline comment "A 21-element chain is enough to prove the old cap is gone without blowing past the per-test time budget" is slightly imprecise -- the chain has 21 edges between 22 nodes; "21-link chain" or "22-node chain" would read more cleanly. The later "22 nodes -> 21 directed links along the chain" comment is correct.

## Simplicity / maintainability (Important: 0)

The change is ~15 production lines plus comments and a ~55-line test. The node-count computation is local to `enumerate_module_pathways` (the only caller of `find_all_simple_paths` that needs the bound), so duplicating with the slightly-different formulation in `detect_output_ports` is the right call -- extracting a helper for two callers with different surrounding code (one needs `has_incoming` for input-port detection, the other needs `all_nodes` and `has_outgoing`) would not improve clarity.

## Scope discipline (Critical: 0)

PR diff touches three files: `src/simlin-engine/src/ltm.rs` (the fix), `src/simlin-serve/src/git.rs` and `src/simlin-serve/tests/git_integration.rs` (the cherry-pick). No tangential changes inside `ltm.rs`. The cherry-pick is in a different crate and is explicitly disclosed in the PR body. **Coordination flag for the user**: this branch and `ltm/480-array-reducer-polarity` both contain the same `serve: clear inherited GIT_*` commit. Whichever branch merges to `main` first will leave the other carrying a duplicate commit that must be rebased away. Not the implementer's job to resolve, but worth flagging.

## Out-of-scope follow-up (Minor: 1)

The legacy `20` cap may have been doing double duty as both a recursion-depth limit and an implicit path-count limit -- the simple-path count between two nodes can be exponential in a dense graph. With the cap raised to N, pathologically dense user modules could in principle produce many paths. The Johnson's-circuit code in this same file uses an explicit `TruncatedByBudget` mechanism for exactly this reason; `find_all_simple_paths` does not. This is **not** a regression of #312 (the old behavior was wrong about correctness, not safety), but if real-world models surface combinatorial path-count blowup, an explicit budget on `find_all_simple_paths` would be the principled follow-up. Worth tracking, but does not block this PR.

## Hygiene (Critical: 0)

- No emoji, no `Co-Authored-By`, no `Generated with` lines in either commit message or the PR body. Both commits use the project's `component: lowercase description` format.
- PR body explains the bug, the bound, the rationale (`Why N is the correct bound`), and discloses the cherry-pick clearly.

## Praise

- The implementer reverse-validated the test by reasoning from the algorithm rather than hand-waving, and the failure mode is documented inside the `expect()` message so a future regression points the reader straight at the cap.
- Good explanatory comment on the bound (lines 1550-1557) -- explains the *invariant*, not the mechanics.
- Comment chose to reuse `has_incoming` rather than rebuild a fresh `HashSet`. Tiny, but it's the kind of detail that adds up.
- Disclosing the cherry-pick in the PR body and explaining *why* (pre-commit hook recursion via `simlin-serve` git probe tests) instead of bundling it silently. Exactly the right call.

## Verdict

Fix is correct, well-tested, minimal, and within scope. No Critical or Important issues. Two Minor observations (test comment wording, follow-up on path-count budget) noted above; neither blocks merge.

bpowers:
@codex review

bpowers (commented):
## Codex P2 addressed

Switched the GIT_* env-stripping loops to std::env::vars_os() with a
byte-prefix comparison via OsStr::as_encoded_bytes (Rust >= 1.74).
Fixed in be5d9575f5d0ac54936b515146a1b12eb7a2f2f0. Behavior is identical in UTF-8 environments;
the change is panic-robustness only -- a non-UTF-8 env var anywhere
in the parent process can no longer crash production git status
probing.

Note: PRs #491 and #494 carry copies of the previous vars()-based
version of this commit. They will pick up the same fix when one of
the LTM PRs lands and the others rebase. PR #496 already uses an
explicit allowlist via env_remove(name) and is not affected.

PR #490, opened 2026-05-07
engine: surface polarity confidence for LTM loop classification

bpowers:
## Summary

`LoopPolarity::from_runtime_scores` previously collapsed any sign change in a
loop's score series to `Undetermined`, dropping useful information about loops
that briefly express the opposite polarity. This change retains the
Eberlein/Schoenberg polarity-confidence ratio `|r - |b|| / (r + |b|)` alongside
the categorical classification and adds two new variants -- `MostlyReinforcing`
("Rux") and `MostlyBalancing` ("Bux") -- for loops where one polarity dominates
above a documented threshold.

`DetectedLoop` gains a `polarity_confidence: f64` field in `[0.0, 1.0]`. The
structural pipeline currently emits 1.0 for fully-determined R/B loops and 0.0
for U loops; runtime-aware consumers can layer the `from_runtime_scores`
classification on top to surface mostly-R/B loops with their actual confidence
ratio. Existing R/B/U variants are unchanged, so persisted loop names and
deterministic IDs (r1, b1, ...) stay stable.

## Threshold

`POLARITY_CONFIDENCE_THRESHOLD = 0.99`. The value follows
Eberlein & Schoenberg (2020) "Seamlessly Integrating LTM into the Modeling
Process", which uses the same ratio and the same 0.99 cutoff to distinguish
predominantly-R/B loops (Rux/Bux) from genuinely mixed (Ux) loops in
simplified CLDs. References:

- `docs/reference/ltm--loops-that-matter.md` section 13.7 (formula)
- `docs/reference/papers/schoenberg2020.2-thesis--summary.md` section 7.6
  (Rx / Bx / Rux / Bux / Ux labelling)

## Files updated to handle the new variants

This is an additive enum change (existing R/B/U variants are unchanged), but
existing matches on `LoopPolarity` / `DetectedLoopPolarity` are no longer
exhaustive. The following sites now handle the new variants explicitly:

- `src/simlin-engine/src/ltm.rs` -- `assign_loop_ids` (R/B counters share
  with mostly-R/B), `from_runtime_scores` returns `Option<(LoopPolarity, f64)>`
- `src/simlin-engine/src/ltm_finding.rs` -- updated `assign_loop_ids` and the
  call site that previously consumed the old `Option<LoopPolarity>` return
- `src/simlin-engine/src/db_analysis.rs` -- structural pipeline emits
  confidence 1.0 / 0.0 alongside polarity; `DetectedLoop` and
[...]

PR #491, opened 2026-05-07
engine: handle array reducer polarity in ltm analyzer

### LINKED ISSUE #480: ltm: polarity analysis returns Unknown for all array reducers
## Summary

`analyze_expr_polarity_with_context` in `src/simlin-engine/src/ltm.rs` only handles the scalar two-arg forms `Max(a, Some(b))` and `Min(a, Some(b))`. Every array reducer (`Sum`, `Mean`, `Max(_, None)`, `Min(_, None)`, `Stddev`, `Rank`) falls through the catch-all `Expr2::App(_, _, _) => LinkPolarity::Unknown` arm. Any variable computed via `SUM(x[*])` or `MEAN(x[*])` therefore contributes `Unknown` polarity, and every loop crossing such a node is classified `Undetermined`.

## Why this matters

For `Sum` and `Mean`, polarity is trivially the argument's polarity (monotone in every element), so an entire class of arrayed feedback loops is misclassified as undetermined when their polarity is in fact recoverable.

## Locations

- `src/simlin-engine/src/ltm.rs:2042-2075` -- the polarity analyzer with the catch-all `Unknown`
- `src/simlin-engine/src/ltm.rs:2341` -- strict GF monotonicity epsilon (`EPSILON = 1e-10`) flags numeric import noise as `Unknown`

## Suggested fix

- Add explicit cases in the polarity match: pass through the argument's polarity for `Sum` and `Mean`; return `Unknown` for `Stddev` and `Rank` (which have no monotone polarity).
- For array `Max(_, None)` / `Min(_, None)`, return the argument's polarity (max/min of a monotone family is monotone).
- Consider a plateau-tolerant GF monotonicity check that ignores noise below a configurable epsilon rather than the strict 1e-10.

## Severity

medium

## References

- `docs/tech-debt.md` item #21
- LTM tracking issue: see meta-issue with the \`epic\` label

bpowers:
## Summary
- Add explicit polarity cases for SUM, MEAN, array MAX, array MIN (pass-through of arg polarity).
- STDDEV and RANK explicitly return Unknown (non-monotone).
- Previously all array reducers fell through Expr2::App(_, _, _) and contributed Unknown.

## Why
Variables computed via SUM(x[*]) or MEAN(x[*]) now contribute meaningful polarity, so loops crossing them are no longer misclassified Undetermined when the polarity is in fact recoverable.

## Tests
- New unit tests in src/simlin-engine/src/ltm.rs exercising each reducer case.
- Existing simulate_ltm tests pass unchanged.

## Note on the second commit
The first commit on this branch (`serve: clear inherited GIT_* env in git probe and tests`) is a prerequisite that fixes pre-existing test isolation bugs in `simlin-serve`'s git tests. Those tests `git init` a tempdir and run `git commit`, but inherit the outer `GIT_DIR` / `GIT_INDEX_FILE` env vars set by the project's pre-commit hook, causing them to operate on the outer repo and recursively trigger the outer hook. Without this fix the pre-commit hook fails on every commit attempt that touches anything else. The fix is a port of the same change living unmerged on `ltm/308-canonical-cycle-dedup`, extended here to also cover `tests/watcher_git::must_git` (which was missed in the original).

Fixes #480


bpowers (commented):
## Adversarial review

Polarity fix is on the right track for the headline cases in #480 (SUM, single-arg array MAX/MIN, STDDEV, RANK), but the variadic MEAN combiner has a real correctness asymmetry that I was able to reproduce. Other than that, this is clean: scope discipline good, prep commit is strictly better than #489's version, follow-up #492 is well-scoped, all verifications pass locally.

### Verification (run locally)
- `cargo test -p simlin-engine --lib` -> 3156 passed, 0 failed
- `cargo test -p simlin-engine --features file_io --test simulate_ltm` -> 48 passed, 0 failed, 1 ignored
- `cargo test -p simlin-serve --tests` -> all green
- `cargo clippy -p simlin-engine --all-targets --no-deps` -> clean
- `cargo fmt --all -- --check` -> clean
- New test `test_analyze_expr_polarity_array_reducers` verified failing on unmodified production code (checked by reverting only the new match arms): `assertion left == right failed: SUM(x) should pass through positive polarity / left: Unknown / right: Positive`.

### Important

**1. MEAN multi-arg combiner is asymmetric with Add (`ltm.rs:2084-2103`).**

The comment promises ``"we combine arg polarities the same way Add does (any disagreement collapses to Unknown)"``, but the implementation does not. When the FIRST iterated arg returns `Unknown` AND references `from_var` (e.g. `ABS(x)`, `SIN(x)`), the `(Unknown(combined), Unknown(arg_pol))` match takes the first arm and leaves `combined = Unknown` without ever checking `expr_references_var` on that arg. A later arg with known polarity then sets `combined` via the same first arm. Net effect: the prior arg's "Unknown but depends on from_var" is silently dropped.

Reproducer (I added this test temporarily and it fails):
```rust
let abs_x = Expr2::App(BuiltinFn::Abs(Box::new(x_var_expr)), None, Loc::default());
let mean = Expr2::App(BuiltinFn::Mean(vec![abs_x, x_var_expr.clone()]), None, Loc::default());
analyze_expr_polarity_with_context(&mean, &x_var, LinkPolarity::Positive, None);
// returns Positive; the equivalent `Add(ABS(x), x)` returns Unknown.
```

Suggested fix (lifts the references_var check before the match so it covers the first arg too):
```rust
for arg in args {
    let arg_pol = analyze_expr_polarity_with_context(arg, from_var, current_polarity, variables);
    if arg_pol == LinkPolarity::Unknown && expr_references_var(arg, from_var) {
        return LinkPolarity::Unknown;
    }
    match (combined, arg_pol) {
        (LinkPolarity::Unknown, pol) => combined = pol,
        (_, LinkPolarity::Unknown) => {} // prev branch already proved no from_var ref
        (a, b) if a == b => {}
        _ => return LinkPolarity::Unknown,
    }
}
```

For the single-arg `MEAN(x[*])` shape that #480 actually points at, this code path is correct (one iteration, returns the arg polarity). But the variadic scalar `MEAN(a, b, c)` form is also Vec-shaped at the AST level and routes through this same loop, so the bug is reachable in practice.

**2. The new test does not cover the multi-arg MEAN path.**

Both `mean_pos` and `mean_neg` construct `BuiltinFn::Mean(vec![single_inner])` -- the loop runs exactly once, so the asymmetry above is not exercised. Adding at least one `MEAN(positive_arg, negative_arg)` -> Unknown case (and ideally the `MEAN(ABS(x), x)` case) would have caught this. The TDD claim ``"each new case was confirmed failing on unmodified code before the fix landed"`` is true but only for cases the test set actually covers.

### Minor

**3. Commit message uses ``"Fixes #480"``.**

Project guidance in `CLAUDE.md` says explicitly: ``DO NOT use "fixes"/"resolves" or emoji in commit messages``. PR body has the same trailer. Easy fix on the next force-push: drop the trailer (PR body already has the gh-supported ``"Fixes #480"`` trailer for auto-close, so the commit message version is redundant anyway).

### Praise

- Builtin variant arities match `builtins.rs` exactly: `Sum(Box<Expr>)`, `Mean(Vec<Expr>)`, `Max/Min(Box<Expr>, Option<Box<Expr>>)`, `Stddev(Box<Expr>)`, `Rank(Box<Expr>, Box<Expr>)`. No tuple-arity bugs.
- Single-arg `Max(_, None)` / `Min(_, None)` correctly distinguished from the existing scalar two-arg `Max(_, Some(_))` / `Min(_, Some(_))` arm at line 2043. The two arms cannot collide.
- Each non-MEAN reducer test covers BOTH positive and negative argument polarity, so a buggy "always Positive" fallthrough would have failed.
- STDDEV and RANK explicitly return Unknown; both arms exercised by tests.
- Test is colocated with other `analyze_expr_polarity_*` tests, runs in <1ms.
- Issue #492 is real, well-scoped to the GF strict-monotonicity epsilon (NOT the reducer cases, which are correctly fixed here), references #480 and #491, carries the `ltm` label.
- Scope discipline is excellent: only the polarity fix and the prep commit; no drive-by churn.
- No emoji, no `Co-Authored-By`, no "Generated with" trailers.

### Cross-branch coordination on the GIT_* prep commit

I diffed against the version of this commit on `ltm/312-module-pathway-depth` (PR #489, sha `323744d7`):

- The shared files (`src/simlin-serve/src/git.rs`, `src/simlin-serve/tests/git_integration.rs`) are byte-identical: ``git diff 323744d7 16c59b07 -- src/simlin-serve/src/git.rs src/simlin-serve/tests/git_integration.rs`` returns empty.
- This branch's version (16c59b07) additionally extends the env-stripping to `tests/watcher_git::must_git`. That extension is genuinely needed -- `tests/watcher_git.rs` lines 129-137 do `git init`, `git config user.email/user.name`, `git add`, `git commit` on a `TempDir`, which is exactly the path that breaks under the project pre-commit hook.
- The commit message on this branch correctly mentions `tests/watcher_git::must_git` and pluralizes ``"workspace test failures"``. So the message and the diff agree, no yellow flag there.

**Concrete recommendation:** Land **this PR's prep commit first** (or extract it to a standalone PR), then rebase PR #489 to drop the duplicate. PR #489's version is a strict subset; landing it first would either force this branch to rebase (clean -- the shared diff is already identical) or to carry the watcher_git extension as a follow-up commit. The former is fine; the latter loses the bisect-friendliness of having one prep commit own the whole fix.

For reference: PRs **#490 and #493 do NOT carry this prep commit** (contrary to the review brief's expectation). Their diffs only touch engine sources, so they don't cross the `simlin-serve/git.rs` -> pre-commit-hook landmine. Once #491's prep commit lands on `main`, #489 rebases cleanly and #490/#493 are unaffected.

### Verdict

Comment-only, **changes-required** (Important #1 should be fixed before merge -- the bug is small but real, the analyzer is supposed to be conservative and this branch slightly weakens that, and the fix is one extra line). Once the MEAN combiner is corrected and the test set is extended to cover the multi-arg path, this is ready.

bpowers (commented):
## Review feedback addressed

Important (variadic MEAN): fixed in 46b3f732. Added the suggested test cases (MEAN(ABS(x), x) -> Unknown, MEAN(x, x) -> Positive, MEAN(x, -x) -> Unknown). Verified the reproducer test fails on the prior commit and passes here.

Minor ("Fixes #480" in commit message): noted. Per CLAUDE.md "create a new commit rather than amending" rule, the prior commit stands; the PR body's "Fixes #480" is the canonical auto-close.

bpowers:
@codex review

bpowers (commented):

P1 (Subscript not handled): confirmed the bug -- new test exercising
Sum(Subscript(x, [Wildcard])) returned Unknown on the prior commit.
Fixed in 70fa3696 by adding an Expr2::Subscript arm mirroring
the Var handler. Added unit tests covering the real parsed shape
(SUM(x[*]) Positive/Negative pass-through, no-match Unknown,
MEAN/MAX/MIN(x[*]) pass-through, STDDEV/RANK(x[*]) Unknown). Also
extended test_cross_dim_sum_algebraic with a link-polarity assertion
(population -> total_pop must be Positive) and a detected-loop check
(no loop should remain Undetermined); both fail on the prior commit
and pass here.

The two earlier P2 comments about variadic MEAN order-dependence
were already addressed in 46b3f732.

bpowers (commented):
## Rebased onto PR #490 (stacked PR)

This PR is now stacked on top of #490 (`ltm/485-polarity-confidence`). When #490 merges, GitHub auto-updates this PR's base.

- The polarity analyzer arms for SUM, MEAN, array MAX, array MIN, STDDEV, RANK now live in `src/simlin-engine/src/ltm/polarity.rs` (post-#498 module split).
- The variadic MEAN fix (hoisted `expr_references_var` check) is preserved.
- The `Expr2::Subscript` handler that finally delivered the headline `SUM(x[*])` case is preserved with its dedicated tests.
- The catch-all `_ => Unknown` arm removal is preserved (all `Expr2` variants now matched).
- Unit tests split between `ltm/polarity.rs`'s tests sub-mod and `ltm/tests.rs` per the new layout.
- `tests/simulate_ltm.rs::test_cross_dim_sum_algebraic` polarity assertions preserved.
- Older GIT_* prep commits dropped; supersession chain is `#496 narrowed allowlist > #489 vars_os > older vars()`.

mergeable: MERGEABLE on top of base ltm/485-polarity-confidence.

PR #494, opened 2026-05-07
engine: dedup ltm cycles by canonical edge sequence (#308)

### LINKED ISSUE #308: ltm: dedup by node set merges distinct directed loops
## Summary
Loop deduplication is currently done by **sorted node set**, which merges distinct directed loops that traverse the same variables in different orders/directions.

## Why this is a problem
In feedback analysis, loop identity is defined by directed causal structure (ordered links), not just the set of participating variables. Collapsing by node set can remove structurally distinct loops and alter loop counts, reported polarity composition, and score attribution.

The LTM literature discusses distinct loops in directed graphs and evaluates discovery completeness against exhaustive directed-loop enumeration. Treating two different directed cycles as one because they share nodes is not equivalent to elementary circuit identity.

- Exhaustive path dedup (`ltm.rs`) key = sorted `from` node names.
- Discovery dedup (`ltm_finding.rs`) key = sorted node names from DFS stack.
- Integration test notes this explicitly in arms-race expectations (7 instead of 8 due reverse three-way loop collapse).

Locations:
- `src/simlin-engine/src/ltm.rs` (`deduplicate_loops`)
- `src/simlin-engine/src/ltm_finding.rs` (`add_loop_if_unique`)
- `src/simlin-engine/tests/simulate_ltm.rs` (`discovery_arms_race_3party` comments)

## Impact
- Distinct loops can be lost, especially in symmetric systems where opposite-direction cycles exist over same node set.
- Discovery-vs-exhaustive comparisons may look better/worse for the wrong reason (identity mismatch).

## Success criteria
1. Define loop identity using directed edges (canonicalized by rotation, but preserving orientation/order), not node-set membership.
2. Apply the same identity rule in both exhaustive and discovery paths.
3. Update tests to validate that opposite-direction cycles on the same node set are represented as distinct loops when structurally distinct.
4. Revisit affected expected loop counts (e.g., arms race fixtures) under the corrected identity.
5. Ensure deterministic IDs still remain stable across runs with the new identity rule.

## Notes
This issue does not request any change to the known flow-to-stock time alignment decision.

bpowers:
## Summary

LTM currently dedups feedback loops by **sorted node-set**, collapsing distinct directed cycles like `A -> B -> C -> A` and `A -> C -> B -> A` into one `Loop`. This contradicts the elementary-circuit identity used in the LTM literature: cycle identity is determined by the directed edge sequence, not just the set of participating variables. Two cycles with the same node set can have different polarity products and different scores.

This PR replaces the sorted-node-set dedup key with a **canonical edge-sequence rotation**: rotate each circuit so it starts at the lex-smallest position, then compare the ordered node sequence (which on an elementary cycle determines the edge sequence). Two distinct directed cycles over the same node set now canonicalize to different keys and are retained as separate loops.

Fixes #308.

## Identity rule

A circuit `[v0, v1, ..., v_{n-1}]` represents the directed elementary cycle `v0 -> v1 -> ... -> v_{n-1} -> v0`. Two circuits represent the same cycle iff one is a cyclic rotation of the other. The canonical rotation is the one whose full ordered node sequence is lex-smallest among all `n` rotations. Element-wise comparison proceeds left-to-right; the first differing position decides. The `(node, next_node)` lex order is the natural tiebreaker for any hypothetical input with equal starting nodes (elementary cycles never have repeated nodes).

The new helper `canonical_rotation<T: Ord + Clone>` is shared by all three call sites; the IndexedGraph path keys on `Vec<u32>` and the discovery path on `Vec<String>`.

Design note: [`docs/design-plans/2026-05-06-ltm-308-canonical-cycle-dedup.md`](https://github.com/bpowers/simlin/blob/ltm/308-canonical-cycle-dedup/docs/design-plans/2026-05-06-ltm-308-canonical-cycle-dedup.md).

## Application points

| Site | File | Key type |
|---|---|---|
| `IndexedGraph::johnson_circuit` | `src/simlin-engine/src/ltm.rs` | `Vec<u32>` (rapidhash fingerprint) |
| `SearchGraph::add_loop_if_unique` | `src/simlin-engine/src/ltm_finding.rs` | `Vec<String>` |
| `discover_loops_with_graph` (cross-timestep dedup) | `src/simlin-engine/src/ltm_finding.rs` | `Vec<String>` |

## Fixture changes

| Fixture | Before | After | Notes |
|---|---|---|---|
| `discovery_arms_race_3party` (`tests/simulate_ltm.rs`) | 7 loops | 8 loops | 3 self-balancing + 3 pairwise reinforcing + 2 three-way (forward + reverse) |
| `johnson_complete_k3_all_directed_cycles` (`src/ltm.rs`) | 4 circuits | 5 circuits | K3 multidigraph: 3 two-cycles + 2 distinct three-cycles |
| `johnson_multidigraph_dedups_to_single_node_set` (`src/ltm.rs`) | 1 three-cycle | 2 three-cycles | Renamed `johnson_multidigraph_keeps_distinct_directed_cycles` |
| `johnson_budget_charges_duplicate_raw_cycles` (`src/ltm.rs`) | 11 circuits | 20 circuits | Renamed `johnson_budget_bounds_unique_directed_cycles`. K4 has 6 + 8 + 6 distinct directed cycles |

The `wrld3_element_level_enumeration_is_uncapped` test (`#[ignore]`d) had its assertion loosened from `==1_863_803` to `>=1_863_803` because the exact post-fix count will be remeasured next time it is run on-demand.

## Determinism / id stability

- The canonical rotation is uniquely defined by the input slice, so rotation order is fully deterministic across runs.
- Loop ids assigned by `assign_loop_ids` (sorted by underlying variable set) are stable: existing loops keep their ids; previously-collapsed loops simply gain new ids.
- Salsa cache keys (`LoopCircuitsResult` from `model_loop_circuits`) remain deterministic. The rapidhash fingerprint seed is unchanged; only the bytes hashed differ (canonical rotation vs sorted set), and the cache stores actual circuits, not fingerprints.

The `dedup_deterministic_across_calls` unit test continues to pin within-process determinism on K3 with all 6 directed edges.

## Risks

- This is a correctness behavior change. Loop counts grow on any model whose dependency graph contains symmetric / multidigraph SCCs. Empirically rare in real SD models -- only the explicitly-symmetric arms-race fixture among existing tests was affected.
- Loop ids previously assigned to a node-set may now point to one of two siblings (forward direction); the second direction gets a new id. Downstream callers that assume "one loop per node set" must be updated separately if any exist outside this PR's scope -- a search of the repo did not surface any.

## Drive-by fix in the same branch (commit 1)

`src/simlin-serve/src/git.rs::run_git` and the in-test `must_git` helpers now strip every `GIT_*` env variable before spawning. Without this, `cargo test --workspace` invoked from inside the project pre-commit hook inherits `GIT_DIR`/`GIT_WORK_TREE` and the inner `git commit` operates on the outer repository instead of the freshly-`git init`'d temp dir, producing pathspec failures and recursive pre-commit hook invocation. Discovered while running this branch's pre-commit hook.

## Test plan

- [x] `cargo test -p simlin-engine` passes (3,161 lib tests + 48 simulate_ltm + every other integration suite).
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean.
- [x] All three application sites have failing-test-first regressions (`canonical_rotation_distinguishes_opposite_directions`, `deduplicate_keeps_both_directed_three_cycles_in_multidigraph`, `add_loop_if_unique_keeps_distinct_directed_three_cycles`) plus the existing `discovery_arms_race_3party` integration test now exercises the change end-to-end.
- [ ] Workspace tests via pre-commit hook (run on review).


bpowers (commented):
# Adversarial review: canonical edge-sequence cycle dedup (#494)

Verification: ran `cargo test -p simlin-engine --lib`, `cargo test -p simlin-engine --features file_io --test simulate_ltm`, `cargo test --workspace --features file_io`, and `cargo clippy --all-targets -- -D warnings`. All test suites pass; clippy is clean.

## Critical

### `cargo fmt --check` fails in 4 places -- pre-commit hook will block this PR

```
src/simlin-engine/src/ltm.rs:948
state
    .hash_scratch
    .extend(canonical_rotation(&state.path));
-> single line.

src/simlin-engine/src/ltm.rs:5439
let mut seconds: Vec<&str> = three_loops
    .iter()
    .map(|l| l.links[0].to.as_str())
    .collect();
-> single line.

src/simlin-engine/src/ltm_finding.rs:715
let path_strings: Vec<String> =
    path.iter().map(|id| id.as_str().to_string()).collect();
-> single line.

src/simlin-engine/src/ltm_finding.rs:1169
let forward: Vec<Ident<Canonical>> = vec![Ident::new("a"), Ident::new("b"), Ident::new("c")];
let reverse: Vec<Ident<Canonical>> = vec![Ident::new("a"), Ident::new("c"), Ident::new("b")];
-> wrap to two lines each.
```

The PR body's test-plan note ("Workspace tests via pre-commit hook (run on review)") is unchecked, which is presumably how this slipped through. Run `cargo fmt -p simlin-engine` and amend.

## Important

### `wrld3_element_level_enumeration_is_uncapped` -- `>=1_863_803` is too permissive

The original assertion `==1_863_803` was a tight regression lock pinning the post-dedup count under the previous (sorted-node-set) rule. Switching to `>=` means a future regression that *drops* the count to e.g. 1,200,000 would silently pass. Since the test is `#[ignore]`d and runs only on demand, the only protection it offered was the exact equality.

The design doc (lines 172-178) explicitly committed to "adjust its expected value when the change lands, with the new count documented in the test comment for traceability." The PR retreated from that plan. Required action: run `cargo test --release -p simlin-engine --test wrld3_ltm_panic -- --ignored wrld3_element_level_enumeration_is_uncapped` once, capture the new count, and pin it back to `==`. ~30s in release; one-time cost.

If you genuinely cannot remeasure now, file a tracked issue ("re-pin wrld3 element-level enumeration count") so the loose bound doesn't survive indefinitely. A permanent `>=` on a regression-locking test is a tech-debt landmine.

### Drive-by GIT_* env scrub is missing the `tests/watcher_git.rs` extension

Branch `ltm/480-array-reducer-polarity` extends the same fix to `src/simlin-serve/tests/watcher_git.rs::must_git`. The brief noted "this branch has the original (pre-watcher_git extension) version" -- confirmed: branch #494's commit `34c4befa` only patches `src/git.rs`, `src/git.rs::tests::must_git`, and `tests/git_integration.rs`. `watcher_git.rs::must_git` (lines 41 and 116-171 use the helper) is untouched.

If this is a genuine production GitProbe bug, every git-spawning helper is equally affected. Either:
1. Extend this PR's commit to cover `watcher_git.rs::must_git`, OR
2. Land a follow-up commit before merging that aligns with the 480 version.

Don't merge with an inconsistent fix across siblings.

### Loop-id stability is more subtle than the PR claims (still likely OK in practice)

The PR/design doc state: "Loop ids assigned by `assign_loop_ids` (sorted by underlying variable set) are stable: existing loops keep their ids; previously-collapsed loops simply gain new ids."

`assign_loop_ids` (`ltm.rs:1967-1977`) sorts by the dedup'd-and-joined variable set, which is identical for forward and reverse cycles over the same node set. Stability between forward and reverse depends entirely on `sort_by_key`'s stable-sort fallback preserving the *enumeration order* fed to it. That ordering, in turn, depends on Johnson's DFS visit sequence, which is unchanged by this PR -- so the cycle that master kept as "the survivor" is still the first one out of Johnson's, gets the lower ID, and is preserved. The newly-added direction gets the next available counter.

This is fine in practice but the doc oversells it. Suggested clarification: "Existing IDs remain stable as long as Johnson's DFS visit order is unchanged; this PR does not modify Johnson's traversal logic, so within-process ID assignments are identical to master for previously-existing loops. The newly-uncollapsed direction gets the next sequential ID per polarity bucket."

This matters if any downstream consumer (model snapshots, persisted UI state) holds onto loop IDs across versions: since the new loop IDs cluster (`r3` appears alongside `r2` for the same node set), any UI that treats IDs as opaque keys is fine; any UI that infers structure from the ID sequence could be surprised.

## Minor

### `canonicalize_circuits` test helper is now redundant with the inline dedup

`src/ltm.rs:5141-5148` -- `canonicalize_circuits` first canonical-rotates then sorts/dedups. Johnson's emissions are already canonical-rotation-deduped during enumeration, so the only useful work this helper does on the Johnson side is sorting. It's still needed on the Tiernan oracle side. Keep as-is, but a one-line comment clarifying "the canonical_rotation step is a no-op for Johnson's output but reduces Tiernan's redundant rotations" would help readers.

### `johnson_complete_k3_all_directed_cycles` still uses the sorted-node-set assertion as a sanity check

`src/ltm.rs:5301-5311` keeps the original `vec!["a", "b", "c"]` repeated entry as the "expected sorted node sets" form, with a comment explaining the duplicate is real. This is a deliberate cross-check, fine. But the test name still reads "all_directed_cycles" -- the assertion only checks counts and node sets, not directionality. The actual directionality check is in `johnson_multidigraph_keeps_distinct_directed_cycles`. Consider renaming this test to `johnson_complete_k3_circuit_count_and_node_sets` to set expectations correctly, or merge the two tests.

### `arms_race_3party` cross-validation uses node-set equality

`tests/simulate_ltm.rs:435-444` -- the assertion "discovered loop should exist in exhaustive results" still uses sorted node-set comparison, so it cannot distinguish forward from reverse three-way loops. The exact-count `found.len() == 8` assertion saves correctness, but the cross-check is weaker than it appears. Pre-existing weakness, not introduced by this PR; flagging for awareness.

### Minor: O(n^2) where O(n) is available

`canonical_rotation` is the textbook least-rotation problem; Booth's 1980 algorithm gives O(n). For n ≤ 50 (the SCC gate), the current implementation is fine -- 2500 comparisons per emitted cycle. The justification comment is correct. A future tightening would only help WRLD3-scale enumeration where the helper runs millions of times, and even then the hot-loop cost is hash, not rotation. Not a blocker.

## Praise

### Algorithm is correct

I traced the algorithm by hand on `[1,2,3]`, `[2,3,1]`, `[3,1,2]`, the reverse `[1,3,2]`, and the documented edge case `[1,5,1,3]`. The lex-smallest cyclic rotation is computed correctly in all cases. The `(node, next_node)` tiebreaker described in the design doc is implemented as full-rotation-sequence lex compare, which is the right realization (the doc's "(node_index, next_node_index) pair-sequence" reads as if it were a position-by-position score; in practice the two produce identical outputs for any valid input). Forward and reverse three-cycles canonicalize to different keys; rotations of the same directed cycle canonicalize to the same key; empty input returns empty; one- and two-element cycles are trivially correct (no other rotations to consider).

### Generic over `T: Ord + Clone`

Single helper used at three call sites (numeric `Vec<u32>` in IndexedGraph and string `Vec<String>` in two ltm_finding sites). No string-specific assumptions. Good.

### K3=5 / K4=20 math is right

Verified by hand:
- K3: C(3,2)=3 two-cycles + 2 directed three-cycles = 5.
- K4: C(4,2)=6 two-cycles + C(4,3)·2 = 8 three-cycles + (4-1)!/2·2 = 6 directed Hamilton 4-cycles = 20.

### Strong test coverage of the helper

`canonical_rotation_picks_lex_smallest_start`, `_distinguishes_opposite_directions`, `_handles_repeated_starting_element`, `_empty_input_returns_empty` -- comprehensive. The repeated-starting-element test is particularly nice because elementary cycles never trigger that path; it's defensive coverage.

### Direct regression tests at all three application sites

`deduplicate_keeps_both_directed_three_cycles_in_multidigraph` (Johnson's), `add_loop_if_unique_keeps_distinct_directed_three_cycles` (discovery), and `discovery_arms_race_3party` (cross-timestep). Each fails-loud on a regression in the corresponding site.

### Determinism preserved

`dedup_deterministic_across_calls` continues to pin within-process byte-identical output on the multidigraph. Rapidhash seed unchanged; canonical_rotation is fully deterministic (no float ops, no hashmap iteration). Salsa cache stability holds.

### Comments updated alongside behavior

`has_no_duplicate_node_sets` -> `has_no_duplicate_canonical_rotations`, `loops_have_unique_node_sets` -> `loops_have_unique_canonical_rotations`, plus updated rustdoc explaining why the rotation matters and what it doesn't catch. The dedup-in-place comment block in `johnson_circuit` correctly explains both the multidigraph case and that the budget charges raw emissions before dedup as defense-in-depth.

### Design doc is excellent

Identity rule, application points, stability, fixture impact, polarity composition with a worked example -- all answered. The polarity composition example is hypothetical (the actual arms-race fixture has symmetric polarities), which is appropriately flagged. K4 math is shown.

### Multi-commit scope discipline

Three logical commits (design doc, algorithm + tests, fixture update) plus the unrelated GIT_* drive-by. No drift into unrelated tests; no comment cleanup beyond the dedup change.

## Overall verdict

Block on the formatting issue (Critical). After fmt, fix the wrld3 `>=` (Important) and align watcher_git.rs (Important). Then ship.

The core change -- replacing sorted-node-set with canonical-rotation dedup at three call sites -- is correct, well-tested, and matches the elementary-circuit identity used in the LTM literature. This is the right fix for #308.

bpowers (commented):
## Review feedback addressed

Critical (fmt): fixed in 58a47fbf. `cargo fmt --check` now clean.

Important (wrld3 assertion): re-measured and pinned in ad483c11.
New count: 1,863,803 (identical to the 2026-04-18 sorted-node-set
measurement, because WRLD3-03's element-level graph happens to
contain no pairs of opposite-direction cycles over the same node
set). Comment notes the post-canonical-dedup source.

Important (watcher_git extension): aligned with PR #491 in baeeef48.
This branch's version of the GIT_* fix now matches the canonical
version carried on `ltm/480-array-reducer-polarity` (commit 16c59b07
on that branch). Whichever PR lands first carries the canonical
version and the other can drop the duplicate during rebase.

End-to-end pre-commit was green on each of the three new commits:
fmt, cbindgen, clippy, `cargo test --workspace`, TypeScript checks,
and pysimlin tests all pass.

PR remains in draft.

bpowers:
@codex review

bpowers (commented):

P3 (stale hash_u32_slice rustdoc): fixed in f01755c4. The non-hash overhead now references `canonical_rotation` (Booth's O(n)) instead of the long-removed `sort_unstable` step.

Per-emission Vec allocation: addressed in db5da917. The Johnson hot path now calls `lex_smallest_rotation_start` with a long-lived `rotation_scratch` on `JohnsonState` and writes the canonical rotation halves directly into the existing reusable `hash_scratch` via `extend_from_slice`. Both per-emission allocations the bot review and owner review flagged (the throwaway `Vec<u32>` path copy and Booth's failure-function `Vec<i32>`) are gone.

Algorithmic improvement: addressed in 359ce6a0. `canonical_rotation` now uses Booth's algorithm (1980) for an O(n) lex-smallest-rotation lookup instead of the prior O(n^2) cyclic comparison loop. The new `lex_smallest_rotation_start` helper exposes a scratch parameter so the hot path can amortize Booth's failure-function buffer; `canonical_rotation` itself is a thin wrapper that allocates a temporary scratch. Booth output is verified against the previous O(n^2) reference on every existing fixture (empty, singleton, periodic, all-equal, repeated-start) plus a deterministic 256-input sweep over short random inputs with small alphabets.

bpowers (commented):
## Empirical correction: O(n) algorithms slower than the simple loop here

I posted a previous reply claiming Booth's was a win; that turned out to be wrong on measurement and I've reset the branch to a single clean commit (879811da). Apologies for the noise.

### What changed

The new commit lands the per-emission Vec allocation fix the bot review and owner review both flagged: `canonical_rotation` is split into a start-index helper (`lex_smallest_rotation_start`) plus a thin allocating wrapper, and `johnson_circuit` calls the helper directly with `extend_from_slice` into the reusable `hash_scratch`, eliminating the throwaway `Vec<u32>` per raw cycle.

The stale `hash_u32_slice` rustdoc snippet (P3) is also fixed in the same commit; it now references `canonical_rotation` rather than the long-removed `sort_unstable` step.

### What didn't change (and why)

I implemented Booth's (1980) and Duval/Lyndon factorization to address the algorithmic improvement question. Both are provably O(n) and beat the simple cyclic-comparison loop's worst-case O(n^2). On this specific workload they are slower, not faster, end-to-end.

WRLD3 element-level enumeration wall time on three runs each, release with the workspace's default `opt-level = "z"`:

| Variant                                    | Wall time |
|--------------------------------------------|-----------|
| Baseline (before any of this PR's edits)    | 1.70s     |
| Booth's + alloc fix                         | 1.94s (+14%) |
| Duval + alloc fix                           | 1.86s (+9%)  |
| Simple O(n^2) + alloc fix (this PR)         | 1.57s (-8%)  |

At `opt-level = 3` the relative ranking is identical: 1.25s baseline -> 1.53s Booth -> 1.40s Duval -> 1.18s simple-with-alloc-fix.

The reason is structural to LTM cycle data: cycles use distinct `u32` node indices (no repeated elements), so the simple loop's inner comparison almost always exits at the first differing element after one or two reads. The algorithm runs in roughly O(n) on real inputs while paying near-zero per-element overhead. Booth/Duval pay constant-factor overhead per position (failure-function bookkeeping or Lyndon-word extension state) that dominates for `n` capped at MAX_LTM_SCC_NODES = 50.

The rustdoc on `lex_smallest_rotation_start` records this empirical finding with a pointer back to the PR so a future maintainer doesn't relitigate it without measurement.

### Verification

- `cargo test -p simlin-engine --features file_io` -- all integration tests green; no test value changes.
- `cargo clippy --workspace --all-targets -- -D warnings` -- clean.
- `cargo fmt --check` -- clean.
- WRLD3 ignored benchmark: count still 1,863,803, wall time 1.70s -> 1.57s in release at default opt-level.

PR remains marked ready for review.

bpowers (commented):
## Rebased onto PR #491 (top of stacked chain)

This PR is now stacked on top of #491 (`ltm/480-array-reducer-polarity`), which itself stacks #490 -> #496 -> main. As each PR below merges, GitHub auto-updates this PR's base.

- Canonical-rotation helper + start-index variant now in `src/simlin-engine/src/ltm/mod.rs` (post-#498 module split).
- `IndexedGraph::johnson_circuit` allocation fix preserved: writes the canonical rotation directly into the reusable hash_scratch via `lex_smallest_rotation_start` + extend_from_slice (no per-emission Vec<u32>).
- Empirical decision to keep the simple O(n^2) inner over Booth/Duval is preserved; rustdoc on `lex_smallest_rotation_start` documents the wall-time benchmark.
- All three call sites use the helper: `IndexedGraph::johnson_circuit` (now in `ltm/indexed.rs`), `SearchGraph::add_loop_if_unique` (`ltm_finding.rs`), `discover_loops_with_graph` cross-timestep dedup (`ltm_finding.rs`).
- K3=5, K4=20, multidigraph distinct cycles, arms_race_3party=8 fixture updates preserved in `ltm/tests.rs` and `tests/simulate_ltm.rs`.
- WRLD3 count pinned at the exact 1,863,803 with the empirical comment explaining why post-canonical-dedup matches pre-canonical-dedup on this fixture.
- Older GIT_* prep commits dropped; the chain `#496 narrowed allowlist > #489 vars_os > older vars()` is established on the parent branches.
- Design doc preserved.

mergeable: MERGEABLE on top of base ltm/480-array-reducer-polarity.

Issue #25, opened 2020-12-13
array support


bpowers:
roughly in priority order
- [x] lexer + parser support for subscripts
- [ ] simulation support
- [ ] lexer + parser support for transpose (`'`) unary postfix operator
- [ ] use dimension names in A2A equations
- [ ] use `Dimension.SubscriptName` directly in equations to get the integer subscript value (1, 2, etc)
- [ ] Apply-to-all arrays with non-apply-to-all graphical functions
- [ ] Slicing -- `A[1, *]` and `SUM(A[1:3])`
- [ ] "Dimension positions"

bpowers:
I had written a bunch of notes on sd.js a few years ago that are relevant (more or less)

- diagram changes (minimal)
- figure out if it makes sense to do single-dimensional first, or multi-
  dimensional from the start
- parser:
  - array reference/slicing
  - transpose operator
- semantic analysis/validation:
  - validate indexing
  - validate slicing
  - transpose using dimension names
  - transpose using positions
  - array slicing (`A[1, *]`)
- optimization?
  - the simple thing to do is have a nest of for loops for each individual
    variable.  This is simple, I worry it will be slow for large models (which
    are very common users of arrays).  If we're doing operations on multiple
    variables with the same dimensions in a row, we can merge them into a
    single loop.  This is straightforward logically, but there is no
    optimization framework in place yet, so that would need to be added.
- codegen:
  - apply-to-all equations
    - nested for loop
  - non-A2A equations
  - non-A2A graphical functions
  - array slicing (`A[1, *]`)
  - transpose using dimension names
  - transpose using positions
- runtime:
  - know about defined dimensions + their subscripts
  - allocate correct amount of space for arrayed variables
  - array builtins: `MIN`, `MEAN`, `MAX`, `RANK`, `SIZE`, `STDDEV`, `SUM`.
  - be able to enumerate all subscripted values for CSV output
  - be able to return results for `arrayed_variable[int_or_named_dimension]`
  - right now, the runtime is pretty dead-simple.  Every builtin
    function expects one or more numbers as input.  With the array
    builtins, this is no longer the case.
    - index into arrays with non-constant offsets:
      `constants[INT(RANDOM(1, SIZE(foods)))]`
    - create slices of arrays: `SUM(array[chosen_dim, *])`, where
      `chosen_dim` is an auxiliary variable.
    - this means runtime type checking, and I think runtime memory
      allocation (right now memory is allocated once, in one chunk, when a
[...]

PR #496, opened 2026-05-07
ltm: enumerate variable-level loops first, expand only cross-element subgraph

### LINKED ISSUE #482: ltm: enumerate variable-level loops first, expand only cross-element subgraph
## Summary

`model_element_loop_circuits` in `src/simlin-engine/src/db_analysis.rs` enumerates circuits directly on the element graph. For a pure-A2A model with V variables over an N-element dimension, every variable-level cycle inflates to N element-level cycles that `build_element_level_loops` then collapses back to one A2A loop. Cost is multiplicative when it should be additive in N.

The 2026-04-25 per-reference graph refactor removed the spurious N x N edges that made FixedIndex models worst-case, but the structural multiplication for pure-A2A loops remains. Dense-feedback A2A subgraphs still trip the `MAX_LTM_SCC_NODES = 50` auto-flip gate sooner than variable-level enumeration would.

## Locations

- `src/simlin-engine/src/db_analysis.rs:1497-1508` -- `model_element_loop_circuits` calls `find_indexed_circuits_with_limit(usize::MAX)` on the element graph directly
- `src/simlin-engine/src/db_ltm.rs::build_element_level_loops` -- the collapsing step

## Suggested fix

1. Enumerate cycles at the variable level first.
2. Tag each cycle's edges by `RefShape`: same-element (A2A pass-through), cross-element, or scalar.
3. For cycles that are pure same-element, emit one Loop per cycle with `dimensions` set; do not enter the element-level enumerator.
4. Only run element-level enumeration on the cross-element subgraph (or cycles whose tag set is mixed/cross-element).

Cost becomes additive in N for pure-A2A loops instead of multiplicative; SCC width stays below the auto-flip threshold for many real models.

## Severity

medium

## References

- `docs/tech-debt.md` item #25
- Related: closed issue #448 (per-reference graph refactor that reduced edge density)

bpowers:
## Summary

Implements the variable-level loop enumeration refactor planned in
`docs/design-plans/2026-05-06-ltm-482-variable-level-loop-enumeration.md`.
LTM exhaustive-mode loop detection now runs Johnson on the variable graph
first, classifies each cycle by its `RefShape` composition, and emits
pure-A2A and pure-scalar cycles directly as Loops without entering
element-level Johnson. Only cross-element / mixed cycles flow through
element-level Johnson, and only on the subgraph induced by their nodes.

Cost asymmetry that made pure-A2A cycles inflate multiplicatively in N
is gone. The auto-flip gate's domain narrows to the variable graph and
the cross-element subgraph.

Fixes #482.

## Architecture

| Step | Today | After this PR |
|------|-------|---------------|
| Cycle enumeration | Johnson on the full element graph (`model_element_loop_circuits`) | Johnson on the variable graph + classify by `RefShape` + Johnson on the slow-path subgraph (`model_loop_circuits_tiered`) |
| Pure-A2A cycle | N element-level circuits, collapsed back to 1 Loop in `build_element_level_loops` | 1 fast-path circuit, materialized directly into 1 Loop |
| Pure-scalar cycle | 1 element-level circuit -> 1 Loop | 1 fast-path circuit -> 1 Loop |
| Cross-element cycle | element-level Johnson on full element graph | element-level Johnson on the *slow-path subgraph* (the element graph restricted to cross-element variables) |
| Auto-flip gate | full element-graph SCC > 50 | variable-level SCC > 50 (early gate) OR slow-path subgraph SCC > 50 (late gate) |

The new `model_edge_shapes` salsa-tracked function records the
multiset of `RefShape` values seen at every reference site for each
variable-level edge. The cycle classifier
(`classify_cycle`) reads this map plus per-variable dimension
information to decide PureScalar / PureSameElementA2A /
CrossElementOrMixed.

## Design plan

[`docs/design-plans/2026-05-06-ltm-482-variable-level-loop-enumeration.md`](docs/design-plans/2026-05-06-ltm-482-variable-level-loop-enumeration.md)
covers the staging breakdown, cost model, fixture impact predictions,
auto-flip threshold reasoning, and risks. The measurement postscript
in that document is filled in with the actuals captured by the new
integration tests.

## Measurements

Captured by `tests/simulate_ltm.rs::measurement_postscript_*` on the
post-Stage-7 branch tip. Numbers come from
`model_loop_circuits_tiered` (fast / slow path counts and slow-path
SCC) plus the legacy `model_element_loop_circuits` (for the legacy
column).

| Fixture | var SCC | elem SCC | var circuits | elem circuits (legacy) | fast path | slow path | slow-path SCC | auto-flip |
|---------|--------:|---------:|-------------:|------------------------:|----------:|----------:|--------------:|-----------|
| cross_element_ltm | 5 | 10 | 3 | 8 | 1 | 6 | 8 | no / no |
| arrayed_population_ltm | 3 | 3 | 2 | 6 | 2 | 0 | 0 | no / no |

Notes:

- **arrayed_population_ltm** (pure-A2A, 3 regions): legacy enumerated
  6 element-level circuits (2 cycles * 3 regions); tiered enumerator
  produces 0 element-level circuits and 2 fast-path cycles. Slow-path
  subgraph is empty.
- **cross_element_ltm** (mixed): the population<->births A2A cycle
  classifies as PureSameElementA2A (1 fast-path entry); the two
  FixedIndex-driven migration cycles drive the slow-path subgraph.
  Slow-path Johnson enumerates 6 element-level circuits versus 8 in
  the legacy full-element Johnson run because the slow-path subgraph
  excludes births[*].

WRLD3-03 (scalar, 166-node SCC) auto-flips on the new variable-level
SCC gate exactly as it did on the legacy element-graph SCC gate
(WRLD3 is fully scalar, so var SCC == elem SCC). hero_culture_ltm
(scalar) classifies every cycle as PureScalar; fast_path equals
var_circuits and slow_path is empty.

## Loop output stability

All 48 tests in `src/simlin-engine/tests/simulate_ltm.rs` pass
without modification. Loop IDs are content-stable: `assign_loop_ids`
sorts loops by sorted distinct variable names before numbering, so
the merged Loop list (fast + slow) gets consistent r1/b1/u1
assignments across the refactor.

## New fixture

`tests::pure_a2a_eliminates_per_element_circuit_redundancy` in
`src/simlin-engine/src/db_analysis.rs` builds a 30-element pure-A2A
fixture and asserts that the legacy enumerator produces 30
element-level circuits while the tiered enumerator produces 1
fast-path cycle and 0 slow-path Johnson runs. This is the
structural-win demonstration described in the design plan.

## Drive-by fixes

Two simlin-serve test helpers (`must_git` in `git.rs`,
`tests/git_integration.rs`, and `tests/watcher_git.rs`) inherited
`GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` from the parent
process. When the workspace test suite runs from inside an outer
`git commit` (the project pre-commit hook invokes
`cargo test --workspace`), those env vars made the test's inner
`git commit` operate on the OUTER repository instead of the
freshly-`git init`'d temp dir, causing recursive pre-commit-hook
invocation. Fixed by stripping `GIT_*` env vars before spawning.
Discovered while running this PR's pre-commit on a developer
machine that has `core.hooksPath` configured globally.

## Test plan

- [x] `cargo test -p simlin-engine` -- all 3176 tests pass
- [x] `cargo test -p simlin-engine --features file_io --test simulate_ltm` -- all 50 tests pass (48 prior + 2 new measurement-postscript tests)
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` -- clean
- [x] `cargo build -p simlin --no-default-features` -- clean (verifies `TieredCircuitsResult` works without the `debug-derive` feature)
- [x] Pre-commit hook passes on every commit in this branch


bpowers (commented):
# Code Review: PR #496 -- LTM Variable-Level Loop Enumeration

## Status
**CHANGES REQUIRED**

## Issue Summary
**Critical: 1 | Important: 2 | Minor: 4 | Praise: several**

## Verification Evidence

```
Tests: cargo test -p simlin-engine --features file_io
       -> 3,476+ tests passed; 14 ignored; 0 failed (across all binaries).
       simulate_ltm: 50 passed + 1 ignored. db tests / lib tests: 3,212 passed.
       Wall clock: ~10s total. Well under the 3-minute cap.

Build: cargo build -p simlin --no-default-features
       -> clean (verifies tracked types compile without debug-derive feature).

Clippy: cargo clippy --workspace --all-targets --all-features
        -> clean (only unrelated xmutil C++ warnings).
```

All existing `tests/simulate_ltm.rs` tests pass without modification (no expected-value
updates), confirming the byte-identical loop output claim _on the existing test
fixtures_. See Critical-1 below for a topology not exercised by any existing test.

## Plan Alignment

### Implemented Requirements
- AC1 (Edge-shape extraction): met.
- AC2 (Cycle classification): met.
- AC3 (Tiered enumeration): met.
- AC4 (Loop output stability on existing fixtures): met for the fixtures that
  exist; **not met in general** -- see Critical-1.
- AC5 (Structural-win demonstration): partially met. The plan called for a
  test-only override of `MAX_LTM_SCC_NODES` to demonstrate before/after
  auto-flip behavior. The implementation instead pins circuit-count
  inequality on a 30-element pure-A2A fixture, which is also fine and the
  PR description acknowledges the gate-behavior win is a no-op on existing
  fixtures (since #448 already broke up element SCCs into N independent
  size-2 SCCs).
- AC6 (Documentation): met -- design plan, engine CLAUDE.md, LTM design
  doc, and tech-debt all updated.

### Deviations from Plan
- Stage 5 plan said "leave the gate alone, document the change, and file a
  follow-up issue to retarget the gate after measurement." The PR instead
  retargets the gate to `max(variable_scc, slow_path_largest_scc)`.
  **Justified**: the design plan also reasons through the retarget
  rationale on lines 786-805, and the new gate is strictly safer
  (pure-A2A models with huge N stop tripping the gate; WRLD3 still trips
  it from the variable-level SCC). Moving Stage 5 forward is the right
  call.

## Critical Issues (count: 1)

### Critical-1: Double-counting of pure-A2A loops when their variables overlap with a slow-path cycle's variable set

- **Issue**: A pure-A2A cycle whose variables are a SUBSET of the variables
  in some `CrossElementOrMixed` cycle gets emitted TWICE as a Loop.
- **Location**: `src/simlin-engine/src/db_analysis.rs:1928-2013` (slow-path
  subgraph induction in `model_loop_circuits_tiered`).
- **Impact**: Violates the stated "byte-identical loop output" guarantee.
  Produces duplicate `loop_score` synthetic variables in `model_ltm_variables`,
  which means the same A2A reinforcing/balancing loop ends up with two
  distinct loop IDs (e.g., `r1` AND `r2` for the same content), inflating
  the user-visible loop list and breaking downstream rel-loop-score
  normalization (post-sim divides by sum of |scores| within a partition;
  the duplicate inflates the denominator).
- **Reproduction** (verified by code probe; minimal fixture):

  ```
  state[Region] stock with inflow a[r]
  a[r]   = b + c        (flow)
  b[r]   = state + 0.01 (Bare reference to state)
  c[r]   = b[NYC] + 0.05 (FixedIndex reference to b)
  ```

  Variable cycles:
  - `[a, state, b]`: every edge Bare/structural -> classified PureSameElementA2A -> fast.
  - `[a, state, b, c]`: `b -> c` is FixedIndex -> CrossElementOrMixed -> slow.

  `slow_path_var_nodes = {a, state, b, c}` (every var in the slow cycle).
  The induced element-level subgraph keeps `a[*]`, `state[*]`, `b[*]`,
  `c[*]` element nodes AND every element edge among them. Johnson on this
  subgraph produces:
  - `a[NYC] -> state[NYC] -> b[NYC] -> a[NYC]` (the A2A cycle reflected at NYC)
  - `a[Boston] -> state[Boston] -> b[Boston] -> a[Boston]` (the A2A cycle reflected at Boston)
  - `a[NYC] -> state[NYC] -> b[NYC] -> c[NYC] -> a[NYC]` (the legitimate cross-element cycle)

  Then `build_element_level_loops` on the slow path groups the two pure-A2A
  element circuits by their stripped key `a\0state\0b` and emits ONE A2A
  Loop -- a duplicate of the fast-path emission.

  Observed (PR branch):
  ```
  fast_path: 1
    fast: ["a", "state", "b"] dims=["region"]
  slow_path: 3 circuits
    slow: ["a[boston]", "state[boston]", "b[boston]"]   <-- duplicate of fast path
    slow: ["a[nyc]",    "state[nyc]",    "b[nyc]"]      <-- duplicate of fast path
    slow: ["a[nyc]",    "state[nyc]",    "b[nyc]", "c[nyc]"]
  --- LTM loop_score vars: 3 ---
    $⁚ltm⁚loop_score⁚r1 dims=["Region"]   <-- A2A
    $⁚ltm⁚loop_score⁚r2 dims=["Region"]   <-- DUPLICATE A2A (same cycle as r1)
    $⁚ltm⁚loop_score⁚u1 dims=["Region"]   <-- legitimate cross-element loop
  ```

  Same fixture on `main`:
  ```
  --- main: LTM loop_score vars: 2 ---
    $⁚ltm⁚loop_score⁚r1 dims=["Region"]   <-- A2A
    $⁚ltm⁚loop_score⁚u1 dims=["Region"]   <-- legitimate cross-element loop
  ```

  i.e. main emits the structurally correct 2 loops; this PR emits 3.

  Why existing fixtures don't catch this: in `cross_element_ltm` the A2A
  cycle is `[population, births]` and `births` does not appear in any
  slow-path-classified cycle (migration cycles use population +
  migration_pressure + migration_in/out). So `slow_path_var_nodes`
  excludes `births`, the induced subgraph drops `births[*]` element
  nodes, and the A2A cycle's element-level reflection cannot close in
  the subgraph. The bug only manifests when every variable of a
  fast-path cycle ALSO participates in a slow-path cycle. None of the
  existing test fixtures (cross_element_ltm, arrayed_population_ltm,
  hero_culture_ltm, WRLD3, the in-tree synthetic 30-element pure-A2A
  fixture) construct that topology.

- **Fix**: Two viable approaches:

  1. After Johnson on the slow-path subgraph, drop any element-level
     circuit whose stripped variable-level node sequence (modulo cyclic
     rotation) matches a `FastPathCircuit.variables` sequence. The
     fast-path emission is the canonical source for those cycles; the
     slow-path enumeration must yield *additional* cycles only.
  2. Before running Johnson on the slow-path subgraph, prune element-level
     edges that participate ONLY in same-element pure-A2A cycles -- i.e.
     restrict the subgraph to edges that have a non-Bare reference shape
     OR cross-element same-shape variants. (Trickier to get right at
     element-edge granularity; approach 1 is simpler.)

  Either fix needs a regression test that pins the fixture above (or any
  other model that constructs the overlapping-cycle topology). The fixture
  in this review is small enough to live in `simulate_ltm.rs` as a unit
  test.

## Important Issues (count: 2)

### Important-1: `MAX_LTM_SCC_NODES` docstring not updated for the new gate semantics

- **Issue**: `src/simlin-engine/src/ltm.rs:1-46` documents the constant as
  applying to "the largest SCC of the element graph." Stage 5 of the
  design plan explicitly called for updating this docstring to reflect
  the new two-tier gate (variable-level SCC OR slow-path subgraph SCC).
  The new gate code in `db_ltm.rs` documents the change in-line, but the
  constant's own doc comment is now misleading.
- **Location**: `src/simlin-engine/src/ltm.rs:5-46`.
- **Fix**: Update the docstring to enumerate the two SCCs the gate now
  applies to and link to the 2026-05-06 design plan alongside the
  existing 2026-04-18 link.

### Important-2: Legacy `model_element_loop_circuits` retained but documented as legacy

- **Issue**: The PR description and engine CLAUDE.md correctly state
  that `model_element_loop_circuits` is now legacy (only the
  measurement test and the new `pure_a2a_eliminates_per_element_circuit_redundancy`
  unit test consume it). Keeping the legacy function alive in production
  code is a deliberate choice (Stage 8 in the plan defers its removal),
  but there is no `#[deprecated]` attribute and no compile-time
  enforcement that production callers don't accidentally re-introduce
  a dependency on it. As tech-debt #25 notes "(...) the auto-flip gate
  retargeting is a follow-up", but the legacy enumerator and the new
  enumerator now diverge on the topology described in Critical-1, so
  any future caller that picks the legacy path will silently get the
  pre-#482 (correct) behavior while the LTM path gets the buggy
  behavior. Either `#[deprecated]` it, or add an internal `#[doc(hidden)]`
  / module-level rustdoc warning, or convert it into a thin wrapper
  around the tiered enumerator's slow-path output (deferred to Stage 8
  in the plan, fine for now -- but track the divergence in tech-debt).

## Minor Issues (count: 4)

### Minor-1: `measurement_postscript_cross_element_ltm` doc comment claims `var_scc = 4` but actual is 5

- **Location**: `src/simlin-engine/tests/simulate_ltm.rs:3895-3905`.
- **Issue**: The doc comment says `var_scc = 4 (population, migration_pressure,
  migration_in, migration_out are in one variable-level SCC; total_population
  is acyclic)`, but the actual measured value (printed by `--nocapture`)
  is 5, matching the design plan's table on line 753 and the PR
  description. The comment lists 4 variables (missing births) but the
  actual SCC is 5 (population, births, migration_pressure, migration_in,
  migration_out). The test doesn't assert the exact value (only loose
  inequalities) so this isn't a test failure -- just a stale comment.
- **Fix**: Either update the comment to `var_scc = 5` and list all 5
  participating variables, or tighten the assertion to pin the exact
  numbers from the design-plan table.

### Minor-2: Tech-debt #25 marked RESOLVED before merge

- **Location**: `docs/tech-debt.md:225-233`.
- **Issue**: Same convention concern as PR #490: marking RESOLVED is
  premature if the PR isn't merged. If the PR is force-pushed or
  abandoned, the tech-debt status would silently lie. Other entries
  (#20, #23, #26) follow the same convention so this isn't unique to
  this PR, but the project has not yet adopted a "(pending #N)" prefix
  to make the conditional visible.
- **Fix**: Either revert the resolved marker until merge, or rephrase
  to "RESOLVED (pending #496)" to make the conditional explicit.

### Minor-3: `FastPathCircuit.dimensions` fallback could produce mismatched casing

- **Location**: `src/simlin-engine/src/db_ltm.rs:2002-2015`.
- **Issue**: `build_loops_from_tiered` maps each canonical dim name back
  to its raw datamodel name via `dm_dims.iter().find(...)`. The fallback
  `unwrap_or_else(|| canonical.clone())` returns the canonical (lower-case)
  name, which would produce a `Loop.dimensions` entry that doesn't match
  any datamodel dim and would then fail to resolve when the equation
  parser feeds it into `Equation::ApplyToAll`. This shouldn't happen for
  well-formed projects because every cycle node's dim must be in
  `dm_dims`, but there's no diagnostic if it ever does.
- **Fix**: Either add a debug-assert that the canonical name was found
  (panics in debug, defaults silently in release) or document the
  invariant in a code comment so the reader knows the fallback is a
  defensive no-op.

### Minor-4: Slow-path subgraph induction is not commented to explain the no-double-counting reasoning

- **Location**: `src/simlin-engine/src/db_analysis.rs:1928-1965`.
- **Issue**: The slow-path induction excludes element nodes for variables
  not in `slow_path_var_nodes`. The reasoning that this excluded set
  prevents double-counting is critical to correctness, but the comment
  block doesn't explain it. With Critical-1 above showing the reasoning
  is in fact incomplete (the bug arises because variables CAN be in
  both sets when fast-path cycles' variable sets are subsets of slow
  cycles' variable sets), this is exactly the kind of comment that
  should have been there to make the bug visible during review.
- **Fix**: Whichever fix is chosen for Critical-1, document the
  invariant the slow-path subgraph is supposed to preserve, and the
  test that pins the topology. (This minor is largely subsumed by the
  Critical-1 fix.)

## Praise

- **Comprehensive design plan**: 805 lines, walks through architecture,
  cost model, fixture predictions vs measurements (postscript filled in
  with actuals), risks (R1-R5), 8 stages, ACs. Honest about what
  post-#448 the gate-behavior win is, and what is purely circuit-count
  reduction.
- **Honest measurement table** in the design plan and PR description:
  the post-#448 baseline is acknowledged, and the actual win is
  characterized correctly as additive-not-multiplicative cost (not as
  a gate-behavior fix).
- **GIT_* drive-by commits are byte-identical** to the equivalent commits
  on PRs #491 (`16c59b07`) and #494 (`34c4befa`, `baeeef48`) modulo a
  minor docstring rewording on `watcher_git.rs`. Consistent across PRs.
- **No emoji, no Co-Authored-By, no "Generated with"** in commits, PR
  body, or design plan. Lowercase commit-message prefixes follow the
  project convention.
- **`#[salsa::Update]` derives correctly applied** to the new tracked
  types (`EdgeShapesResult`, `FastPathCircuit`, `TieredCircuitsResult`).
  `Hash`/`Ord` on `RefShape` make `BTreeSet<RefShape>` deterministic.
  `cargo build --no-default-features` clean confirms `Debug` on the new
  types is unconditional and correct.
- **Test budget within targets**: each new test runs sub-second on debug
  build. The `pure_a2a_eliminates_per_element_circuit_redundancy` 30-element
  fixture finishes effectively instantly.
- **Engine CLAUDE.md update is accurate and concise**: describes the
  tiered enumerator entry points, `classify_cycle` contract, and the
  new auto-flip gate semantics in one paragraph each.
- **`build_loops_from_tiered`'s `assign_loop_ids` re-run** on the merged
  fast/slow Loop list correctly preserves the legacy ID assignment
  scheme (sort by stripped distinct var-name set). The merging strategy
  is right; the bug is in WHAT goes into the merged list.

## Skills Applied

- `coding-effectively` (defense-in-depth, writing-good-tests):
  applied to validate that the existing tests don't actually pin the
  byte-identical loop output claim across all topologies. The test
  fixtures don't construct the variable-set-overlap topology.
- Code reading: traced cycle classification, slow-path subgraph
  induction, and `build_loops_from_tiered` end-to-end on a worked
  example before and after the change.
- Verification before completion: ran the test suite, clippy, no-default-features
  build, and a custom probe to verify the double-counting hypothesis
  on both branches.

## Decision

**BLOCKED -- CHANGES REQUIRED**

Critical-1 is a real correctness regression: this PR emits duplicate Loops
on a topology that any moderately complex arrayed model can produce. The
test suite happens to not exercise it, so CI is green, but the bug is
demonstrable in two evaluations of an 8-line `TestProject` fixture.

Recommended path forward:
1. Add the worked-example fixture (above) as a new `simulate_ltm.rs` test
   that asserts the loop count matches main (or asserts no two Loops
   share a stripped variable-name multiset).
2. Watch it fail on this PR.
3. Fix via the slow-path-circuit dedup approach (Critical-1 fix
   suggestion 1) or the slow-path-subgraph edge-pruning approach
   (suggestion 2). Suggestion 1 is simpler.
4. Re-run the full LTM test suite; confirm all numbers in the
   measurement-postscript tests still match the design-plan table.
5. Address Important-1 (MAX_LTM_SCC_NODES docstring) and the four
   Minor issues in the same fix-up pass.

bpowers (commented):
## Review feedback addressed

Critical (slow-path A2A duplicate Loop): fixed in b74364ff. Approach:
**Option A** -- dedupe slow-path circuits against fast-path emissions
inside `model_loop_circuits_tiered`. After the slow-path Johnson run,
each circuit's stripped variable-level node sequence is compared
(rotation-invariant via `canonical_cycle_rotation`) against the set of
fast-path circuit node sequences and dropped if it matches. This is the
robust fix that stays correct even if the induction set is broader than
strictly necessary; Option B (refining the slow-path induction set) was
considered but the risk of subtle topology misses was higher than the
clarity benefit. Added the reviewer's reproducer (`a/grow/stock/b` pure-
A2A reinforcing + `a/grow/stock/b/c` cross-element with `c[r] = b[NYC] *
0.5`) as `test_dedup_slow_path_a2a_against_fast_path` in
`tests/simulate_ltm.rs`. Verified the fixture fails on the prior commit
(3 loops emitted: `r1`, `r2`, `u1`) and passes after the fix (2 loops).

Important (`MAX_LTM_SCC_NODES` docstring): updated in 41b778cc to
describe the post-#482 two-granularity gate (variable-level SCC plus
slow-path subgraph SCC).

Important (legacy fn `#[deprecated]`): added in 41b778cc.
`model_element_loop_circuits` now carries a `#[deprecated]` attribute.
Existing internal callers (measurement helpers, `db_element_graph_tests`,
`db_ltm_unified_tests`, `ltm_full_bench` example) carry
`#[allow(deprecated)]` with a comment explaining why each callsite
intentionally targets the legacy surface. The salsa-tracked function's
own definition needs an inner `#[allow(deprecated)]` because the
`#[salsa::tracked]` macro expansion references the function by name in
its generated code; the comment in the source explains this constraint.

Minor (`var_scc=4 -> 5` in doc comment): fixed in 98c6193e. Confirmed
post-#482 measurement is `var_scc=5 / elem_scc=10` (births is in the
SCC because its structural inflow plus the `population` reference closes
the population<->births A2A pair).

Minor (tech-debt #25 RESOLVED reverted): fixed in 98c6193e to
`RESOLVED (pending #496 merge)`.

Minor (`FastPathCircuit` casing fallback): fixed in 98c6193e. The
`unwrap_or_else(|| canonical.clone())` in `build_loops_from_tiered`
that masked an unmatched datamodel dimension is now an `unreachable!`
that lists the known dimensions, with a comment documenting the
invariant the panic enforces.

Minor (slow-path induction comment): added in b74364ff alongside the
fix. The slow-path subgraph block now explains why the induction is
intentionally broad and why dedup is required.

PR remains in draft.

bpowers:
@codex review

bpowers (commented):

GIT_* env strip is now narrowed to path-targeting vars only
(GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, GIT_OBJECT_DIRECTORY,
GIT_COMMON_DIR). Fixed in 08c1874e.

Future code that uses run_git to commit, push, or pull will now
inherit GIT_AUTHOR_*, GIT_COMMITTER_*, GIT_SSH_COMMAND, GIT_ASKPASS,
GIT_TERMINAL_PROMPT, etc. from the calling environment as expected.
The original pre-commit-hook reentrancy fix is preserved -- the
narrow set still includes every var that retargets git at the wrong
repository.

(PRs #491 and #494 carry their own copies of the broader strip; the
narrowing will land there as part of the cross-branch rebase once
one of these PRs merges to main.)

bpowers (commented):

GIT_* env strip is now narrowed to path-targeting vars only
(GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, GIT_OBJECT_DIRECTORY,
GIT_COMMON_DIR). Fixed in 08c1874e.

Future code that uses run_git to commit, push, or pull will now
inherit GIT_AUTHOR_*, GIT_COMMITTER_*, GIT_SSH_COMMAND, GIT_ASKPASS,
GIT_TERMINAL_PROMPT, etc. from the calling environment as expected.
The original pre-commit-hook reentrancy fix is preserved -- the
narrow set still includes every var that retargets git at the wrong
repository.

(PRs #491 and #494 carry their own copies of the broader strip; the
narrowing will land there as part of the cross-branch rebase once
one of these PRs merges to main.)

bpowers (commented):
## Rebased onto post-#498 main

- Dropped the redundant older GIT_* commit (main's #489 carried the
  vars_os version; this branch's narrowed allowlist supersedes it).
- Re-applied the narrowed-allowlist commit (strip_git_path_env helper
  with explicit const list of path-targeting vars only) -- preserves
  GIT_AUTHOR_*, GIT_COMMITTER_*, GIT_SSH_COMMAND, GIT_ASKPASS,
  GIT_TERMINAL_PROMPT for future commit/push code.
- Carried the LTM tiered-enumeration changes onto the new ltm/
  module structure. The MAX_LTM_SCC_NODES docstring update now
  lives in src/simlin-engine/src/ltm/mod.rs.
- Codex P2 (build_loops_from_tiered unreachable! on dim lookup
  miss): fixed in a9637ca4. Replaced unreachable! with a
  fallback to the canonical dim name + debug_assert!, mirroring
  build_element_level_loops's behavior.

mergeable: MERGEABLE after rebase + force-push.

This PR is the bottom of the LTM stack. Subsequent rebases of
#490, #491, #494 will set their base branches to the next PR's
branch (then ultimately to main) so the stack flattens as PRs
land.

PR #499, opened 2026-05-08
engine: detect self-indexing in subscript polarity analyzer

bpowers:
## Summary

The `Expr2::Subscript` arm in `analyze_expr_polarity_with_context`
(added in #491) returned `current_polarity` whenever the array's
identifier matched `from_var`, without inspecting the index
expressions. For self-indexing patterns like `arr[INT(arr[i])]`,
the relationship between `arr` and the subscripted value is
non-monotone -- changing `arr` shifts both the lookup target AND
the index. The arm now walks index expressions via
`expr_references_var` and returns `Unknown` when any index
references `from_var`.

The dominant cases (`SUM(arr[*])`, `arr[Region]`, indices that
reference other variables) are unaffected -- those indices do not
reference `from_var`, so the existing behavior of returning
`current_polarity` is preserved.

## Why now

Surfaced by a code review on #491 (the array reducer polarity PR)
after that PR merged. The reviewer flagged it as non-blocking but
worth generalizing because the arm's docstring claimed the case was
fully handled. This PR follows up to make the docstring's claim
true.

## Test plan

- [x] New unit tests covering: literal index, wildcard, var-not-from_var,
      var-from_var (NEW), nested self-indexing (NEW), other-array.
- [x] Existing polarity / cross-dim tests pass unchanged.
- [x] cargo clippy --workspace --all-targets -- -D warnings clean
- [x] cargo fmt --check clean

### DIFF
--- a/src/simlin-engine/src/ltm/polarity.rs
+++ b/src/simlin-engine/src/ltm/polarity.rs
@@ -95,15 +95,38 @@ pub(super) fn analyze_expr_polarity_with_context(
         // the production case `SUM(x[*])` that argument lowers to
         // `Subscript(x, [Wildcard], _, _)`, not `Var(x, ...)`. Mirror the Var
         // handler so the identifier comparison succeeds and the reducer's
-        // monotonicity guarantee carries through. The subscript indices don't
-        // affect the polarity contract for the reducer arms because each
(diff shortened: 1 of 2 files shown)


bpowers:
@codex review

PR #618, opened 2026-05-22
engine: robust (partial-result) unit inference and checking

### LINKED ISSUE #614: engine: unit inference is all-or-nothing -- one conflict discards all inferred units for the model (parallels units.rs:263 TODO)
## Description

Unit inference is all-or-nothing per model: a single dimensional conflict anywhere in a model discards every inferred unit for that whole model, blinding subsequent checks to the rest of the variables.

`units_infer::infer` (`src/simlin-engine/src/units_infer.rs:1147`) returns a single `UnitResult<HashMap<Ident<Canonical>, UnitMap>>`, so it aborts and returns `Err` on the **first** conflict it finds -- discarding all the metavars it had already resolved up to that point.

`db.rs::check_model_units` (`src/simlin-engine/src/db.rs:1484-1485`) then does:

```rust
let inferred_units = crate::units_infer::infer(&models_s1, units_ctx, target_model)
    .unwrap_or_else(|err| {
        // emit one diagnostic, then ...
        Default::default()   // empty inferred-units map for the ENTIRE model
    });
```

So when any one equation conflicts, `inferred_units` collapses to an empty map and every downstream per-variable unit check in that model runs with **no** inferred units. One bad equation degrades unit checking for the whole model.

This directly parallels the pre-existing TODO at `src/simlin-engine/src/units.rs:263`:

```rust
// TODO: we shouldn't discard the whole context if there are errors
if unit_errors.is_empty() {
    Ok(ctx)
} else {
    Err(unit_errors)
}
```

-- the same all-or-nothing failure shape, at the unit-context construction layer rather than the inference layer.

## Why it matters

- **Correctness of diagnostics**: a single dimensional conflict suppresses unit checking for every other variable in the model, so genuine errors elsewhere go unreported until the first one is fixed -- an iterative, one-at-a-time debugging experience instead of a complete report.
- **Cascading masking**: this is one mechanism behind the C-LEARN unit-error behavior (a model-level conflict makes the per-variable picture incomplete), and it makes residual-diagnostic triage harder because the set surfaced depends on which conflict aborts first (compounded by the nondeterministic ordering tracked in #474).
- **Maintainability**: the `unwrap_or_else(|_| Default::default())` shape is a code smell that hides the partial-result information the inferencer actually computed.

## Components affected

- `src/simlin-engine/src/units_infer.rs` (`infer` -- the single-`Result` return type)
- `src/simlin-engine/src/db.rs` (`check_model_units` -- the `unwrap_or_else(|_| Default::default())` discard)
- `src/simlin-engine/src/units.rs:263` (the parallel TODO at the context-construction layer)

## Possible approaches

1. Have `infer` return partial results alongside conflicts -- e.g. `(HashMap<Ident, UnitMap>, Vec<UnitError>)` or a struct -- so the metavars it *did* resolve survive even when some equation conflicts. `check_model_units` then keeps the resolved units and reports the conflicts, rather than throwing both away.
2. Apply the same partial-result shape to `Context` construction (the `units.rs:263` TODO) so a duplicate/conflicting unit declaration does not discard the whole context either.
3. Continue inference past the first conflict (collect all conflicts rather than short-circuiting), which also yields a more complete diagnostic set in one pass.

## Context

Identified while fixing the C-LEARN unit-error flood on branch `fix-clearn-unit-errors`. The four fixes there addressed specific over-flagging root causes; this architectural all-or-nothing discard is a separate, more general robustness issue that is out of scope for that branch. Related: #474 (order-dependence of which conflict surfaces first), #35 (general unit-checking tracker), tech-debt item 2 (`unwrap_or_default()` usage in simlin-engine).

bpowers:
## Summary

Makes the unit subsystem partial-result throughout, so one bad declaration or one dimensional conflict never blinds the rest of the model. Closes #614 and the long-standing `units.rs:263` TODO (its context-layer parallel), plus two robustness/semantic fixes surfaced along the way.

Five commits, each TDD'd with the full pre-commit gate green:

1. **`gen_constraints` is total.** It only ever returned `Ok`, so the `UnitResult<Units>` was vestigial and its `.unwrap()` in `gen_all_constraints` was a latent panic. Now returns `Units`. This also closes a propagation gap: the `None`-arm `continue` skipped the declared-units constraint for equation-less variables, so a variable with declared units but no equation (the editor's "units typed, equation not yet written" state) never informed inference of its dependents.

2. **RANK results are dimensionless.** A rank is an ordinal index, not the units of the ranked array; fixed in both inference and checking.

3. **`infer` returns partial results (#614).** `InferenceResult { resolved, conflicts }`. Solving continues past a conflict keeping the first binding -- a contradiction is confined to its connected component of the constraint graph, since substitution only flows along shared metavariables, so this can never corrupt an independent component -- and `find_constraint_mismatches` collects every residual contradiction rather than the first. `check_model_units` keeps the resolved units (so the rest of the model is still checked) and surfaces conflicts as one umbrella warning rather than one-per-conflict.

   Keeping the resolved units unmasked a pre-existing leak the old all-or-nothing behavior was accidentally suppressing: a Vensim macro can annotate its body variables' units with the formal parameter *names* (`~ xfrom` inside `RAMP FROM TO`) -- a polymorphic unit, not a concrete base unit. Inference recurses into macro instances, so those names leaked as literal units into every instantiation and conflicted with the real argument units. `ModelStage0`/`ModelStage1` gain an `is_macro` flag and inference now skips declared-units constraints for macro bodies, letting them be inferred polymorphically from the equation and cross-module parameter bindings (mirroring `check_model_units`, which already skips macro models). This restores C-LEARN to its documented 14-diagnostic residual.

4. **`Context::new` returns partial context.** It built the full context then discarded it on any duplicate/conflicting declaration; `project_units_context` then fell back to an empty context, losing project-wide alias normalization (yr/year, person/people) and re-creating a spurious mismatch flood. Now returns `(Context, Vec<errors>)`: the context always holds every valid declaration, with the conflicting ones reported alongside.

5. **Docs.** Engine `CLAUDE.md` Unit analysis section updated to describe the new contracts.

## PL framing

Units form a free abelian group, so a unit-constraint set is a linear system over the integers and an inconsistency is a single unsatisfiable row -- it says nothing about the other rows. The fix mirrors how production type systems accumulate errors and keep a partial result rather than aborting at the first failure (rustc error tainting, GHC deferred type errors, the `Validation` applicative); the natural unit of fault isolation here is the connected component of the constraint graph, which the substitution mechanism already respects.

## Testing

New tests, each written RED then GREEN: partial-results-survive-conflict (+ a second independent conflict still reported), declared-units-without-equation propagation, macro-polymorphic-units (reproduces the C-LEARN `xfrom` leak in miniature), RANK dimensionless in inference and in checking, and partial-context-on-conflict. Full `cargo test --workspace` passes; the `#[ignore]` C-LEARN flood guard (`tests/clearn_unit_errors.rs`) and the wrld3 unit tests pass unchanged, with C-LEARN holding at its documented 14-diagnostic residual (deterministically).


bpowers:
Addressed in 26dfffc9. The ordering critique is correct: `unify` substituted the newly-derived units before deciding whether to keep them, which would propagate the *rejected* units on a conflicting re-derivation. I moved substitution into the first-resolution branch so only the accepted binding is ever propagated.

Worth noting for context: the conflict arm is currently **unreachable**. `substitute` removes the metavariable from every remaining constraint (and the metavar index), so a metavariable becomes a `single_fv` at most once and `resolved_fvs` never already contains it at that check. A genuine over-constraint (the same metavariable forced to two different units) instead reduces to a residual concrete contradiction (e.g. `meter == second`) that `find_constraint_mismatches` reports. I verified this empirically with a probe in the conflict arm: zero hits across the full lib suite (3609 tests), the C-LEARN model (380 conflicts), and the wrld3 unit tests.

So there's no observable bug to reproduce with a test today, but the reorder is the correct robustness fix — it keeps "keep the first binding" sound if that invariant ever changes (e.g. a future change to the substitution strategy). The follow-up of removing the now-provably-redundant in-`unify` conflict detection entirely (relying solely on `find_constraint_mismatches`) is noted for a separate cleanup.

PR #632, opened 2026-05-23
engine: reconcile mid-run getValue across VM and wasm (#625)

### LINKED ISSUE #625: engine: mid-run getValue diverges between VM and wasm for non-stock vars (live curr-chunk overshoot handling)
## Summary

After a `runTo(t)` that stops **mid-interval** (i.e. before the simulation's end time), the bytecode VM and the wasm simulation blob leave **different content** in their live "current" data chunk for **non-stock** variables (flow / aux / constant). A mid-run `getValue(name)` reads that live curr chunk on both backends, so the two backends agree mid-run **only** for stocks and the reserved time variables; for non-stock variables the value differs mid-run. After a *full* run (`runToEnd`) all variables agree exactly, and the saved series (`getSeries`) agree for every variable at every saved step.

This is **not a wasm defect** -- it is a faithful difference in the two executors' handling of the integration overshoot chunk. The wasm curr chunk is arguably the more self-consistent of the two. The Phase 2 tests correctly scope mid-run `getValue` parity to stocks + time. This issue exists to **document the contract** (and optionally reconcile the two executors) so a future caller relying on a mid-run `getValue` of a non-stock variable is not surprised.

## Verified mechanism

**VM (bytecode interpreter):**
- The Euler loop breaks on `curr[TIME] > end` **before** evaluating the resting chunk (`src/simlin-engine/src/vm.rs:711-723`).
- Its non-save advance does `curr.copy_from_slice(next)` (`vm.rs:695-698`), copying the **entire** next row -- but that row's flow/aux slots were never recomputed for the new time. So the VM's live curr chunk holds correct integrated **stocks** + reserved **time** vars, but **stale** flow/aux/constant slots.
- A mid-run read goes through `get_value_now`, which reads the live curr chunk at base 0.

**wasm blob:**
- `emit_run_to` / `emit_save_advance` (`src/simlin-engine/src/wasmgen/module.rs:1147-1206`) copy **only** the stock offsets next->curr, leaving the freshly-evaluated flow/aux in curr.
- A mid-run read is a base-0 `DataView` read of the same live curr chunk.

**Consequence:** mid-run `getValue` agrees between VM and wasm for stocks + reserved time vars (both hold the integrated/advanced values there), but differs for flow/aux/constant variables (VM = stale previous-step values copied wholesale; wasm = freshly evaluated values for the resting point).

## Why it matters

The wasm backend's entire purpose is **interactive scrubbing** (`runTo` + read), so mid-run reads are a first-class use case, not an edge case. Today the guaranteed-VM-equal mid-run surface is implicitly only {stocks, reserved time vars, saved series}. A future consumer that does a mid-run `getValue` of a flow or aux and expects it to match the VM (or to be "the value at the stop time") would get a backend-dependent answer with no documented contract telling them so.

To be explicit about current correctness scope (all verified):
- Mid-run `getValue`: VM == wasm for stocks + reserved time vars only.
- Full run (`runToEnd`) `getValue`: VM == wasm for **all** variables.
- Saved series (`getSeries`): VM == wasm for **every** variable at **every** saved step.

## Component(s) affected

- `src/simlin-engine/src/vm.rs` (Euler loop `:711-723`, non-save advance `:695-698`, `get_value_now`)
- `src/simlin-engine/src/wasmgen/module.rs` (`emit_run_to` / `emit_save_advance`, `:1147-1206`)
- `src/engine` (the TS `getValue` path that surfaces the live curr chunk for interactive scrubbing)

## Possible approaches for resolution

This is a documentation-first item; pick one or both:

1. **Document the contract.** State plainly (in the engine `CLAUDE.md` and/or rustdoc near `get_value_now` / the wasm `emit_run_to`) that only stocks + reserved time vars + saved series are guaranteed VM-equal **mid-run**; non-stock live-curr values mid-`runTo` are executor-internal and may differ between the VM and wasm backends.
2. **Reconcile the two executors' overshoot-chunk handling** so a mid-run read of a non-stock variable is well-defined and identical across backends -- e.g. have the VM evaluate the resting chunk's flow/aux before breaking (matching wasm's freshly-evaluated curr), or have both backends agree on a single defined semantics for "value at the mid-run stop time."

## How it was discovered

Identified during the Phase 2 code review of the `@simlin/engine` wasm simulation backend integration (`docs/implementation-plans/2026-05-22-engine-wasm-sim/`, branch `engine-wasm-sim`). The reviewer confirmed the mechanism in `vm.rs` and `wasmgen/module.rs`, confirmed that full-run and saved-series parity hold exactly, and confirmed the Phase 2 tests deliberately scope mid-run `getValue` parity to stocks + time. This is filed as the tracking item for the undocumented mid-run semantic difference. Related but distinct from #616 (VM per-instance `set_value` override clobbering) and #617 (inline-fixture tests bypassing the wasm parity hook).

bpowers:
## Summary

Closes #625. After a `runTo(t)` that stops **mid-interval**, the bytecode VM and the wasm blob left **different content** in their live `curr` chunk for **non-stock** variables, so a mid-run `getValue(name)` agreed between backends only for stocks + reserved time vars. Verified still live on `main` (post-#630) and actually worse than the issue described -- the VM returns **0** for constants mid-run:

| variable | kind | VM `getValue` | wasm `getValue` |
|---|---|---|---|
| `teacup_temperature` | stock | (agree) | (agree) |
| `room_temperature` | constant | **0** | 70 |
| `characteristic_time` | constant | **0** | 10 |
| `heat_loss_to_room` | flow | **0** | 2.43… (one `dt` behind) |

Per the chosen resolution (reconcile, not document), this makes mid-run `getValue` self-consistent and identical across backends for **every** variable.

## Root cause

Both integration loops break on `curr[TIME] > target` **after** the save+advance tail. The advance steps time and integrates stocks but does not recompute flow/aux/constants for the advanced time:
- **VM:** Euler's wholesale `curr.copy_from_slice(next)` (and the chunk-ring advance) lands `curr` on a chunk whose non-stock slots are stale -- 0 for a constant the overshoot chunk never held.
- **wasm:** `emit_save_advance` copies only the stock offsets `next -> curr`, leaving flow/aux one step behind the advanced time + stocks.

So `curr`'s stocks/time are at the resting point but its non-stock slots are not -- internally inconsistent on both, and divergent between them.

## Fix

After the `run_to` loop, re-evaluate **root flows once at the resting `curr`** on both backends (`vm.rs` reuses the same `StepPart::Flows` eval the RK paths already use; `wasmgen/module.rs` emits one `flows(0)` call after the loop). The live `curr` chunk becomes fully self-consistent -- "the value at the current time" -- and byte-identical VM-vs-wasm for every variable.

Well-scoped, by construction:
- **Saved series unaffected:** every results row was already committed; the VM's `get_series` reads chunks `[0, curr_chunk)`, excluding the resting chunk.
- **Full run unaffected:** `runToEnd` reads the last *results* row, and its break path already leaves a freshly-evaluated `curr`, so the re-eval is idempotent (a resumed `run_to` on a full slab stays the #630 no-op).
- **Resume unaffected:** a resumed `run_to` re-evaluates from scratch, and the re-eval does **not** re-snapshot `prev_values`, so `PREVIOUS` still sees the last completed step.

Applies uniformly to Euler/RK2/RK4. No TS/host change -- the host just reads `curr`.

## Tests (TDD)

- `wasmgen/module.rs`: `mid_run_curr_is_self_consistent_and_matches_vm` -- a flow-phase aux `doubled = level * 2` exposes the one-step lag; asserts every var's mid-run `curr` equals the VM's `get_value_now`, and that `doubled == level * 2` at the resting time (20, not the lagged 16). Watched fail first (`inflow_rate` wasm=2 vs vm=0).
- `wasm-backend.test.ts`: AC2.2 widened from stocks+time to **every** variable. Watched fail first (constant diff of 10) against the pre-fix wasm, green after the rebuild.

Full pre-commit gate passes: `cargo test --workspace` (incl. the always-on corpus `wasm_parity_hook`), `pnpm -C src/engine test` (412), clippy, tsc, pysimlin.


bpowers:
@codex review

bpowers:
@codex review

Issue #645, opened 2026-05-31
engine: db/tests.rs is 79 lines from the 6000-line per-file lint cap


bpowers:
## Problem

`src/simlin-engine/src/db/tests.rs` is at ~5921 / 6000 lines. The repo-wide lint (`scripts/lint-project.sh`, `MAX_LINES=6000`, enforced in pre-commit and CI) will trip on the next non-trivial test addition, blocking an unrelated commit until someone does a tangential split.

This is the same class of problem as #518 (`db.rs`), #588 (`vm.rs`), and #471 (`vdf.rs`), but for a distinct file: the salsa-incremental test module.

## Why it matters

`db/tests.rs` is a high-traffic test file for the salsa incremental pipeline -- new incremental-compilation and analysis tests land here regularly. The next addition over ~79 lines will fail `pnpm lint` on an unrelated diff. Splitting it now (with headroom) is lower-risk than under duress.

## What's in db/tests.rs

The file already has clean `// --` section headers that map to natural sibling test files:

- Accumulator parity
- Incremental sync
- Incremental compilation
- AC acceptance
- PREVIOUS / INIT

## Possible approach

Split `db/tests.rs` into 2-3 sibling test files along the existing section headers, following the `#[path = "..."] mod ...;` submodule precedent already used in the `db/` module.

## Component

- `src/simlin-engine/src/db/tests.rs`
- Lint rule in `scripts/lint-project.sh`

## Severity

low / maintenance.

## Context

Identified during the salsa-pipeline-cleanup refactor.

Issue #644, opened 2026-05-31
engine: two offset machineries disagree on submodel size with a table-only variable (latent off-by-N)


bpowers:
## Problem

Two independent offset machineries size a module-typed body variable differently when the submodel contains a table-only (lookup-only) variable:

- `calc_flattened_offsets_incremental` (`src/simlin-engine/src/db/assemble.rs`) sizes a module-typed body variable by the **sum of its `sub_offsets` entry sizes**, which **excludes** table-only / lookup-only variables (they are skipped from the offsets map).
- `compute_layout` (`src/simlin-engine/src/db/layout.rs`) sizes the same submodule by the full `sub_layout.n_slots`, which **does count** table-only slots.

If a submodel contains a table-only variable (the standalone lookup-only kind from #606) and is instantiated as a module, the two machineries can **disagree** on the submodule's size, corrupting result-map slot offsets (an off-by-N in the result map).

## Why it matters

This is a latent correctness bug: result-map slot offsets could silently point at the wrong slots for any model that instantiates such a submodel as a module. It produces wrong numbers, not a crash.

## Current status

- **Not tripped by any in-corpus model today** -- no existing fixture has a submodel containing a lookup-only variable instantiated as a module.
- **Not introduced by the `is_root` refactor** -- this is pre-existing; the salsa-pipeline-cleanup refactor surfaced it.
- Related to #606 (standalone lookup-only variable handling) but distinct: #606 is about the `gf(Time)` lowering being a unit error; this is about the two offset machineries disagreeing on submodule size when such a variable is present.

## Components affected

- `src/simlin-engine/src/db/assemble.rs` (`calc_flattened_offsets_incremental`)
- `src/simlin-engine/src/db/layout.rs` (`compute_layout`)

## Possible approach

Reconcile the two size computations so both either include or exclude table-only slots consistently. Add a regression test: a submodel containing a lookup-only variable, instantiated as a module, asserting the results-map offsets match the assembled layout (i.e. the two machineries agree).

## Severity

medium / latent correctness bug.

## Context

Identified during the salsa-pipeline-cleanup refactor.

Issue #643, opened 2026-05-31
engine: compile_project_incremental deep-clones CompiledSimulation out of the Arc on every call


bpowers:
## Problem

After the assembly-memoization change, `assemble_simulation` returns `Arc<CompiledSimulation>` (a salsa-tracked memo). But `compile_project_incremental` (`src/simlin-engine/src/db.rs`) does `(*arc).clone()` to return an owned `CompiledSimulation`, preserving its public signature.

The big win lands: a no-op recompile correctly **skips** assembly (the memo is reused). But the call still deep-clones the entire `CompiledSimulation` -- the `modules` / `offsets` / `cached_constant_info` maps -- on every call, even when nothing changed.

## Why it matters

The deep clone is pure overhead on the hot recompile path. Returning `Arc<CompiledSimulation>` to callers would drop even that clone. The downstream callers (libsimlin / pysimlin / cli) construct `Vm::new` by value, so threading the `Arc` through (or having `Vm::new` accept `&CompiledSimulation` / `Arc<CompiledSimulation>`) would eliminate the clone end-to-end.

## Not a correctness issue

This is a performance follow-up only -- results are unaffected.

## Components affected

- `src/simlin-engine/src/db.rs` (`compile_project_incremental`)
- Callers constructing `Vm::new`: libsimlin, pysimlin, simlin-cli

## Severity

low / perf.

## Context

Identified during the salsa-pipeline-cleanup refactor (the assembly-memoization change that introduced `Arc<CompiledSimulation>` from `assemble_simulation`).

Issue #642, opened 2026-05-31
engine: salsa::Update on compiled bytecode uses PartialEq over f64/NaN (benign missed-backdate)


bpowers:
## Problem

`CompiledModule` / `CompiledSimulation` derive `salsa::Update`, which (via autoref specialization) falls back to the derived `PartialEq` for the bytecode side-channels. Those side-channels contain `f64` values:

- `ByteCode.literals` and `ByteCodeContext.graphical_functions` (`src/simlin-engine/src/compiler/symbolic.rs`)
- `crate::vm` `Specs` (`src/simlin-engine/src/results.rs`)

NaN literals are reachable: out-of-range subscripts compile to `Const(NaN)`. Because `NaN != NaN` under `PartialEq`, `old == new` can report "changed" on bit-identical output, defeating salsa's backdating for these outputs.

## Why this is benign today

A cache hit is decided by the inputs' revisions, not by the output's `PartialEq`. The only consumer of these outputs (`compile_project_incremental`) is non-tracked, so an output that is never backdated costs at most one extra `assemble_simulation` re-exec -- never a wrong result and never a defeated cache hit.

It is also a pre-existing property of the already-tracked `compile_var_fragment` / `PerVarBytecodes`, so this change in framing does not introduce new risk.

## Why it still matters

The derived `PartialEq` is a latent footgun: a future change that makes one of these outputs feed a *tracked* query would silently lose backdating on any model with NaN literals, with no compile error to flag it. Relying on "the only consumer happens to be non-tracked" is fragile.

## Components affected

- `src/simlin-engine/src/compiler/symbolic.rs` (`ByteCode.literals`, `ByteCodeContext.graphical_functions`)
- `src/simlin-engine/src/results.rs` (`crate::vm` `Specs`)

## Possible approach

Hand-write `Update` / `PartialEq` for the bytecode types using bit-pattern (`f64::to_bits`) float comparison, so NaN compares equal to itself and bit-identical bytecode backdates correctly.

## Severity

low / cleanup.

## Context

Identified during the salsa-pipeline-cleanup refactor.

bpowers:
Cross-reference: #987 records the same `f64`-in-a-derived-`PartialEq` root for `ModelStage0` / `ModelStage1` / `db::query::ParsedVariableResult` (alongside #981 for the LTM types).

[...]

PR #646, opened 2026-05-31
engine: simplify and harden the salsa compilation pipeline

bpowers:
## Summary

A 22-commit cleanup of the salsa-based compilation pipeline (`src/simlin-engine/src/db*`), driven by an adversarial design review. Every commit passed the full pre-commit hook (`cargo test --workspace` + clippy + fmt + TypeScript + Python) **and** a fresh adversarial code review with the fix-loop run to zero findings. The net production logic shrinks substantially; the line count is roughly flat because the simplification is offset by considerable new test coverage and the module-split scaffolding.

### Structure / navigability
- Moved the dozen `db_*.rs` files into a `db/` directory (dropping the redundant prefix).
- Split the two giant files into cohesive submodules: `db.rs` 5152 -> 615 lines (`db/{input,query,sync,diagnostic,layout,fragment_compile,assemble}.rs`); `db/ltm.rs` -> a `db/ltm/` directory (`mod`/`parse`/`compile`/`loops`/`link_scores`). Every file is now well under the per-file line cap.

### Simplification (dead/redundant code removed)
- Deleted the `Source*` mirror types and the `reconstruct_variable` round-trip; salsa inputs now hold `datamodel::*` directly (per-field invalidation granularity preserved by salsa's per-field read tracking).
- Removed the always-no-op `IN_TRACKED_CONTEXT` accumulator gating.
- Collapsed four `variable_direct_dependencies_*` query variants into one keyed on an interned `ModuleInputSet`; same for `model_dependency_graph(_with_inputs)`.
- Removed the dead interned-id (`ModelId`/`VariableId`) double-bookkeeping from the sync state.
- Unified `dt_walk_successors`/`init_walk_successors` into one `walk_successors(.., phase)` (per-phase output byte-identical).
- Dropped the `is_root` parameter from the tracked per-model queries so a submodule's fragments are shared between the diagnostic pass and assembly (was 2x cache entries).

### Capability / correctness
- `assemble_module`/`assemble_simulation` are now salsa-tracked (Arc results): a no-op recompile is a pure cache hit instead of re-running all module assembly.
- `macro_registry_build_error` is now derived by a tracked query rather than precomputed-and-stored on the input (recomputed every sync before); typed-error byte-identity proven, including for non-canonical macro names.
- `PersistentSyncState` moved into `SimlinDb` (new `db.sync`/`sync_staged`/`restore` API); the patch stage/commit/rollback contract is preserved exactly and libsimlin's two-lock dance collapses to one.
- Stdlib model salsa inputs are now built once per db session (handles stable across syncs) instead of re-walked every sync.
- **Bug fix:** 2-arg `RAMP(slope, start)` read `final_time` via a layout-relative `LoadVar`, which is correct only at the root model and silently breaks (drops the fragment) inside a submodule. Now reads it as a `LoadGlobalVar`. Dormant in the corpus (no test used 2-arg RAMP in a submodule); discovered while removing `is_root`.

### Descoped
- An explicit `ElementOrderedScc` IR was considered but descoped: `ResolvedScc` already is that minimal `{members, element_order, phase}` triple, so extracting a new type would be a pure rename of the hardest/most-tested code for no comprehension gain. Captured the intended clarity as a one-sentence rustdoc on `ResolvedScc` instead.

## Test plan
- [x] Full `cargo test --workspace` (via pre-commit hook on every commit; HEAD green).
- [x] `tests/simulate` + `tests/simulate_systems` numeric corpora (incl. inline wasm-backend parity).
- [x] LTM suites: `simulate_ltm`, `simulate_ltm_wasm` (wasm parity), `db::ltm_*`, `db::element_graph_*`.
- [x] clippy zero warnings; downstream crates (libsimlin, cli, serve, mcp-core, pysimlin) build.
- New behavior-guard tests added: assembly no-op cache-hit, is_root offset-machinery lockstep (with a SMOOTH implicit-var fixture + mutation-checked), non-canonical macro message identity, the 2-arg-RAMP-in-submodule regression, and a rewritten cross-thread issue-#297 concurrency test.

## Follow-ups (filed)
- #642 — salsa::Update over f64/NaN PartialEq on compiled bytecode (benign missed-backdate)
- #643 — `compile_project_incremental` deep-clones `CompiledSimulation` out of the Arc
- #644 — two offset machineries can disagree on submodel size with a table-only var (latent off-by-N)
- #645 — `db/tests.rs` near the per-file line cap


bpowers:
## Benchmark: C-LEARN v77 compile, main vs this branch

32-core Linux box, idle. Criterion 0.5.1, 100-sample, back-to-back on the same machine (separate worktree target dirs), each side running its own version of `benches/compiler.rs`. Medians:

### Cold compile (C-LEARN)
| Benchmark | main | branch | Δ |
|---|---|---|---|
| `mdl_parse/clearn` | 31.19 ms | 30.89 ms | -1.0% (wash) |
| `bytecode_compile/clearn` | 432.0 ms | 423.3 ms | **-2.0%** |
| `full_pipeline/clearn` | 467.6 ms | 457.0 ms | **-2.3%** |

Cold compile is slightly faster, not slower: the Source* mirror removal + interned `ModuleInputSet` keys + build-stdlib-once edge out the added Arc wrap/clone in `compile_project_incremental`. Parse is untouched (noise).

### Incremental recompile (synthetic 100-var chain models)
| Benchmark | main | branch | Δ |
|---|---|---|---|
| `salsa_incremental/equation_edit/incremental/100` | 90.9 µs | 37.4 µs | **2.43x faster** |
| `salsa_incremental/add_remove/incremental_remove/100` | 100.2 µs | 38.3 µs | **2.61x faster** |
| `salsa_incremental/add_remove/incremental_add/100` | 1.318 ms | 1.215 ms | -7.8% |

The refactor's payoff lands on the warm-cache incremental path it was designed for (assembly memoization + interned keys + stdlib-once): single-variable edit/remove recompile ~2.4-2.6x faster. The add path gains less (~8%) since it forces more re-layout/assembly. Cold C-LEARN can't reuse a warm cache, hence only the ~2% there.

Reproduce: `cargo bench --bench compiler -- clearn` and `... -- salsa_incremental`.

Issue #667, opened 2026-06-02
engine: test_rank_and_filter_truncates_to_max_loops builds 250 loops to trip production MAX_LOOPS=200


bpowers:
## Problem

`test_rank_and_filter_truncates_to_max_loops` in `src/simlin-engine/src/ltm_finding.rs` (around line 3740) builds `MAX_LOOPS + 50` = 250 `FoundLoop`s to verify truncation against the **production** `MAX_LOOPS = 200` threshold:

```rust
let stock_names: Vec<String> = (0..MAX_LOOPS + 50)
    .map(|i| format!("stock_{i:04}"))
    .collect();
let mut loops: Vec<FoundLoop> = (0..MAX_LOOPS + 50) ...
```

This violates the documented rule in [docs/dev/rust.md](https://github.com/bpowers/simlin/blob/main/docs/dev/rust.md) ("Do not test production threshold gates by building fixtures large enough to trip them -- use a test-only override and a tiny fixture").

## Why it's an easy fix now

The test-only override already exists: `MaxLoopsGuard` + `MAX_LOOPS_OVERRIDE` (the `#[cfg(test)]` thread-local consulted by `effective_max_loops`/`max_loops_for_call`) were added in commit 76db318d. So the migration is trivial: wrap the test in a `MaxLoopsGuard` with a small cap (e.g. 3) and build a tiny fixture (cap + a few loops) instead of 250.

## Severity

Low (pure test hygiene). The test is cheap today (<1ms), so this is not a test-time-budget problem -- it is a consistency-with-the-documented-pattern cleanup, and it future-proofs against the fixture cost scaling if `MAX_LOOPS` is ever raised.

## Components affected

- `src/simlin-engine/src/ltm_finding.rs` (`test_rank_and_filter_truncates_to_max_loops`)

## Discovery context

Identified during an LTM-hardening session while reviewing the `ltm_finding.rs` test suite against the `MaxLoopsGuard` override that commit 76db318d introduced.

Issue #660, opened 2026-06-02
ltm: analyze_model swallows compile errors and returns an empty ModelAnalysis silently


bpowers:
## Problem

`run_ltm_pipeline` in `src/simlin-engine/src/analysis.rs` (around line 161) does:

```rust
let compiled_sim =
    crate::db::compile_project_incremental(db, source_project, &canonical_name).ok()?;
```

On **any** compile failure it `.ok()?`s to `None`, and `analyze_model` then returns an empty `ModelAnalysis` (no loops, intact model snapshot) with no explanation. The rustdoc on `analyze_model` documents this as intended ("Returns gracefully on simulation failure (empty loop fields, model snapshot intact)").

This was a benign graceful-degradation pattern until GH #486 landed the hard-fail compile guard for LTM + non-Euler integration (`assemble_simulation` returns an `Err` when `ltm_enabled` and the effective method is RK2/RK4 with stocks present). Now `analyze_model` on an RK4 model returns "no loops" silently -- the actionable, specific Euler error from the assemble path is discarded by the `.ok()?`, so the caller cannot tell "this model genuinely has no loops" apart from "this model could not be compiled for LTM analysis."

## Why it matters

Diagnostic/UX gap. The MCP analyze surfaces are the primary consumers:
- `src/simlin-mcp-core` `read_model.rs` (around line 108)
- `src/simlin-mcp-core` `edit_model.rs` (around line 320)

An AI agent or CLI caller asking "what loops does this model have?" on an RK4+LTM model gets an empty result and no hint that the real issue is the non-Euler integration method. The error is genuinely actionable (switch to Euler, or disable LTM) but is thrown away.

## Components affected

- `src/simlin-engine/src/analysis.rs` (`run_ltm_pipeline`, `analyze_model`)
- `src/simlin-mcp-core/src/tools/read_model.rs`, `edit_model.rs` (consumers)

## Possible approaches

- Have `run_ltm_pipeline` / `analyze_model` distinguish "compiled, no loops" from "failed to compile" and surface the compile `Err` string (or a structured diagnostic) instead of collapsing both to an empty `ModelAnalysis`.
- At minimum, propagate the assemble-stage `Err` message into `ModelAnalysis` so the MCP analyze surfaces can show it.

## Severity

Low-medium (UX / diagnostic visibility; not a correctness bug -- the result is empty, not wrong).

## Discovery context

[...]

Issue #652, opened 2026-06-01
LTM: raw link scores are unusable for ranking; relative link scores not exposed


bpowers:
## Summary

`Sim.get_links()` returns each link's **raw** LTM link-score time series. The intuitive thing a user does -- rank links by `Link.average_score()` magnitude to find "the most important links" -- surfaces numerically degenerate links instead of meaningful ones. There is no relative (normalized) link score exposed, even though the engine computes everything needed to derive one.

## Observed behavior

On a mid-size model run, the top links by `|average raw score|` had magnitudes like `6.4e22` and `1.1e16`, all on links *between near-constant policy/parameter variables*. These are not the important links -- they are an artifact of the link-score formula.

This is inherent to the LTM link-score definition: the denominator is the change in the **target** variable, so links into variables that barely change (near-constant parameters, equilibria) produce astronomically large scores. The LTM literature notes exactly this -- raw scores "can become very large ... near equilibria where denominators approach zero." Consequently, raw link scores are **not comparable across different target variables**, and ranking links globally by raw magnitude is meaningless.

## The established fix: relative link score

The LTM literature's answer is the **relative link score**: for each target variable, normalize a link's score by the sum of `|link scores|` of *all inputs into that same target* at each timestep. This yields a `[0, 1]` value -- "the fraction of this variable's change attributable to this input" -- which is comparable across the model and is the correct ranking key.

References: Schoenberg 2020, section 4; the LoopX link-thickness discussion in the thesis.

The engine already computes the per-link scores and knows each link's `to` target, so the relative score is derivable without new simulation machinery -- it just isn't exposed.

## Why it matters

- **Correctness / usability**: the most natural user workflow (rank links by importance) currently produces misleading results. Users will draw wrong conclusions about which causal links drive model behavior.
- **Consistency with loop-level work**: loop scores already have a relative (partition-normalized) counterpart (`rel_loop_score`); links have no equivalent exposed, leaving the link-level story incomplete.
- **Developer experience**: this is a core part of the LTM value proposition ("see which links/loops drive behavior"), so the gap directly undercuts the feature.

## Component(s) affected

- `src/simlin-engine` -- link-score analysis (where per-link scores and their `to` targets are available; `analyze_links_core` or equivalent).
- `src/libsimlin` -- FFI surface for `get_links`.
- `src/pysimlin` -- `Sim.get_links()` / `Link` (`average_score()` docstring, and the new relative-score surface).

## Possible approaches

Either:

[...]

Issue #649, opened 2026-06-01
ltm: enumerate_module_pathways has no cap on the number of pathways (exponential synthetic-var blowup)


bpowers:
## Problem

`CausalGraph::enumerate_module_pathways` (`src/simlin-engine/src/ltm/graph.rs:550`) calls `find_all_simple_paths(input_port, output_name, node_count)` (line 589) with a **depth** cap (`node_count`) but **no cap on the number of simple paths returned**. The inner DFS `dfs_simple_paths` (`graph.rs:518-546`) enforces simple-path semantics via a visited set and bounds recursion depth at `max_depth`, but pushes *every* discovered simple path onto the result `Vec` unconditionally.

The number of simple paths between two nodes can be exponential in graph size. A module body that is a chain of N diamond patterns has `2^N` simple paths from input to output, even though every such path is short and the depth cap never trips.

Each returned pathway becomes one LTM synthetic variable: `model_ltm_variables` emits a `$⁚ltm⁚path⁚{input}⁚{idx}` var per pathway (`src/db/ltm/mod.rs:696-735`). So a pathological module body could generate millions of synthetic variables and exhaust memory.

## Why it matters

This is a **defense-in-depth gap**. A related exponential-memory bug (the LTM composite-score equation text `generate_max_abs_chain_str` building `O(2^n)`-sized equation text) was discovered and is being fixed on the `layout-hill-climb` branch by folding the strongest-pathway composite through O(1)-sized accumulator helper variables. That fix stops the *per-equation* blowup, but the *number* of pathways feeding it is still unbounded — so a pathological module body could still mint an exponential number of synthetic variables and OOM even after the composite-equation fix lands.

The pathway count is **model-controlled and unbounded**: it depends entirely on the topology of a user-defined (or stdlib macro) module body. There is no production threshold protecting against an adversarial or accidentally-dense module.

## Real-world evidence

`covid19-us-homer`'s SSTATS macro module (`test/metasd/covid19-us-homer/homer v8/Covid19US v8.mdl`) already produces 270+ pathways per input port. That is manageable today, but it is the same enumeration with no upper bound — a denser module body scales super-linearly.

## Component(s) affected

- `src/simlin-engine/src/ltm/graph.rs` — `enumerate_module_pathways` / `find_all_simple_paths` / `dfs_simple_paths`
[...]

PR #668, opened 2026-06-02
LTM hardening: close out the top tracked gaps from epic #488

### LINKED ISSUE #311: ltm-augment: parse fallback can silently break ceteris-paribus logic
## Summary
If partial-equation parsing fails during ceteris-paribus equation generation, the code silently falls back to lowercased original equation text without wrapping non-source dependencies in `PREVIOUS()`.

## Why this is a problem
Ceteris-paribus re-evaluation is central to LTM link score definition (`Δ_x(z)`). Silent fallback can produce equations that no longer compute partial change correctly, while still appearing to succeed. That creates hidden attribution errors.

Reference Section 3.1 requires re-evaluating with one source current and all other inputs previous. A fallback that omits dependency wrapping violates the core algorithm.

`build_partial_equation()`:
- returns `equation_text.to_lowercase()` when parser returns `Err` or `Ok(None)`.

Location:
- `src/simlin-engine/src/ltm_augment.rs` (`build_partial_equation`)

## Impact
- Silent correctness degradation in edge-case expressions.
- Difficult-to-debug mismatch between expected and actual link-score attribution.

## Success criteria
1. Eliminate silent fallback that changes semantics.
2. On parse/transform failure, either:
   - return a structured error and fail augmentation, or
   - use a semantics-preserving fallback that still enforces ceteris-paribus wrapping.
3. Add tests that force the failure path and assert deterministic, explicit behavior (error or guaranteed-correct fallback).
4. Add diagnostic context in errors (variable name, original equation text).

## Notes
This issue is distinct from the intentional flow-to-stock timing decision.

bpowers:
Closes out the highest-priority gaps from the LTM tracking epic #488, one reviewed commit per gap. Every fix was implemented TDD-first by a dedicated agent and adversarially reviewed by a separate one (red-test verification against the parent commit, full pre-commit suite on every commit); review findings were folded back in as follow-up commits.

## Correctness fixes (silent wrong results)

- **#311** (`81f74818`): the ceteris-paribus partial-equation builders silently fell back to un-wrapped equation text on parse failure, collapsing link scores to magnitude 1. Now: structured error, the variable is skipped, and a Warning names it (loud-failure pattern).
- **#486** (`0fb59f52`, `b4437978`): LTM + RK2/RK4 silently produced mathematically wrong flow-to-stock scores. Now rejected at `assemble_simulation` with an Euler-assumption message, resolving the effective method exactly as the VM does (main model's override, else project specs) and only when the assembled sim actually has stocks.
- **#542** (`58a5667e`): one NaN loop score poisoned every relative loop score in its cycle partition. NaN summands are now excluded from partition denominators (matching the discovery path); Inf is deliberately kept (real dominance-inflection signal).
- **#525 crash half** (`77be8aec`): codegen's array-view index walk `.unwrap()`-panicked on valid arrayed models with LTM enabled, defeating the graceful-stub design. Errors now propagate; the fragment stubs to 0 with a Warning. The mis-scoring root cause stays open on #525 with a full diagnosis posted there.
- **#541** (`1736f9aa`, `206f8b93`): nested `PREVIOUS` of a bare arrayed name in an A2A equation hard-failed plain compilation. Fixed via arrayed `ApplyToAll` helpers, scoped (after a bisect caught a C-LEARN regression the first review missed) to subscript-free arguments so mapped-dimension subscripts keep the per-element scalar path. All three ignored C-LEARN tests verified green in debug, plus the release-mode pinned-loop gate.

## Diagnostics and UX

- **#466** (`116e8bc9`, `42b24e2a`): LTM diagnostics (auto-flip, fragment failures, #311 warnings) were unreachable through `simlin_project_get_errors`. Now surfaced via an `ltm_requested` latch with LTM scoped as an analysis overlay: project compilability is assessed LTM-off, so LTM-incompatibility never flags a fine model as broken. `SimlinErrorDetail` gains a `severity` field threaded through pysimlin and the TS engine, so warnings stop displaying as errors.

## Determinism and ranking

- **#497** (`1539329d`): loop IDs could flap across processes for multidigraph sibling cycles; the sort key now tiebreaks on the canonical edge sequence in both ID-assignment sites.
- **#543 + #310** (`76db318d`): discovery ranked and truncated loops by raw absolute magnitude (degenerate huge scores dominated) and truncated before the partition-aware filter. Now: filter first, rank by mean partition-relative contribution over active steps, truncate second, content-based tie-breaking.

## Capability and robustness

- **#652** (`fc39df76`): relative link scores (per-target normalization) exposed end to end -- engine core shared by the VM and wasm analyze paths, FFI struct extension, pysimlin `Link.relative_score` + ranking helper, TS parity -- with docstrings warning that raw scores are not comparable across targets.
- **#649** (`63b83cdf`): module pathway enumeration was unbounded (2^N simple paths -> OOM). Now work-bounded and deterministic with a budget (8192, above the measured corpus max of 6744 on covid19-us-homer), truncation flag, and Warning.

## Housekeeping

- `f2bc9b28`: refreshed stale discovery-test docs (World3 single-timestep test un-ignored -- it runs in ~0.6s post-#540-fix) and accumulated review nits.

## Related issue activity from this effort

- Closed as already-fixed on main (verified empirically): #546, #548, #647, #540, #587.
- Filed from discoveries during review: #660-#667.
- This PR resolves: #311, #486, #466, #542, #497, #543, #310, #649, #652, #541 (left without closing keywords so they can be closed with verification notes on merge).


bpowers:
@codex review

Issue #701, opened 2026-06-03
ltm: discovery-mode agg_recovery_truncated signal never reaches FFI/pysimlin/TS (asymmetry vs exhaustive)


bpowers:
## Summary

The discovery-mode cross-agg recovery truncation signal introduced for the #696 fix (commit `ebd9a038`, \"ltm: recover cross-agg reducer loops in discovery mode\") is exposed **only** on the in-process Rust `ModelAnalysis` API. It does **not** reach the libsimlin FFI, pysimlin, or the TypeScript engine surface. A large reducer-in-feedback model that auto-flips to discovery mode and clips cross-agg loop recovery at `MAX_CROSS_AGG_LOOPS` (256) therefore hands FFI/pysimlin consumers an incomplete loop list with **no completeness signal** -- the exact \"looks complete but isn't\" failure #696 set out to eliminate, now reintroduced one layer out.

## The asymmetry

- **Discovery mode**: `DiscoveryResult::agg_recovery_truncated` (`src/simlin-engine/src/ltm_finding.rs:205`) is threaded into `ModelAnalysis::agg_recovery_truncated` (`src/simlin-engine/src/analysis.rs:60`, set at `:136`/`:260`). It stops there -- it is a plain field on the Rust struct with no FFI/diagnostic egress.
- **Exhaustive mode**: the analogue `LtmVariablesResult::agg_recovery_truncated` (`src/simlin-engine/src/db.rs:377`) reaches FFI callers via the salsa `Warning` -> `simlin_project_get_errors` channel, so exhaustive-mode consumers *do* learn their loop list was clipped.

So two callers running the same conceptual analysis on the same model -- one that stays exhaustive, one that auto-flips to discovery because the SCC is large -- get asymmetric completeness signals. The discovery caller, which is the one that hit a scale limit and is *most* likely to have clipped loops, is the one left in the dark.

## Concrete gap in the FFI/pysimlin layer

- `SimlinDiscoveryResult` (`src/libsimlin/src/ffi.rs:233`) carries a single `truncated` bool, populated at `src/libsimlin/src/analysis.rs:739` from `analysis.truncated` -- the **time-budget** truncation only. There is no field for cross-agg recovery truncation.
- pysimlin `Analysis.truncated` (`src/pysimlin/simlin/analysis.py:433`) is likewise mapped from `result_ptr.truncated` (`src/pysimlin/simlin/model.py:806`) -- same time-budget-only meaning.

[...]

Issue #700, opened 2026-06-03
build: dev-init.sh in a git worktree re-points the SHARED pre-commit hook at the worktree, silently disabling the hook on main when the worktree is removed


bpowers:
## Problem

`scripts/dev-init.sh` installs the project's pre-commit hook by symlinking the shared git hooks entry to the checkout's copy of `scripts/pre-commit`. When run inside a **linked git worktree** (e.g. the `.claude/worktrees/agent-*` worktrees Claude Code creates), it re-points the *shared* hook at the *worktree's* transient copy of the script. When that worktree is later removed, the symlink dangles and git **silently skips the pre-commit hook for all subsequent commits in the main checkout** -- with no warning. This defeats the project's "lean on the pre-commit hook" workflow.

This actually happened during an LTM review session: a commit on `main` bypassed fmt/clippy/tests because five agent worktrees had each run `dev-init.sh`.

Observed dangling state:

```
.git/hooks/pre-commit -> /home/bpowers/src/simlin2/.claude/worktrees/agent-ae83caaeb128b9a49/scripts/pre-commit
```

## Root cause

In `scripts/dev-init.sh`:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
...
# 1. Git hooks
if HOOK_PATH="$(git rev-parse --git-path hooks/pre-commit 2>/dev/null)"; then
    DESIRED="$REPO_ROOT/scripts/pre-commit"
    if ! { [ -L "$HOOK_PATH" ] && [ "$(readlink "$HOOK_PATH")" = "$DESIRED" ]; }; then
        mkdir -p "$(dirname "$HOOK_PATH")"
        rm -f "$HOOK_PATH"
        ln -s "$DESIRED" "$HOOK_PATH"
    fi
fi
```

The two sides of the symlink resolve to different repositories when run inside a linked worktree:

- `REPO_ROOT` resolves to the **worktree's** checkout (the directory containing this copy of `dev-init.sh`), so `DESIRED` points at `<worktree>/scripts/pre-commit`.
- `git rev-parse --git-path hooks/pre-commit` resolves to the **shared common** `.git/hooks/pre-commit` -- linked worktrees share the main repo's hooks directory (`core.hooksPath` resolves to the main `.git/hooks`).

[...]

Issue #699, opened 2026-06-03
ltm-finding: discovery zero-edge-exclusion 'loses no loop' claim is overstated; baton-passing loops are invisible


bpowers:
## Summary

`SearchGraph` in `src/simlin-engine/src/ltm_finding.rs` excludes zero-score (and NaN) edges from each per-timestep search graph, justified by an in-code completeness claim that is too strong. Three comments assert variants of "any loop that ever matters has all-nonzero links at the step where it does, and remains discoverable there":

- `src/simlin-engine/src/ltm_finding.rs` ~L204-208 (`SearchGraph::from_edges`): "Any loop that ever matters has all-nonzero links at the timestep where it does, and remains discoverable there."
- `src/simlin-engine/src/ltm_finding.rs` ~L1277-1281 (`step_adj`): "any loop that ever matters has all-nonzero links at the step where it does -- where it remains discoverable (GH #647)."
- `src/simlin-engine/src/ltm_finding.rs` ~L3071-3074 (`test_inactive_loop_found_at_active_step` doc): "Discovery runs at every timestep, so per-step exclusion of inactive edges loses no loop that ever matters."

The claim is **false** for a "baton-passing" loop whose links are each active at different times but are never *all simultaneously* nonzero at any single (sampled) timestep. The exclusion is correct per-step (a loop with a zero-score link genuinely has loop score 0 there), but the completeness conclusion does not follow: there is no requirement that all of a real loop's links be simultaneously nonzero at *some* timestep.

## Why this matters

- **Correctness of attribution / overstated guarantee.** The comments imply discovery is complete with respect to "loops that ever matter," which is not what the algorithm delivers. A class of dynamically-real loops is silently invisible to discovery while exhaustive mode reports them.
- **A test enshrines the gap with a misattributed cause.** `discovery_decoupled_stocks` in `src/simlin-engine/tests/simulate_ltm.rs` asserts `found.len() == 2` (the two self-loops) and its comment blames the *removed* `best_score` pruning ("the strong self-loop paths set high best_scores on shared nodes ... pruning the weaker cross-stock path"). The real cause is the zero-edge exclusion plus non-overlapping link activity, not best-score pruning. The comment is stale/wrong.

## Empirical demonstrations

[...]

Issue #698, opened 2026-06-03
ltm: discovery mode can report the wrong loop polarity (sign error) through a multi-output module


bpowers:
## Summary

In **discovery mode**, the `input -> module` link score for a loop that traverses a multi-output-port submodel can be the **wrong polarity (inverted sign)** relative to the exhaustive-mode score for the same loop. This is not merely an approximation or a cross-process nondeterminism gap (cf. #680); it is a deterministic sign error on the structural interpretation of a loop, and discovery is the mode that large / auto-flipped models fall into.

## Root cause

PR #684 added a per-exit-port pathway-selection override (`compute_module_link_overrides` in `src/simlin-engine/src/db/ltm/mod.rs`) so that, in **exhaustive** mode, a loop through a multi-output module scores against the pathway ending at the exit port the loop actually traverses. That override is keyed by `(loop_id, link_index)` and threaded into the loop-score equation builder.

Discovery mode emits **no** loop-score variables (loops are ranked post-simulation by the strongest-path search), so there is nothing to override. The base `input -> module` link score in discovery mode is therefore the **composite**, which `max-abs`-selects across ALL output ports. For single-dependency exit ports every pathway normalizes to magnitude exactly 1, so the `>=` first-index tie-break picks an arbitrary port -- which may be a different port than the one the loop traverses, and that port can have the opposite sign.

The design doc (`docs/design/ltm--loops-that-matter.md`, "Discovery-mode asymmetry", ~line 474) currently frames this as a "paper-faithful per-edge approximation" / "asymmetry", which understates the consequence: the resulting score can be the wrong sign, not just a less-precise magnitude.

## Empirical repro

A submodel exposing two outputs:

- `pos = input * 0.02`
- `neg = -input`

A parent with a feedback loop that reads `m·pos`, and a side variable that reads `m·neg`:

- Exhaustive raw loop score = **+1.0** (correct: reinforcing).
- Discovery `signal -> m` link score = **-1.0** (the composite selected the `neg` exit port), inverting the loop's polarity.

## Why it matters

Polarity is the primary structural story LTM tells a modeler ("this is a reinforcing loop" vs. "balancing loop"). A flipped sign on a loop through a multi-output submodel can mislead structural interpretation, and discovery is exactly the regime large/auto-flipped models run in -- so this is not confined to a rare edge. Correctness, not just precision, of a user-facing label.

## Component(s) affected

[...]

PR #705, opened 2026-06-04
ltm: discovery-mode correctness fixes from deep review

bpowers:
A multi-agent deep review of the LTM implementation against the four papers empirically validated the scoring core (paper Tables 1-3, chain-rule and aggregation invariance, isolated-loop +/-1, Bass inflection crossing all reproduce exactly) and found the residual risk concentrated in discovery mode. This PR lands the documentation honesty fixes plus three discovery-mode fixes, each implemented TDD-first and passed through an independent adversarial review loop (two of the three needed a fix round after review caught real gaps).

## Commits

- **ltm: document discovery completeness gaps and dt scaling** -- softens the overstated "loses nothing" zero-edge-exclusion claims (the baton-passing limitation, GH #699), corrects stale best_score comments that misattributed the decoupled-stocks miss, fixes the design doc's false claim that discovery flows through build_element_level_loops, and documents the load-bearing `time_step` factor under divergence 6.

- **build: resolve pre-commit hook target against the main checkout** -- dev-init.sh in a linked worktree re-pointed the shared `.git/hooks/pre-commit` symlink at the worktree's copy, leaving it dangling (and git silently skipping the hook) once the worktree was removed. GH #700.

- **ltm: fix discovery-mode wrong loop polarity through multi-output modules** (GH #698) -- discovery's input->module link score was the composite, whose max-abs tie-break picks an arbitrary output port (single-dependency pathways all normalize to magnitude 1), empirically inverting loop polarity (+1 exhaustive vs -1 discovery). Discovery now applies the same per-exit-port pathway selection exhaustive mode's PR #684 override applies, during the post-discovery score recompute; wired into the production analyze_model path via causal_graph_from_element_edges_with_modules (review round 1 caught the first attempt being a production no-op). Regression tests drive analyze_model end-to-end for both the aux-reader and module->module exit-port arms.

- **ltm: recover cross-agg reducer loops in discovery mode** (GH #696) -- the discovery DFS forbids node revisits, so cross-element loops through a hoisted reducer's `$:ltm:agg` node (non-elementary) were structurally unreachable and silently dropped; worst exactly when a reducer-in-feedback model auto-flips to discovery. Discovery now stitches discovered single-agg petals using the same combinatorial core exhaustive mode uses (extracted into stitch_cross_agg_petals / collect_agg_petals, shared by both paths). On the 3-element repro discovery now returns the same 7 loops as exhaustive with score series matching by value; C-LEARN loop count and discovery runtime unchanged. Truncation surfaces as DiscoveryResult::agg_recovery_truncated (FFI surfacing tracked as GH #701).

- **ltm: correct discover_reducer_feedback helper docstring** -- the test helper claimed production wiring it does not use.

- **ltm: add in-engine runtime loop-polarity reclassification primitive** (GH #679) -- reclassify_loops_from_results reads a loop's loop_score series and applies from_runtime_scores, keeping loop ids stable and falling back to structural polarity for never-active loops. Honest scope (per review): pysimlin Run.loops already reclassified via its own Python mirror (now pinned by a regression test); libsimlin/WASM/TS get_loops is intentionally structural-only (no Results in hand) -- wiring that surface is GH #495's; the primitive has no production caller yet.

## Verification

Every commit passed the full pre-commit suite (Rust fmt/clippy/tests, TS lint/build/tsc/tests, pysimlin), and the full suite was re-run on the rebased result. The ignored C-LEARN LTM tests (clearn_pinned_climate_loop_is_scored, clearn_ltm_discovery_compiles, clearn_with_ltm_simulates_model_vars_identically, discovery_clearn_matches_vm_wasm) were run explicitly in release mode for both behavior-changing commits. Independent reviewers verified the discriminating power of the new regression tests by stub-reverting the fixes and watching them fail.

Related issues: GH #696, #698, #699, #700, #701, #679, #495.


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

Issue #760, opened 2026-06-10
engine: @simlin/engine never binds simlin_analyze_get_loops_runtime (runtime Rux/Bux polarity and real confidence unreachable from TS)


bpowers:
## Problem

`@simlin/engine` (the TypeScript WASM surface) never binds `simlin_analyze_get_loops_runtime`, so TS consumers can never observe runtime-reclassified loop polarity (`MostlyReinforcing`/`MostlyBalancing`, "Rux"/"Bux") or a real `polarity_confidence` -- even though the enum variants and the `Loop.polarityConfidence` field already exist in `src/engine/src/types.ts` (the #495 type plumbing).

Concretely, `src/engine/src/direct-backend.ts` (~line 424) binds only the structural entry point:

- `simlin_analyze_get_loops` (`src/libsimlin/src/analysis.rs` ~474) -- model-only, pre-simulation labels; `polarity_confidence` is hardcoded `1.0` (all links signed) or `0.0` (any link unknown), and Rux/Bux are structurally unreachable (#679).

The runtime entry point that does the honest post-simulation reclassification exists in libsimlin:

- `simlin_analyze_get_loops_runtime` (`src/libsimlin/src/analysis.rs` ~557) takes a run `SimlinSim` and reclassifies via `LoopPolarity::from_runtime_scores` over the simulated loop-score series (the #679 surface).

pysimlin binds it (`Sim.get_loops_runtime`, and `Run.loops` consumes the same reclassification), so Python sees honest post-sim labels. TS does not, so the diagram/web UI cannot render them.

## Why it matters

Capability/parity gap (severity low-medium): Python consumers get honest Rux/Bux labels with real confidence values; TS/web consumers are stuck with pre-simulation structural labels and a degenerate 1.0/0.0 confidence. This is the same #495-follow-on plumbing class as #701 (runtime/discovery signals that exist in the engine and FFI but never reach TS).

It is also a prerequisite for #731 (diagram: render Rux/Bux with confidence-gray treatment): #731's body states "the data is now available to the diagram via the `@simlin/engine` `Loop` type", but that is only true of the *types* -- no binding actually delivers Rux/Bux or a non-trivial `polarityConfidence` to TS at runtime. #731 cannot render honest post-simulation polarity until this lands.

## Components affected

- `src/engine/src/direct-backend.ts` (bind the runtime entry point; expose e.g. `Sim.getLoops()` / a runtime-loops method on the sim handle)
[...]

Issue #759, opened 2026-06-10
ltm-augment: partial builders PREVIOUS-wrap dimension-name subscript indices (Wildcard and Bare shapes), silently stubbing scores


bpowers:
## Summary

The ceteris-paribus partial builders in `src/simlin-engine/src/ltm_augment.rs` (`build_partial_equation_shaped` -> `wrap_non_matching_in_previous` -> `wrap_index_non_matching_in_previous`) PREVIOUS-wrap a bare **dimension-name** identifier sitting **inside a subscript index**. Originally filed against the live `Wildcard` reference shape (`SUM(matrix[State, *])` becoming `SUM(matrix[PREVIOUS(state), *])`), the defect is **also reachable through a mundane Bare-shape apply-to-all equation with no reducer at all** (see "Demonstrated reachable, fully silent path" below), where the dimension name leaks into the co-source `other_deps` set and the frozen co-source's subscript index gets wrapped.

`PREVIOUS` of a dimension name is meaningless: the fragment fails compilation and the synthetic link-score variable is silently stubbed -- the #546/#548/#587 hazard class ("LTM fragment compiler can't handle a shape -> synthetic var stubs to 0 -> degraded score masquerades as inactive"). **Correction to the original filing**: when the doomed expression routes through a synthesized *implicit helper*, the failure produces **no diagnostic at all** -- `model_ltm_fragment_diagnostics` never covers implicit helpers (the #741 class) -- so the original claim that the defect "is surfaced as an Assembly Warning when reached" is wrong for this route.

## Demonstrated reachable, fully silent path (GH #743 review, raises severity to medium)

Verified independently by an implementor (discovery) and an adversarial reviewer (probe) during GH #743 work on branch `ltm-fix-batch-2` at commit `7012d9f0`:

```
growth[D1] = matrix[D1, c1] * frac[D1]
```

A Bare-shape apply-to-all equation with a **pinned literal co-source index** (`c1`), no reducer involved. The changed-first partial for the `frac -> growth` edge freezes the co-source as

```
PREVIOUS(matrix[PREVIOUS(d1), d2·c1])
```

-- the iterated-dim **name** `d1` is PREVIOUS-wrapped inside the subscript index.

[...]

Issue #756, opened 2026-06-10
engine: executed A2A lowering ignores explicit element maps and resolves positionally (silent wrong results for permutation maps)


bpowers:
## Problem

Explicit dimension element maps (the XMILE `simlin:mapping`/`simlin:elem` vendor extension) are honored by some compiler resolution paths but **ignored by the executed apply-to-all (A2A) lowering, which resolves positionally**. Different resolution paths in the engine disagree about what an element map means, and the path that actually executes is the one that ignores it -- producing silent wrong simulation results.

### Probe

- `Region{a,b}`, `State{s1,s2}` with an explicit element map `s1 -> b`, `s2 -> a` (a permutation, so sizes match and the model compiles)
- `x[a] = 10`, `x[b] = 20`
- `target[State] = x[State] * 1` (the bare form `target[State] = x * 1` behaves identically)

Expected per the element map: `target[s1] = x[b] = 20`, `target[s2] = x[a] = 10`.

Actual: `target[s1] = 10 (= x[a])`, `target[s2] = 20 (= x[b])` -- **positional** correspondence; the element map is silently ignored.

A duplicate-target map (`s1 -> a`, `s2 -> a`) likewise executes positionally (reads `x[a]`, `x[b]`) rather than reading `x[a]` twice.

### The inconsistency

`translate_via_mapping` and the per-element fallbacks DO honor element maps:

- `src/simlin-engine/src/compiler/context.rs` (~line 1697)
- `src/simlin-engine/src/compiler/subscript.rs` (~line 381)

So the engine has two answers to "what does an element map mean", and the executed A2A lowering gives the wrong (positional) one. Consistent with positional-only execution, a many-to-one (different-cardinality) map fails to compile outright -- that sibling failure is tracked as #753.

## Why it matters

- **Silent wrong numbers** for any model using a non-identity explicit element map in an A2A reference. Permutation maps compile cleanly and run wrong -- no diagnostic.
- The `simlin:mapping`/`simlin:elem` vendor extension exists precisely to express element-level maps; either the executed lowering must honor them, or the docs/parser should reject non-positional maps loudly.
- Blocks LTM from using element-map diagonals: the #527 fix on `ltm-core-batch` is being made conservative (broadcast) for non-empty element maps until this is resolved. The `mapped_element_correspondence` rustdoc references this issue as the re-enablement gate.

## Component

[...]

Issue #753, opened 2026-06-10
engine: different-cardinality element-map reference fails to compile (failed to compile fragments)


bpowers:
## Problem

A variable reference that resolves through an explicit element mapping between dimensions of **different cardinality** fails to compile with a hard error:

```
failed to compile fragments for variables: target
```

Fixture shape (legal XMILE via `simlin:mapping`):

- `State{s1,s2,s3}` carries an explicit element mapping to `Region{a,b}` (many-to-one: s1↦a, s2↦a, s3↦b)
- `x` is declared over `Region`
- apply-to-all equation: `target[State] = x[State] * c` — the `x[State]` reference must resolve through the State→Region mapping

Same-size element maps (including reordered/permuted ones) compile fine; the failure is specific to `|State| != |Region|`.

## Why it matters

- This is a legal model shape a user can write (many-to-one element mappings are the natural way to aggregate, e.g. states into regions); the result is a hard compile error rather than a diagnostic or correct code.
- It also blocks end-to-end testing of different-cardinality mapped models in LTM (the #527 element graph handles the shape correctly; only simulation compilation fails).

Severity: low-medium. Hard error, not silent miscomputation.

## Component

`src/simlin-engine` compiler — the `translate_via_mapping` subscript-resolution path is the likely site. Suspect dimension-size unification or per-element offset resolution rejects the size mismatch (same-size maps reduce to a permutation, which works; many-to-one requires per-element offsets that don't form a bijection).

## Possible approach

In the apply-to-all lowering, when a referenced variable's dimension is reached via an element mapping, resolve each iterated element of the equation's dimension to the mapped element's offset in the referenced variable's dimension (per-element indirection), rather than requiring the two dimensions to unify by size. A regression test with the 3→2 many-to-one fixture above (plus a 2→3 reference in the other direction, if meaningful) should gate the fix.

## Context

Pre-existing engine gap (not LTM-specific), discovered during #527 work on branch `ltm-core-batch` at commit a010931e while testing the LTM element-graph diagonal for mapped dimensions.

bpowers:
Sibling issue filed: #756 -- same-cardinality (permutation) element maps DO compile, but the executed A2A lowering ignores the map and resolves positionally, producing silent wrong results. The positional-only execution there is consistent with the different-cardinality compile failure here; a per-element-indirection fix for #756 likely generalizes to resolve this issue too.

PR #761, opened 2026-06-10
ltm: core-algorithm batch of root-caused fixes from epic #488

### LINKED ISSUE #527: ltm: expand_same_element is over-conservative for mapped dimensions (full cross-product instead of the positional-mapping diagonal)
## Summary

`expand_same_element` in `src/simlin-engine/src/db_analysis.rs` (~lines 335-463) -- the helper that projects a same-element / shared-dims LTM element-graph edge into per-element edges -- matches source and target dimensions **by name only** (`to_dim_positions: HashMap<&str, usize>` keyed on `to_dims.iter().map(|d| d.name())`, then `from_to_target_pos = from_dims.iter().map(|d| to_dim_positions.get(d.name()))`). When the element-graph edge runs from a source over one dimension to a target over a *differently-named but mapped* dimension -- e.g. source over `Region={a,b}`, target over `State={s1,s2,s3}` with a `State→Region` mapping (the AC3.5 case in Phase 3 of the LTM arrays hardening epic, where an `x[State]` reference inside an A2A `target[State]` equation now classifies `RefShape::Bare` via `classify_iterated_dim_shape`) -- `Region` and `State` are different names, so `from_to_target_pos` records `None` for the source dimension, `all_mapped` is false, and the **broadcast branch** (lines 423-461) fires: it emits the full `Region × State` cross-product of element edges (`x[a]→target[s1]`, `x[a]→target[s2]`, `x[a]→target[s3]`, `x[b]→target[s1]`, ...) rather than the **positional-mapping diagonal** the `State→Region` mapping defines (`s1↦a`, `s2↦a`, `s3↦b` or whatever the mapping says ⇒ `x[a]→target[s1]`, `x[a]→target[s2]`, `x[b]→target[s3]`).

This is **pre-existing -- not introduced by Phase 3**. Phase 3's `element_graph_mapped_iterated_dim_matches_bare_baseline` test (`src/simlin-engine/src/db_element_graph_tests.rs:~906`) deliberately asserts only that the *new* iterated-dim-subscript form (`target[State] = x[State] * c`) produces the *same* element edges as the *old* bare-reference form (`target[State] = x * c`) -- i.e. no regression -- and its doc comment explicitly notes: *"`expand_same_element` matches dimension names, so a disjoint-named pair like `Region`/`State` is the broadcast case (every source element feeds every target element) for both the subscripted and the bare form; the assertion below pins 'subscripted iterated == bare', not the projection shape."* So the over-conservatism is on record in a test comment but not in an issue.

### Why it matters

Low impact. Over-conservative element edges mean a few extra spurious causal edges in the element graph for mapped-dimension models -- which can slightly inflate which loops/edges get enumerated and scored -- but it is **not a correctness violation of the LTM contract**: more edges than necessary, never fewer (every emitted edge for a true cross-product would have a real link score; the spurious ones get effectively-zero runtime scores, the same way pre-#448 FixedIndex N×N over-approximation did before that was fixed). Only models that actually use dimension *mappings* (a `<dimension ... maps_to>` / `simlin:mapping`) between the source and target of a same-element LTM edge are affected.

### Components affected

- `src/simlin-engine/src/db_analysis.rs` -- `expand_same_element` (~lines 335-463; the by-name `to_dim_positions` lookup and the `all_mapped`-false broadcast fallback), and its callers in `emit_edges_for_reference` (the `RefShape::Bare` arrayed arm, ~lines 214-250)
- `src/simlin-engine/src/dimensions.rs` -- `DimensionsContext::has_mapping_to` / `SubdimensionRelation` (the element-level mapping data a mapping-aware projection would consult)
- `src/simlin-engine/src/db_element_graph_tests.rs` -- `element_graph_mapped_iterated_dim_matches_bare_baseline` (~line 906; would gain a stronger assertion once the projection is mapping-aware)

### Possible approaches

- Make `expand_same_element` mapping-aware: when a source dimension has no same-name target dimension, check whether some target dimension *maps to* it (or it maps to some target dimension) via the `DimensionsContext`, and if so project along the mapping's element correspondence instead of broadcasting. This would need `expand_same_element` (and `emit_edges_for_reference`) to take a `&DimensionsContext` -- which `db_ltm_ir`'s `classify_iterated_dim_shape` already threads in for the AC3.5 recognition, so the data is available at the call site.
- Once mapping-aware, strengthen the `element_graph_mapped_iterated_dim_matches_bare_baseline` test (and/or `db_ltm_ir_tests::ir_mapped_iterated_dim_subscript_is_bare`'s downstream edge check) to assert the *diagonal* edge set, not just "subscripted == bare".

### Discovery context

Identified during the Phase 3 code review of the LTM arrays hardening epic (#488; design plan and per-phase tasks under `docs/implementation-plans/2026-05-11-ltm-arrays-hardening/`, see `phase_03.md` AC3.5 and the `element_graph_mapped_iterated_dim_matches_bare_baseline` test). Pre-existing on `main` -- not a regression; Phase 3's `element_graph_mapped_iterated_dim_matches_bare_baseline` test correctly asserts the new iterated-dim form is no worse than the existing bare form.

### Tracking

Part of LTM tracking epic #488. Related to #448 (FixedIndex N×N over-approximation -- closed; the analogous fix replaced an all-pairs expansion with a per-reference `RefShape` emitter; this is the same flavor of over-approximation but for the mapped-dimension same-element projection rather than FixedIndex), #511/#525 (iterated-dim subscript handling), and the LTM array-support umbrella #273.

### LINKED ISSUE #528: ltm: agg→target link score over-subscripts an arrayed synthetic agg in the broadcast case (agg[D1] feeding to[D1,D2])
## Summary

`emit_agg_to_target_link_scores` in `src/simlin-engine/src/db_ltm.rs` (~lines 4224-4360, nested in `model_ltm_variables`) builds the *agg → target* half of the link score for a reference routed through a synthetic aggregate node. When that synthetic agg is **arrayed** (`AggNode.result_dims` non-empty -- e.g. `$⁚ltm⁚agg⁚0` over `[D1]` minted from a sliced reducer `SUM(matrix[D1,*])` inside an apply-to-all-over-`D1` target equation), the per-target-element link-score *equation* subscripts the agg ident by the **full** target element tuple, but it should subscript by the target element's **projection onto `agg.result_dims`'s axes**.

Concretely, at ~line 4292-4302 there is already a comment acknowledging this:

```rust
// An arrayed agg (`result_dims` non-empty) is element-pinned in the
// per-target-element equation just like an arrayed dep that shares
// `to`'s dims: `$⁚ltm⁚agg⁚0` → `$⁚ltm⁚agg⁚0[<element>]`. This is
// exact when `result_dims` equals `to`'s dimensions (the diagonal
// `agg[d1] → to[d1]` case -- a partial-reduce sub-expression in an
// A2A target over exactly that dim, e.g. `x[D1] = ... + SUM(matrix[D1,*])`);
// for the rarer broadcast case (`agg[D1]` into `to[D1,D2]`) the
// element tuple over-subscripts the agg, which is a known imprecision.
if agg_is_arrayed {
    deps_to_subscript.insert(agg_canonical.clone());
}
```

The link-score *name* is already projected correctly: `agg_name_for_target(element)` (~line 4318) projects the target element tuple onto `result_dim_positions` (the positions of `agg.result_dims` within `to`'s dims) to produce `agg[<slot>]`. It's the *equation body* that is wrong: the agg ident lands in `deps_to_subscript`, so the per-target-element scalar equation (built by `build_partial_equation_shaped` / the per-element machinery) pins it to `to`'s full element tuple `[<element>]` instead of `[<slot>]`. For the diagonal case (`agg[D1] → to[D1]`, `result_dims == to`'s dims) the full tuple **is** the projection, so it's exact. For the broadcast case (`agg[D1]` feeding `to[D1,D2]`, where the agg result is broadcast across `D2`) the equation references `$⁚ltm⁚agg⁚0[d1,d2]` -- an out-of-shape over-subscript of an agg that is only dimensioned over `[D1]`.

## Why it matters

Low. It's a magnitude/shape imprecision in the agg→target link score for that one (rare) broadcast topology -- **not a crash**, and the polarity/sign behavior is unaffected. The common case -- a partial-reduce sub-expression in an A2A target over exactly the reduced-into dimension (`x[D1] = ... + SUM(matrix[D1,*])`, the diagonal) -- is correct and is covered by Phase 4's AC4.2 tests. No required Phase-4 test exercises the broadcast case through `model_ltm_variables`: variable-backed partial reducers (`row_sum[D1] = SUM(matrix[D1,*])`, where `row_sum` *is* the agg) route through `try_cross_dimensional_link_scores`, which already handles partial-reduce naming correctly and does **not** go through `emit_agg_to_target_link_scores`; only an *inline* sliced reducer whose enclosing equation has *more* iterated dims than the reducer's `Iterated` axes would hit this path, which no current model or fixture does. Worth tracking so the imprecision is on record rather than silently relied upon.

## Components affected

- `src/simlin-engine/src/db_ltm.rs` -- `emit_agg_to_target_link_scores` (~lines 4224-4360; the `deps_to_subscript.insert(agg_canonical)` at ~4300, and the per-target-element equation construction below it -- the agg side of the link-score *name* via `agg_name_for_target` / `result_dim_positions` is already correct; the equation body is the gap). The symmetric source side, `emit_source_to_agg_link_scores` (~lines 4106-4208) + `read_slice_rows` / `ReadSliceRow`, is **correct** (it iterates only the read-slice rows with the proper `Iterated`-axis `slot`); this issue is the agg→target half only.
- `src/simlin-engine/src/ltm_augment.rs` -- `build_partial_equation_shaped` / `wrap_non_matching_in_previous` / `classify_expr0_subscript_shape` (the per-element partial builder that consumes `deps_to_subscript` and pins the agg ident to the full target element).
- `src/simlin-engine/src/ltm_agg.rs` -- `AggNode.read_slice` / `AxisRead` / `compute_read_slice` / `result_dims` (the descriptor that records which axes are `Iterated`, i.e. what the projection target is).
- `src/simlin-engine/src/db_analysis.rs` -- `emit_agg_routed_edges` (the element-graph side already projects the target element onto `agg.result_dims` for the `agg[<iterated>] → to[e]` fan-out, the same way `expand_same_element`'s `Bare` arm does -- the fix should mirror that projection on the link-score equation side).

## Possible approaches

- When emitting the agg→target link score, project the target element tuple onto just `agg.result_dims`'s axes (using `result_dim_positions`, which is already computed for the name) before pinning the agg ident in the per-target-element equation -- i.e. pass the projected slot, not the full tuple, into the partial builder for the agg ident specifically (the other shared-dim deps still pin to the full tuple). This is the same projection `emit_agg_routed_edges` / `expand_same_element`'s `Bare` arm does in the element graph.
- This naturally piggybacks on #520 (unify the element-graph and link-score reference-site walkers behind one classification IR): once the link-score emitter consumes the same `ThroughAgg { agg }` routing and `agg.read_slice` the element graph already does, the projection logic exists in exactly one place.
- If the inline-sliced-reducer-with-broadcast topology is judged not worth the extra projection plumbing right now, the diagonal case is at least correct and tested, and this issue documents the gap with a pointer to the comment.

## Discovery context

Identified during the Phase 4 code review of the LTM arrays hardening epic (#488; commits `ec018d0b` / `bfb61d08` on branch `ltm-arrays-hardening`; design plan and per-phase tasks under `docs/implementation-plans/2026-05-11-ltm-arrays-hardening/`, see `phase_04.md` AC4.2). Phase 4 introduced sliced-reducer hoisting and arrayed synthetic aggregate nodes; this is a documented edge case in that freshly-added code path -- not a regression (no arrayed synthetic agg existed before Phase 4). The carve-out comment in `emit_agg_to_target_link_scores` is on record but not yet in an issue.

## Tracking

Part of LTM tracking epic #488. Related to: #514 (sliced-reducer hoisting -- the Phase-4 work that introduced this code path; distinct: #514 is about *whether* sliced reducers are hoisted at all, this is a shape imprecision in the agg→target link score *once* an arrayed synthetic agg exists), #510 (degenerate link score for *disjoint*-dimension arrayed→arrayed edges with per-element target equations -- a different topology and a different code path, `try_cross_dimensional_link_scores`), #520 (unify the reference-site walkers -- the refactor whose IR would make the fix a one-liner), #527 (`expand_same_element` over-conservative for *mapped* dims -- same flavor of "the projection should mirror the element graph" but a different site, `db_analysis.rs`, and a different axis-correspondence question), and the LTM array-support umbrella #273.

### LINKED ISSUE #534: ltm: mapped-dimension sliced reducers are not hoisted into aggregate nodes (compute_read_slice declines a remapped iterated axis)
## Summary

`compute_read_slice` in `src/simlin-engine/src/ltm_agg.rs` (added in Phase 4 / #514) decides whether an inlined array-reducer subexpression is hoistable into a synthetic `$⁚ltm⁚agg⁚{n}` node by computing, per source axis, an `AxisRead ∈ {Pinned(elem), Iterated(dim), Reduced}`. For the `Iterated` case -- a subscript index that names one of the *enclosing* (apply-to-all) equation's iterated dimensions in the position matching the source's declared dimension order -- the index dimension must match the source's row-axis dimension. As resolved by code review (commit `7f837928`), the iterated-index branch now returns **`None`** (not hoistable) when the index matches the source's row axis only *via a dimension mapping* rather than literally being the same dimension -- e.g. `SUM(matrix[State,*])` inside an apply-to-all-over-`State` body, where `matrix` is declared `matrix[Region,D2]` and there is a `State→Region` mapping.

So a *mapped-dimension sliced reducer* falls back to the conservative full-cross-product element-graph path (`Direct` `Wildcard` reference → `emit_edges_for_reference`'s cross-product arm, a `…→{from}` per-shape link score) instead of being hoisted into a synthetic agg with read-slice-driven element edges and per-read-row link scores. (Whole-extent reducers `SUM(matrix[*,*])`, *literally*-same-dim sliced reducers `SUM(matrix[D1,*])` over `matrix[D1,D2]`, and pinned-slice reducers `SUM(matrix[NYC,*])` are all still hoisted -- this is specifically the *mapped*-iterated case that is declined.)

The reason: the `Iterated`-axis-driven agg-routing machinery -- `emit_agg_routed_edges` (`db_analysis.rs`), `emit_source_to_agg_link_scores` / `read_slice_rows`, `emit_agg_to_target_link_scores` (`db_ltm.rs`) -- assumes the agg's result axis (recorded in `AggNode::result_dims`) and the arrayed source's row axis are *literally the same dimension*. Honoring a mapping would require the `Iterated` `AxisRead` (and `result_dims`) to carry *both* the target-equation dim and the source-row dim, and to remap between them in all three emitters (the element-graph fan-out, the source→agg link-score row enumeration, and the agg→target link-score fan-out). Rather than thread a `(target dim, source dim)` pair through everywhere for a case no current model exercises, Phase 4 declined the mapped axis. Removing the mapping handling also left `compute_read_slice`'s `DimensionsContext` parameter dead, so it was dropped (see commit `7f837928`).

## Why it matters

Low. It's a conservatism, never a crash: a mapped-dimension sliced reducer gets less-precise LTM scoring (the conservative full-cross-product element graph + a lumped per-shape link score) instead of the fine-grained per-read-row aggregate edges + per-row reducer link scores. The conservative path is still *correct* (more edges than necessary, never fewer; the spurious ones get effectively-zero runtime scores) -- just imprecise, the same flavor of over-approximation as the pre-Phase-4 wildcard-link-score path for *all* sliced reducers (#514).

## Components affected

- `src/simlin-engine/src/ltm_agg.rs` -- `compute_read_slice` (the mapped-iterated-axis branch that returns `None`; ~the `Iterated` recognition logic, post-#514), `AggNode::read_slice` / `AxisRead` / `result_dims` (the descriptor that would need the `Iterated` variant to carry a `(target dim, source dim)` pair)
- `src/simlin-engine/src/db_analysis.rs` -- `emit_agg_routed_edges` (assumes `result_dims` ⊆ the source's row dims by name; would need to remap an `Iterated` axis from the source's declared dim to the target's iterated dim when fanning `source[<...iterated...>] → agg[<iterated>]` and `agg[<iterated>] → to[e]`)
- `src/simlin-engine/src/db_ltm.rs` -- `emit_source_to_agg_link_scores` / `read_slice_rows` / `ReadSliceRow` (the source→agg per-read-row link-score emitter; the row enumeration would need the mapping to know which source elements map to which agg result slot), `emit_agg_to_target_link_scores` (the agg→target half)

## Possible approaches

- Thread the `(target dim, source dim)` pair through the `Iterated` `AxisRead` (and correspondingly through `AggNode::result_dims` or a parallel structure), and remap between them in the three emitters via `DimensionsContext`'s element-level mapping (`SubdimensionRelation` / `has_mapping_to`). The element-graph side (`emit_agg_routed_edges`) already does a `result_dims`-onto-`to` projection for the broadcast case, and `expand_same_element`'s `Bare` arm is in the process of getting mapping-awareness (#527) -- the same correspondence data is needed here.
- Could plausibly be folded into #527's scope (`expand_same_element` mapping-awareness): both want "make the projection mapping-aware instead of name-only". #527 is the *direct*-reference, element-edge-count side (`db_analysis.rs`); this is the *sliced-reducer-hoisting* side (`ltm_agg.rs`, whether the reducer gets hoisted at all). They share the underlying "honor the dimension mapping in the element correspondence" requirement but touch different code paths -- file as a sibling or extend #527, whichever is cleaner at implementation time.
- If the mapped-dimension-sliced-reducer topology is judged not worth the extra plumbing, the conservative path is at least correct and the carve-out is documented (commit `7f837928`, `src/simlin-engine/CLAUDE.md` under "LTM arrays hardening Phase 4", and `enumerate_agg_nodes`'s carve-out list in `ltm_agg.rs`); this issue records the gap.

## Discovery context

Identified during the Phase 4 code review of the LTM arrays hardening epic (#488; commit `7f837928` on branch `ltm-arrays-hardening`; design plan and per-phase tasks under `docs/implementation-plans/2026-05-11-ltm-arrays-hardening/`, see `phase_04.md` -- the sliced-reducer-hoisting work). The mapped-axis carve-out was the deliberate resolution of a code-review IMPORTANT issue: the `Iterated`-axis machinery assumes the agg result axis and the source row axis are the same dimension, and Phase 4 chose to decline the mapped axis rather than thread a `(target dim, source dim)` pair through all three emitters for an unexercised case. Not a regression -- no sliced reducer was hoisted at all before Phase 4.

## Tracking

Part of LTM tracking epic #488. Related to: #514 (sliced-reducer hoisting -- the Phase-4 work; distinct: #514 was about *whether sliced reducers are hoisted at all* in the un-mapped case, this is the residual mapped-dimension carve-out within that work), #527 (`expand_same_element` over-conservative for mapped dims in *direct* references -- same "honor the mapping in the element correspondence" requirement, different code path: `db_analysis.rs` element-edge count vs `ltm_agg.rs` whether-to-hoist; could be folded together), #528 (agg→target link-score over-subscripts an arrayed synthetic agg -- another arrayed-synthetic-agg edge case from the same Phase-4 work), #520 (unify the reference-site walkers), and the LTM array-support umbrella #273.

### LINKED ISSUE #673: ltm: pin whose only stock is module-internal is wrongly rejected as 'contains no stock'
## Problem

`model_pinned_loops` (`src/simlin-engine/src/db/ltm/pinned.rs`, the `has_stock` validation around line 156) checks the pin's raw, pre-enrichment cycle against the parent-level `model_causal_edges.stocks` set:

```rust
let has_stock = cycle.iter().any(|n| edges.stocks.contains(n.as_str()));
if !has_stock {
    // reject: "pinned loop '...' contains no stock"
}
```

`edges.stocks` contains only **parent-level** stocks. A `DynamicModule` node (e.g. a SMOOTH instance whose internal stock is the loop's only state) is not in that set. So a valid feedback loop such as `aux1 -> smooth -> aux2 -> aux1`, whose only stock lives *inside* the SMOOTH module, fails the check and the pin is rejected with "contains no stock".

## Inconsistency

The same variable set is happily found and scored by the loop **enumerator**, which applies no `has_stock` filter and enriches module-internal stocks via `enrich_with_module_stocks`. The result is that the exact same cycle is **enumerable but unpinnable** -- the discovery path accepts it, the pin path rejects it. That divergence is the core defect.

## Why it matters

- **Correctness / consistency**: a loop that discovery treats as a real feedback loop cannot be pinned, so users (or agents) cannot pin-and-track a loop the enumerator already recognizes.
- **Module support**: this is a concrete gap in pinning loops that route through stdlib modules (SMOOTH/DELAY), which is exactly the case where state hides inside a module.

## Component(s) affected

- `src/simlin-engine/src/db/ltm/pinned.rs` (`model_pinned_loops`, `has_stock` validation)

## Test gap

There are currently **no tests covering a pin through a module at all**.

## Possible approaches

- Treat a `DynamicModule` node in the cycle as stock-bearing for the `has_stock` check (`classify_module_for_ltm` already knows which modules carry internal state), or
- Run module stock enrichment (`enrich_with_module_stocks`) *before* the `has_stock` check so module-internal stocks are visible to the validation.

Either way, add a regression test that pins a loop whose only stock is inside a SMOOTH/DELAY module.

## Discovery context

Found during a design review on 2026-06-02; documented in `docs/design/ltm--loops-that-matter.md` (Pinned Loops section) as of commit cc072973.

Severity: minor.

Related but distinct: #282 (enrichment *misses* cycle-only internal stocks -- opposite failure mode, different code site) and #274 (umbrella "improve/generalize module support").

### LINKED ISSUE #676: ltm: cyclic_orderings' (m-1)!/2 mirror-skip rationale is incorrect; cross-agg loop count is arbitrary
## Summary

`cyclic_orderings` (`src/simlin-engine/src/db/ltm/loops.rs`, ~line 1118; also referenced by `MAX_CROSS_AGG_LOOPS` doc at ~line 1041 and the call site in `recover_cross_agg_loops` at ~line 1351) emits `(m-1)!/2` directed cycles per disjoint `m`-petal subset, with a doc comment justifying the `/2` by claiming mirror reversals are skipped because "a directed cycle and its reverse share a score." A 2026-06-02 design review found that rationale is mathematically incorrect, and that the loop *count* is internally inconsistent against either of its two plausible goals.

This is a follow-up critique of the **resolution** of the now-closed #515 (LTM arrays hardening Phase 5; epic #488 line 58 records the `cyclic_orderings(m)` -- `(m-1)!/2` for m>=3, mirror reversals skipped" design). #515 itself was about the `MAX_AGG_PETALS` cap and per-subset (vs per-ordering) under-enumeration; this issue is specifically about the new `cyclic_orderings` enumeration that landed to fix it.

## The bug in the rationale

The involution actually applied in `cyclic_orderings` is **petal-SEQUENCE reversal with each petal still traversed FORWARD**: `[0, p1, .., p_{n-1}] -> [0, p_{n-1}, .., p1]` (the comment at ~line 1128 even says this). That is NOT the edge-reverse of the kept ordering:

- The genuine edge-reverse of a directed cycle traverses each petal *backward*, and those reverse edges are not even present in the element graph.
- So the skipped orderings are distinct directed circuits, not the reverses of the kept ones. The stated justification ("a directed cycle and its reverse share a score") does not apply to the orderings actually being dropped.

## Deeper: every cyclic ordering of a fixed petal subset shares one score

For a fixed petal subset, **every** cyclic ordering produces the **same edge multiset**: each petal contributes the same `agg->head` / internal / `tail->agg` edges regardless of its position in the concatenation. The loop score is a commutative product over that edge multiset, so all orderings share a single `loop_score`. Therefore the count is inconsistent with both intents:

- For loop **identity** (distinct directed elementary circuits): the truth is `(m-1)!`. The current `(m-1)!/2` **under-counts by half** for m>=3.
- For distinct loop **scores**: the truth is `1`. The current `(m-1)!/2` **over-counts** -- it mints `(m-1)!/2` redundant loop IDs with byte-identical dominance profiles.

The test `cross_agg_loop_recovery_four_petals_enumerates_cyclic_orderings` (`src/simlin-engine/src/db/ltm_unified_tests.rs`, ~line 2975) confirms it currently produces 3 distinct four-petal loops that all share one score.

## Why it matters

- **Severity: minor.** Dominance analysis is unaffected -- the variants carry identical scores, so rankings/post-sim normalization are correct.
- The **comment is actively misleading** about why `/2` is applied (it cites edge-reversal symmetry that does not hold).
- The redundant loop IDs **waste the `MAX_CROSS_AGG_LOOPS = 256` budget**: each disjoint subset consumes `(m-1)!/2` budget slots for loops that are indistinguishable for dominance, so the budget is exhausted faster and genuinely-distinct petal subsets can be truncated earlier than necessary.

## Component(s) affected

- `src/simlin-engine/src/db/ltm/loops.rs` -- `cyclic_orderings` (~1118), its doc comment, `MAX_CROSS_AGG_LOOPS` doc (~1041), the call site in `recover_cross_agg_loops` (~1351).
- `src/simlin-engine/src/db/ltm_unified_tests.rs` -- `cross_agg_loop_recovery_four_petals_enumerates_cyclic_orderings` (~2975) would change with the chosen fix.

## Possible approaches

Decide the intent and make the count consistent with it, fixing the comment either way:

1. **Recommended: emit one representative per disjoint petal subset.** Identical scores make the orderings indistinguishable for dominance, so a single loop per subset is sufficient and it frees `MAX_CROSS_AGG_LOOPS` budget for genuinely-distinct subsets. Replace `cyclic_orderings(m)` with "one canonical ordering."
2. **Or, if full directed-circuit identity is wanted, emit `(m-1)!`** (all rotation classes, no mirror quotient) and correct the comment to drop the false edge-reversal justification.

Either way, rewrite the doc comment so it no longer claims the skipped orderings are edge-reverses sharing a score.

## Discovery context

Identified during a 2026-06-02 design review of the #515 resolution (LTM arrays hardening Phase 5). Algebraic verification of the involution and the per-subset single-score property was done by hand and cross-checked against `cross_agg_loop_recovery_four_petals_enumerates_cyclic_orderings`.

Related: #515 (closed; the resolution this critiques), epic #488.

### LINKED ISSUE #679: ltm: exhaustive-mode loops never get runtime polarity reclassification (Rux/Bux unreachable)
## Problem

`LoopPolarity::from_runtime_scores` (`src/simlin-engine/src/ltm/types.rs:254`) implements the LTM papers' Rux/Bux/U runtime classification with the `POLARITY_CONFIDENCE_THRESHOLD` (0.99) confidence gate. But its only **production** caller is the discovery `FoundLoop` path (`src/simlin-engine/src/ltm_finding.rs:2036`). The other in-tree caller (`src/simlin-engine/src/db/analysis.rs:3074`) is inside a `#[cfg(test)]` test, not a production surface.

In **exhaustive mode**, `model_detected_loops` (`src/simlin-engine/src/db/analysis.rs`) reports the *structural* label with `polarity_confidence` hardcoded to `1.0` (every link signed) or `0.0` (any link unknown), and never applies a runtime override. The code comments confirm this is intentional today:

- `db/analysis.rs:808-814` -- "For the structural pipeline (`model_detected_loops`), this is 1.0 when every link ... 0.0 when any link is unknown."
- `db/analysis.rs:829-831` -- "The structural `model_detected_loops` pipeline never produces these variants -- it has no runtime data."

So `MostlyReinforcing`/`MostlyBalancing` (Rux/Bux) are structurally unreachable through the exhaustive-mode detected-loops surface, even though the runtime `loop_score` series exists post-simulation and `from_runtime_scores` would yield the honest Rux/Bux/U label with a real confidence value.

## Consequence

A loop whose polarity genuinely changes sign during the simulation (e.g. a logistic-style single-flow formulation, or the yeast-alcohol model from the LTM papers) is reported through the exhaustive path either:

- with a confidently-wrong steady `R`/`B` label (under the older, looser polarity rules), or
- with a blanket `Undetermined` (after the tightened Mul/Div polarity rules in `cc072973`),

when the truthful answer post-simulation is Rux/Bux/U with a concrete confidence. The `loop_score` series that would resolve this is available; it just is never consulted on the exhaustive path. This is a *labelling* fidelity gap -- loop **scores** themselves are unaffected.

## Why it matters

Correctness/honesty of the polarity label that downstream consumers (UI rendering, pysimlin, MCP analyze) present to modelers. A "mostly reinforcing, confidence 0.97" loop and a "reinforcing, confidence 1.0" loop are materially different stories about the model's behavior, and the engine has the data to tell them apart but doesn't on the exhaustive surface.

## Components affected

- `src/simlin-engine/src/ltm/types.rs` (`LoopPolarity::from_runtime_scores`)
- `src/simlin-engine/src/db/analysis.rs` (`model_detected_loops`, `DetectedLoop::polarity_confidence`, `DetectedLoopPolarity`)
- `src/libsimlin/src/analysis.rs` (the post-sim analyze path that has the `Results` in hand and already coalesces the variants -- see #495)

## Suggested approach

`model_detected_loops` itself is a *pre-simulation* salsa query (no runtime series), so the override does not belong there. After simulation, reclassify exhaustive-mode loops from their per-loop `loop_score` series via `from_runtime_scores`. The libsimlin analyze path already holds the simulation `Results` and the loop ids, so the reclassification (structural label + `polarity_confidence` -> runtime label + real confidence) belongs in that post-sim surface, mirroring what the discovery `FoundLoop` path already does in-engine. This also unblocks the FFI/TS plumbing tracked in #495 from ever seeing a non-trivial confidence on the exhaustive path.

## Related

- #495 -- surfacing `MostlyReinforcing`/`MostlyBalancing` + `polarity_confidence` through the JSON/TS/WASM/FFI layers (assumes the engine *produces* the variants; this issue is that exhaustive mode never produces them at all -- they are complementary)
- #506 -- the unreachable `MostlyBalancing` tie-break *inside* `from_runtime_scores` (code clarity; distinct from this caller-coverage gap)
- #516 -- agg-hop static polarity always `Undetermined` (a different code site)
- Epic: #488

## Context

Found during a design review on 2026-06-02. Severity: moderate (labels only; loop scores unaffected).

### LINKED ISSUE #737: ltm: loop-score builder's PureScalar fast path never routes a loop through a scalar synthetic agg node (silent zero loop score)
## Summary

`build_loops_from_tiered` (`src/simlin-engine/src/db/ltm/loops.rs`) never routes a pure-scalar feedback loop through a scalar synthetic aggregate node, so a loop whose path passes through a scalar feeder of a hoisted reducer (with a scalar target) gets a loop score composed from a direct variable-level link that skips the agg hop -- and that direct link has no link-score variable, so the loop score silently reads `0`.

Concretely, for

```
total = SUM(pop[*] * scale)     // pop arrayed, scale/total scalar
scale = f(grow)
grow  = f(total)                // closes the loop: total -> grow -> scale -> total
```

the reducer `SUM(pop[*] * scale)` is hoisted into a scalar synthetic agg `$⁚ltm⁚agg⁚0`, and (after the #533 element-graph fix on branch `ltm-bug-batch`, commit `189e7d05`) the element graph correctly contains `scale → $⁚ltm⁚agg⁚0 → total`. But the loop-score builder never consumes that routing for this cycle:

- The variable-level circuit `total → scale → grow → total` classifies as **PureScalar**, so `build_loops_from_tiered` takes the fast path: it materializes the `Loop` straight from the variable-level circuit via `var_graph.circuit_to_links(&var_level_nodes)`, producing a direct `scale → total` link.
- The cross-element-through-aggregate recovery (`recover_cross_agg_loops`) -- the only machinery that knows how to splice `from → $⁚ltm⁚agg⁚{n} → to` hops into a loop's link chain -- runs **only on the slow / cross-element path**, which a PureScalar cycle never reaches.

## Consequence

The feedback loop *is* enumerated (no crash, model compiles and simulates), but its loop score is composed from the `scale → total` link, for which no link-score variable exists / whose chain skips the agg hop -- so the loop score is silently `0`. Silent zero loop score for any scalar loop routed through a reducer aggregate with a scalar target.

This is **pre-existing and independent of #533's element-graph fix**: verified identical behavior with and without that fix, since the element-graph edge the fix adds is never consumed by the PureScalar loop-builder path.

## Severity

Medium -- silent wrong (zero) loop score for a valid and not-exotic pattern (a scalar control variable feeding a reduced aggregate, e.g. share/total structures with a scalar gain in the loop).

## Components affected

- `src/simlin-engine/src/db/ltm/loops.rs` -- `build_loops_from_tiered` (PureScalar fast-path materialization via `circuit_to_links`), `recover_cross_agg_loops` (the agg-hop splicing that the fast path never invokes)

## Characterization test

`src/simlin-engine/tests/integration/ltm_array_agg.rs::scalar_feeder_scalar_target_loop_compiles_and_is_well_formed` pins the current behavior end-to-end (compiles, simulates, loop enumerated, score 0) and documents this gap as item (2) in its doc comment.

## Related / sibling gap

The same test documents a sibling gap, item (1): the scalar-target synthetic agg `$⁚ltm⁚agg⁚0` for `SUM(pop[*] * scale)` itself currently **fails to compile** and is stubbed to constant `0` with an `Assembly` warning (an agg-augmentation gap, distinct from this loop-builder gap). That is being filed separately; even once it is fixed, this loop-builder gap independently keeps the loop score at 0. Cross-link the two issues once both exist.

## Possible approaches

- When a PureScalar (or any fast-path) circuit contains a variable-level link whose reference-site IR routes `ThroughAgg`, don't materialize the direct link from `circuit_to_links`; instead splice the `from → $⁚ltm⁚agg⁚{n} → to` hop into the link chain (the same substitution `recover_cross_agg_loops` performs on the slow path), so the loop score composes `score(scale→agg) * score(agg→total)`.
- Alternatively, classify a variable-level cycle that traverses a `ThroughAgg` site as *not* pure-scalar for loop-building purposes, sending it down the slow path where the existing agg recovery already works.
- Longer-term this is another instance of the #520 theme (all consumers should read agg routing from the unified classification IR rather than re-deriving structure from the raw variable graph).

## Discovery context

Identified during work on #533 (branch `ltm-bug-batch`, commit `189e7d05` -- the element-graph both-scalar fast-path fix), when constructing the scalar-target end-to-end test: with the element edge fixed, the loop score was still 0, and bisection showed the loop-score builder never consults the agg routing on the PureScalar path.

## Tracking

Part of LTM tracking epic #488. Related: #533 (element-graph counterpart of the same scalar-feeder/scalar-target scenario), #516 (closed; agg-hop polarity in `recover_cross_agg_loops`), #520 (closed; unified reference-site IR that a principled fix would lean on), #273 (LTM array-support umbrella).

### LINKED ISSUE #738: ltm: synthetic agg node for SUM(arrayed[*] * scalar) with a scalar target fails to compile and is stubbed to constant 0
## Summary

The synthetic aggregate node that LTM hoists for an inlined reducer of the shape `SUM(arrayed[*] * scalar)` **with a scalar target** fails fragment compilation, so it is silently stubbed to a constant `0` (with an `Assembly` `Warning` from `model_ltm_fragment_diagnostics`). Every link and loop score that routes through that agg is then wrong.

Repro shape (from the characterization fixture):

- `pop[region]` -- an arrayed stock
- `scale = 0.001 * total + 0.01` -- a scalar aux feeding back from the scalar stock `total`
- `grow = 1 + SUM(pop[*] * scale)` -- a scalar flow; `SUM(pop[*] * scale)` is a sub-expression (the `1 +` keeps it from being a whole-RHS, variable-backed agg), so it is hoisted into a synthetic `$⁚ltm⁚agg⁚0`

With LTM enabled the model compiles and simulates, but the hoisted `$⁚ltm⁚agg⁚0` node's **own equation** is rejected by the fragment compiler: `compile_ltm_equation_fragment` returns a result with `flow_bytecodes: None`, `assemble_module` drops the fragment, and the agg keeps its layout slot with no bytecode writing it -- it reads a constant `0` at runtime instead of `SUM(pop[*] * scale)`. `model_ltm_fragment_diagnostics` (`src/simlin-engine/src/db/ltm/compile.rs`, ~line 1147) emits the "failed to compile; ... evaluates to a constant 0" `Warning`, so the degradation is *visible* in diagnostics, but the scores are still silently wrong numerically.

The **arrayed-target twin** of the same expression compiles fine -- this is specific to the scalar-target shape (a scalar variable whose equation embeds a reducer over an arrayed dep multiplied by a scalar feeder).

## Why it matters

Medium. This is a quiet LTM correctness gap for scalar-target reducer loops -- a common, perfectly legal model shape (`total_flow = base + SUM(arr[*] * coeff)`). The agg's runtime value is `0`, so:

- the source→agg link scores (`pop[d] → $⁚ltm⁚agg⁚0`) and the agg→target link score (`$⁚ltm⁚agg⁚0 → grow`) are degraded, and
- every loop score whose chain passes through the agg is wrong.

A `Warning` is emitted (and per GH #466 it may not even reach the FFI diagnostics surface by default), but nothing in the numeric output flags the corruption.

## Components affected

- `src/simlin-engine/src/ltm_agg.rs` -- `enumerate_agg_nodes` (~line 308) and the agg equation construction: the emitted agg equation for the scalar-target `SUM(arrayed[*] * scalar)` shape is what the compiler rejects
- `src/simlin-engine/src/db/ltm/compile.rs` -- `compile_ltm_synthetic_fragment` / `compile_ltm_equation_fragment` (the fragment-compile path that fails) and `model_ltm_fragment_diagnostics` (~line 1147, the Warning that makes the stub visible)

## Possible approaches

- Root-cause why the scalar-target agg equation fails fragment lowering while the arrayed-target twin succeeds (likely a dimension/typing mismatch in how the emitted agg equation or its dependency stubs are shaped when the hoisting target is scalar), and fix the agg equation construction in `ltm_agg.rs` (or the fragment dep-stub shaping in `db/ltm/compile.rs`) so the fragment compiles.
- Once the agg computes, note that the pure-scalar loop `total → scale → grow → total` still won't route through it: `build_loops_from_tiered` materializes a `PureScalar` fast-path cycle straight from the variable-level circuit, linking `scale → grow` directly (gap #2 in the characterization test's doc comment) -- closing that routing gap is adjacent follow-on work.

## Existing test coverage

A characterization test documenting both gaps exists: `scalar_feeder_scalar_target_loop_compiles_and_is_well_formed` in `src/simlin-engine/tests/integration/ltm_array_agg.rs` (see its doc comment, which enumerates this issue as gap #1). It pins compile/simulate well-formedness (finite scores, loop enumerated) without asserting the currently-zero values, so closing this gap will not break it -- but a fix should tighten it to assert a non-zero agg value/score.

## Discovery context

Identified during work on GH #533 (branch `ltm-bug-batch`, commit 189e7d05). Both gaps behave identically with and without #533's element-graph fast-path fix -- the element edge that fix adds is never consumed downstream here -- so this is pre-existing, independent breakage.

## Tracking

Part of LTM tracking epic #488. Related to but distinct from: #533 (element-graph both-scalar fast path bypassing ThroughAgg routing -- the element edge, not the agg node's value), #525 (link-score *partial* equation fails to compile for partially-iterated subscripts -- a different synthetic-equation family), #514/#534 (whether a reducer gets hoisted at all -- here the hoist succeeds, the hoisted node's equation doesn't compile), #546/#548 (closed: other causes of synthetic fragments stubbing to 0).

### LINKED ISSUE #739: ltm: element-graph projection proptest never generates inlined reducers, so ThroughAgg routing is verified only vacuously
## Problem

The element-graph projection property test (`src/simlin-engine/src/db/element_graph_proptest.rs`) has a structural blind spot: `project_spec_strategy()` (line ~125) never generates a model containing an inlined (sub-expression) array reducer, e.g.

```
total = base + SUM(pop[*] * scale)
```

Because no generated model contains such a pattern, the aggregate-node hoisting machinery never fires under the proptest -- no synthetic `$⁚ltm⁚agg⁚{n}` node is ever minted. As a consequence:

- the arrayed `ThroughAgg` routing in `model_element_causal_edges` (`db/analysis.rs`), and
- the new both-scalar `ThroughAgg` routing added by the #533 fix (commit 189e7d05 on branch `ltm-bug-batch`)

are completely unexercised by the projection invariant `element_edges_project_to_variable_edges` (line ~238). The property "element edges project onto variable edges" holds only **vacuously** for the entire agg-routed edge class.

## Why it matters

The projection invariant is the main structural guard on the element graph. The #533 fast-path bug (a scalar feeder of a hoisted reducer with a scalar target losing the agg hop) is exactly the kind of routing mistake this proptest exists to catch, yet the generator could never have found it -- and cannot detect a regression of the fix. Test coverage gap only; no production bug.

## Components affected

- `src/simlin-engine/src/db/element_graph_proptest.rs` (`project_spec_strategy`, `project_to_variable_edges`, `element_edges_project_to_variable_edges`)

## Fix direction

1. Extend `project_spec_strategy` with an inlined-reducer equation pattern (a reducer like `SUM(arr[*])` or `SUM(arr[*] * scalar)` appearing as a sub-expression, not the whole RHS), covering both arrayed and scalar feeders/targets so both `ThroughAgg` arms are reachable.
2. Teach the test's `project_to_variable_edges` to collapse synthetic `$⁚ltm⁚agg⁚{n}` nodes the same way real consumers trim them (splice `from -> agg -> to` into `from -> to` before comparing against the variable-level graph), so the invariant is stated over the trimmed graph.

## Context

Surfaced during adversarial review of the #533 fix (branch `ltm-bug-batch`, commit 189e7d05). Related: #533, #520, #737, #738. Part of epic #488.

bpowers:
This batch works through the nine bug/correctness items in epic #488's Core algorithm group (the two capability features there, #658 and #674, are deferred to a follow-up branch). Each issue was implemented by a fresh agent under TDD and then adversarially reviewed by a separate fresh agent, iterating until zero material findings; several fixes took 2-3 review rounds, and the reviews repeatedly caught real defects before merge (a compile-time regression, an unsound lowering gate, two loop-id-collision mechanisms, and a polarity-soundness hole among them).

## What changed, by cluster

**Scalar-target agg routing (#738, #737, #739).** LTM synthetic fragments now lower with a minimal model scope so arrayed deps carry dimension bounds -- the hoisted `SUM(pop[*] * scale)` agg for a scalar target compiles instead of silently stubbing to 0. The lowering re-pass is gated by a structural walk of Pass-1's actual decomposition node set (an exhaustive match, so a new builtin fails compilation until classified); cold C-LEARN LTM compile cost is +5% (was +26% before mitigation; quantified on #655). Cycles traversing a ThroughAgg edge are classified off the PureScalar fast path so their loop scores compose the agg-half link scores -- the scalar feeder half uses a changed-last attribution (exactly complementary to the per-row changed-first shortcut, so additivity is exact for bilinear bodies; documented as a deviation in the reference doc). The detected surface now splices one loop variant per routed agg so detected/scored loop ids biject for scalar-cycle models (the id join previously read the wrong loop's series); a discriminating body-polarity analysis replaced the blanket monotone source-to-agg label, which was confidently wrong for negating and co-source bodies. The element-graph projection proptest now force-generates inlined reducers and asserts agg-hop routing (the projection invariant alone is provably blind to this bug class), mutation-tested in both directions.

**Cross-agg enumeration (#676).** `cyclic_orderings`' (m-1)!/2 mirror-skip rationale was mathematically wrong (the skipped orderings were not edge-reverses); since every cyclic ordering of a fixed petal subset yields the same edge multiset and hence the same score, recovery now emits one canonical loop per disjoint petal subset, freeing the 256-loop budget for genuinely distinct structure. Counts on >=4-petal models drop (13 -> 11 in the four-petal fixture); both exhaustive and discovery share the change via the common stitcher.

**Pins (#673).** Pin stock-validation now uses the enumeration surface's exact semantics (`find_stocks_in_loop` + `enrich_with_module_stocks`), so a loop whose only stock lives inside a SMOOTH/DELAY or user sub-model validates instead of being rejected "contains no stock"; stockless passthrough cycles are still rejected. First test coverage for pins through modules.

**Arrayed agg shapes (#528).** The broadcast agg-to-target link-score equation body now pins the agg ident to the target element's projection onto `result_dims` (mirroring the already-correct name and denominator projections), so `agg[D1]` feeding `to[D1,D2]` compiles and scores instead of stubbing every loop through it to 0. The diagonal case is byte-identical.

**Mapped dimensions (#527, #534).** Same-element edges across a mapped dimension pair project along the mapping diagonal instead of the full cross-product, and mapped-dimension sliced reducers hoist into aggregate nodes (the Iterated axis now carries a (target dim, source dim) pair remapped via one shared correspondence helper). Both are deliberately restricted to POSITIONAL mappings: review demonstrated the engine's executed A2A lowering ignores explicit element maps and resolves positionally (filed as #756), so element-map cases stay conservatively broadcast/un-hoisted until that is resolved -- a graph-vs-simulation parity test pins the boundary and survives the future engine fix. Mapped whole-RHS reducers mint a synthetic agg (documented exception) because the variable-backed Wildcard partial mangles the iterated index (#759); this turns 8 spurious zero-scored loops into 4 exactly-scored ones.

**Runtime polarity (#679).** The issue's ask (post-sim Rux/Bux reclassification on the exhaustive surface via `from_runtime_scores`) turned out to be satisfied by `reclassify_loops_from_results` + `simlin_analyze_get_loops_runtime` in the branch's ancestry; what was missing was any proof a definite MostlyReinforcing was reachable. This lands mixed-sign fixtures pinning Rux at confidence 0.9999 (and a sub-gate Undetermined control) end-to-end through the engine and the libsimlin C-ABI, with discovery/exhaustive parity bit-identical.

## Discovered along the way

The adversarial reviews surfaced twenty previously-untracked pre-existing defects, all filed and added to epic #488: #741-#752 and #753-#760 (notably #756, silent wrong simulation results for explicit element maps, and #746, the arrayed-cycle loop-id namespace overlap).

Fixes #738
Fixes #737
Fixes #739
Fixes #676
Fixes #673
Fixes #528
Fixes #527
Fixes #534
Fixes #679


bpowers:
@codex review

Issue #785, opened 2026-06-11
ltm: square-source aggs (repeated dim in result_dims) pin by dim name across co-source and agg-to-target halves; only the feeder half is defended


bpowers:
## Summary

A degenerate square-source reducer -- `x[D1] = 1 + SUM(cube[D1,D1,*] * frac[D1,D1])` with `cube[D1,D1,D2]` and `frac[D1,D1]` -- mints a synthetic agg whose `result_dims == ["D1","D1"]`: both axes classify `Iterated` with the same target dim. The LTM machinery enumerates agg slots over the full square while the executed A2A simulation reads only the diagonal, and several emission paths resolve subscript indices by FIRST-MATCH dim name -- which is ambiguous when a dim repeats in `result_dims`.

Commit `1461b6dc` (branch `ltm-shape-phase1`, follow-up to merged PR #784) defended the FEEDER half: `pin_iterated_dim_indices` now bails on duplicate dim names, converting the score to the #743-style `UnfreezablePartial` loud skip (fixture: `square_source_duplicate_dim_feeder_scores_are_loudly_skipped` in `src/simlin-engine/src/db/ltm_unified_tests.rs`).

The other two halves still pin by dim name and likely share the first-match hazard on this shape:

- **Co-source half** (`cube -> agg` row partials via `pin_body_to_row`): may emit confident values pinned to the wrong slot. An adversarial review of `1461b6dc` is concurrently assessing whether this half emits silent wrong values post-fix -- its conclusion may refine this issue's severity.
- **Agg-to-target half** (the #528 projection: `positions_in_to_dims` / `qualified_pin_for_target` in `src/simlin-engine/src/db/ltm/link_scores.rs`): observed emitting only the `agg[r1,r1] -> x[r1]` diagonal form.

## Why it matters

If the co-source half resolves a repeated dim by first match and emits a numerically wrong but confident link score, that violates the no-silent-wrong-numbers invariant (the loud-skip discipline from #743/#762). Even if all halves turn out merely inconsistent rather than silently wrong, the asymmetry (one half loudly skipped, the others not) makes the surface incoherent on this shape.

## Components affected

- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `pin_body_to_row` (co-source row partials), `positions_in_to_dims` / `qualified_pin_for_target` (agg-to-target projection), `pin_iterated_dim_indices` (feeder; already defended)
- `enumerate_agg_nodes` -- mints the square `result_dims` agg in the first place

[...]

Issue #780, opened 2026-06-11
ltm: PartialEquationError link-score skips don't drop dependent loop scores via unscoreable_edges


bpowers:
## Summary

When `generate_link_score_equation_for_link` returns a `PartialEquationError`, the shaped compilation path emits the loud `emit_ltm_partial_equation_warning` and skips the link-score variable -- but the edge is **not** recorded in `unscoreable_edges`. Loop scores traversing the edge therefore reference the never-emitted link-score name, fail fragment compile, and degrade to warned constant-0 stubs.

Observed repro: 1 good edge warning + 21 cascading loop-score warnings, with every loop series constant 0.

## Mechanism

- `link_score_equation_text_shaped`'s generator call (`src/simlin-engine/src/db/ltm/compile.rs:222-225` at commit `fa8fa774`, branch `ltm-shape-phase1`) maps `PartialEquationError` to `emit_ltm_partial_equation_warning` + `return None` -- the GH #311 loud skip.
- The per-shape emission loop in `emit_per_shape_link_scores` (`src/simlin-engine/src/db/ltm/link_scores.rs:1346-1385`) receives `unscoreable_edges: &mut HashSet<(String, String)>` (it already inserts for the GH #758 dim-incompatible class and the broadcast-reduce class at line ~1340/~493), but when the shaped query returns `None` it simply doesn't push the variable -- the edge never enters the set.
- The #758 drop machinery (`traverses_unscoreable` in `src/simlin-engine/src/db/ltm/mod.rs` ~1602, consumed for detected loops ~1622-1628 and pinned loops ~1888) therefore never sees these edges; loop scores through them are emitted, fail fragment compile against the missing link-score name, and `assemble_module` stubs them to constant 0 with per-loop Assembly `Warning`s.

This conflicts with the standing invariant established by #758: ONE loud warning naming the edge, no link-score variable, and dependent loop scores **dropped** -- not warned-garbage zeros.

## Reachability

`PartialEquationError` is reachable via:
- GH #311 parse failures (defense-in-depth; effectively unreachable in production today), and
- since the T7 #526 commit `fa8fa774` (branch `ltm-shape-phase1`), the new both-legs-doomed mismatched-dep class -- which makes the gap practically reachable.

[...]

PR #786, opened 2026-06-11
engine: defend ltm feeder slot pin and agg source slices

bpowers:
Follow-up to #784, fixing the two P3 findings its automated review surfaced after merge. Both harden the phase's no-silent-fallback discipline, and the first turned out to be a real reachable defect rather than defense-in-depth.

**Feeder slot pinning on square-source aggs**: a degenerate square-source reducer (`x[D1] = 1 + SUM(cube[D1,D1,*] * frac[D1,D1])`) mints an agg with `result_dims == [D1, D1]`, and `pin_iterated_dim_indices` resolved indices by first-match dim name -- every duplicated-dim index pinned to the first slot part, silently freezing the wrong source row on off-diagonal feeder scores (confident garbage, zero diagnostics). It now mirrors `resolve_mismatched_index_position`'s uniqueness defense: an ambiguous index bails to the existing `UnfreezablePartial` loud-skip channel (per-row warning, no score emitted). RED-verified pre-fix at both fixture and unit level; byte-identical for every non-repeated-dim shape by construction.

**`agg_sources` release-build fallback**: the helper carried a `debug_assert` that every arrayed reducer source has a per-var slice, then silently substituted the canonical slice in release if the invariant ever broke -- which would mislabel a projection feeder as a co-source and corrupt per-(row, slot) derivations. It now returns `Option` and declines the hoist entirely on a missing slice (the same inert conservative degradation as the dynamic-index carve-outs). The reviewer-suggested empty-slice fallback was rejected because an arrayed source wearing the scalar-feeder encoding would flow into the scalar-agg name grammar -- plausibly wrong rather than inert. The two AST walkers were audited as structurally symmetric, so the path is unreachable today; the defense now matches the asserted invariant.

The remaining square-source hazards (co-source and agg-to-target halves still pin by dim name; phantom off-diagonal link scores reach link-level surfaces unwarned) are pre-existing and tracked as #785/#778, with the loop-score cascade gap as #780. Adversarial review of this commit: APPROVE, with the empirical co-source findings appended to #785.


bpowers:
Superseded by a clean PR: #784 was squash-merged, so this branch's pre-merge history conflicts with main. The single follow-up commit is rebased onto main on branch ltm-p3-hardening; new PR incoming.

PR #795, opened 2026-06-12
engine: close ltm phase 1 residuals

### LINKED ISSUE #754: ltm-finding: discovery mode mints phantom from-side element keys for Bare A2A edges (mapped-dimension and scalar-source legs; loops undiscoverable)
## Summary

Discovery mode cannot score (or even discover) mapped-dimension loops: `expand_a2a_link_offsets` in `src/simlin-engine/src/ltm_finding.rs` (~line 676) -- the per-element expansion of a Bare A2A link score into discovery-graph edges -- subscripts **both** sides of the edge with the same element tuple drawn from the link score's dimensions:

```rust
for (idx, elems) in tuples.iter().enumerate() {
    let subscript = subscript_from_elements(elems);
    let from = Ident::new(&format!("{from_var}[{subscript}]"));
    let to = Ident::new(&format!("{to_var}[{subscript}]"));
    link_offsets.push(((from, to), base_offset + idx));
}
```

This assumes the from-side variable's elements equal the link score's dimensions' elements. The link score's dimensions are the **target's** dimensions, so for a *mapped* edge -- source `x` over `Region{r1,r2}`, target over `State{s1,s2}` with a `State→Region` mapping, link score ApplyToAll over the target's dim (the shape the #527 fix on `ltm-core-batch`, commit `a010931e`, now emits for the element graph's mapped diagonal) -- it mints from-side node keys `x[s1]`/`x[s2]`: elements of the **wrong dimension**. Those keys match no real element node (`x[r1]`/`x[r2]`), so the discovery graph's mapped edges dangle and any loop through a mapped-dimension edge is undiscoverable and unscorable in discovery mode.

Exhaustive mode works correctly post-#527; the asymmetry is discovery-only.

## Why it matters

Mapped-dimension loops are **silently missing** from discovery-mode analysis -- no warning, no diagnostic, the loops simply never appear. Discovery is the mode large models auto-flip into (`MAX_LTM_SCC_NODES` gate), so the models most likely to use dimension mappings are exactly the ones affected. Severity: low-medium (silent completeness gap, not wrong numbers).

## Component

- `src/simlin-engine/src/ltm_finding.rs` -- `expand_a2a_link_offsets` (~line 676) and its caller `parse_link_offsets` (the Bare A2A arm, ~line 610)
- `src/simlin-engine/src/dimensions.rs` -- `DimensionsContext::mapped_element_correspondence` (line 670; the fix's data source)

## Fix direction

Translate the from-side element through the same `DimensionsContext::mapped_element_correspondence` helper the element graph now uses (added in `src/simlin-engine/src/dimensions.rs` by commit `a010931e` for the #527 `expand_same_element` diagonal): when the link score's dimension differs from the from-side variable's declared dimension but a mapping exists, emit `x[<mapped source elem>] → to[<target elem>]` instead of subscripting both sides with the target's element. This requires threading the from-side variable's declared dimensions (available via the `ltm_var_map` / model metadata) and a `DimensionsContext` into the expansion, keeping the discovery graph's node keys in lockstep with the element nodes `model_element_causal_edges` produces.

## Discovery context

Pre-existing -- discovery's per-element expansion predates the branch. Surfaced during GH #527 work on branch `ltm-core-batch` (commit `a010931e`): the exhaustive-mode mapped scoring fix made the discovery asymmetry visible. Related: #527 (element-graph mapped diagonal, the exhaustive-side fix), #273 (LTM arrays umbrella), epic #488.

### LINKED ISSUE #777: ltm: arrayed-owner Pinned/subset broadcast reduce slice loud-skipped; needs the PerElement broadcast rule on variable-backed owners
## Summary

T3 of the LTM shape-expressiveness phase (design: `docs/design-plans/2026-06-11-ltm-shape-expressiveness.md`, Architecture section 6; commit `0a1f27fe` on branch `ltm-shape-phase1`) generalized the #752 gate to `variable_backed_reduce_agg`, admitting scalar-result variable-backed slices (`total = SUM(pop[nyc,*])`) -- but scoped that admission to SCALAR owners (`to_dims.is_empty()`). The ARRAYED-owner sibling is DECLINED into the #758-style loud skip and remains a design-mandated residual:

```
share[Region] = SUM(pop[nyc,*])
```

i.e. a whole-RHS variable-backed reducer whose slice has NO `Iterated` axis (`Pinned` or subset-`Reduced` only) and whose OWNING variable is arrayed over dims unrelated to the slice. The same applies to the subset twin:

```
share[Region] = SUM(arr[*:Sub])
```

## Current behavior (post-T3, the loud skip)

Both the gate and `try_cross_dimensional_link_scores`' partial-reduce branch decline the shape; the edge routes to the #758 loud skip: one `Warning` via `emit_unscoreable_broadcast_reduce_edge_warning` (`src/simlin-engine/src/db/ltm/link_scores.rs`), no link-score variables, and every loop through the edge is dropped from LTM analysis.

This replaced the PRE-T3 state, which was silently wrong (recorded during T3's RED run): the unread rows (`pop[boston,*] -> share[boston]`) scored constant +1.0 although `share[boston] = SUM(pop[nyc,*])` never reads them, while the TRUE dependency (`pop[nyc,*] -> share[boston]`) had NO score at all. Pinned by `arrayed_owner_broadcast_pinned_slice_reduce_skips_loudly` in `src/simlin-engine/tests/integration/ltm_array_agg.rs` (loud skip, no garbage, no phantom names).

## Proper fix

The design's section 3 `PerElement` broadcast rule -- per-`(row, full-target-element)` emission -- applied to variable-backed owners: every target slot reads the same Pinned/subset slice, so emit the read rows fanned out across the target's full element set, with matching per-`(row, slot)` link scores. This is T6-style machinery that the design deliberately kept out of T3 ("specifiable later, but it is T6 machinery and does not belong in T3"). The fix should cover BOTH the Pinned shape (`SUM(pop[nyc,*])`) and the subset twin (`SUM(arr[*:Sub])`).

## Why it matters

Severity: low. Legal but uncommon shape; the degradation is loud and conservative (no wrong numbers -- strictly better than the silent pre-T3 garbage), but each occurrence drops every loop through the edge.

## Components affected

- `src/simlin-engine/src/ltm_agg.rs` -- `variable_backed_reduce_agg` gate's `to_dims.is_empty()` scoping
- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `try_cross_dimensional_link_scores` partial-reduce branch decline; `emit_unscoreable_broadcast_reduce_edge_warning`

## Related

- #764 -- the ITERATED-armed variable-backed broadcast/permuted sibling (this issue is the NO-Iterated-axis sibling; distinct shape class)
- #765 -- Pinned axes in the co-reduced slice (value-level sibling; the pre-T3 full-cartesian garbage for this shape was an unfiled #765 sibling)
- #769 -- FixedIndex into disjoint-dim ApplyToAll target (plain reference, no reducer; same loud-skip gate, same recoverable-in-principle character)
- #758 (closed) -- the loud-skip gate this shape now routes through
- #488 -- LTM tracking epic (cluster A)

## Discovery context

Design-mandated residual recorded during T3 of the LTM shape-expressiveness phase on branch `ltm-shape-phase1` (commit `0a1f27fe`); the design doc explicitly defers it and requires filing if it survives the phase.

### LINKED ISSUE #778: ltm: duplicated-result-dim aggregate fans out phantom off-diagonal agg->target element edges (warned 0-stub loops)
## Summary

A reducer whose Iterated axes **repeat a dimension** -- `out[D1] = SUM(sq[D1,D1,*])` over a square source `sq[D1,D1,D3]`, or the inline form `out[D1] = base[D1] + SUM(sq[D1,D1,*])` -- gets a synthetic agg node arrayed over `result_dims = [D1,D1]`. The agg->target element-edge emitter and the link-score projection disagree about which of those slots are real:

- `emit_agg_routed_edges` (`src/simlin-engine/src/db/analysis.rs`) fans out ALL `[D1,D1]` slots, including off-diagonal ones: `agg[a,b] -> out[a]`. These edges are not causally real -- the target's slot `out[a]` only reads the diagonal slot `agg[a,a]`.
- The score projection (`result_dim_positions` in `src/simlin-engine/src/db/ltm/link_scores.rs`) maps BOTH `D1` occurrences to position 0, so per-slot link scores are emitted for the diagonal slots only.

Circuits through the off-diagonal edges therefore reference never-emitted link-score names, fail fragment compile, and produce **loud, warned 0-stub loop scores**. Measured at `b7c559c1` (branch `ltm-shape-phase1`): the whole-RHS form yields 4 correctly-scored diagonal loops + 8 warned phantom 0-stub loops. The INLINE form behaves identically and **predates the shape-expressiveness phase -- it is reachable on `main` today**.

## Provenance

Pre-existing inline-path defect. The shape-expressiveness phase (T4) widened its reachability to whole-RHS spellings -- a strict improvement over the parent whole-RHS state at T4's base (16 warned / 24 loops with zero correct scores), but the duplicated-axis disagreement itself is unchanged.

## Why it matters

Severity: **low**. Square-matrix self-indexed reducers are rare; the degradation is loud (fragment-diagnostic warnings + 0-stubs), so no silent wrong numbers. But the phantom off-diagonal loops pollute the enumerated-loop surface and the warned count.

## Components affected

- `src/simlin-engine/src/db/analysis.rs` -- `emit_agg_routed_edges` fans out the full `[D1,D1]` cross-product of agg slots
- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `result_dim_positions` maps both duplicated-dim occurrences to one target position, pinning scores to the diagonal

## Fix direction

A duplicated Iterated dim is semantically ONE axis read on the diagonal. Either:

1. **Canonicalize in the agg data model**: dedupe/diagonalize duplicated result axes -- the read slice collapses to a single Iterated axis plus a diagonal constraint, so the agg is arrayed over `[D1]` and both surfaces agree by construction (preferred; matches the epic's "two-surface decisions share one predicate" invariant), or
2. **Patch the emitter**: have `emit_agg_routed_edges` fan out only the diagonal slots when `result_dims` repeats a dimension, matching the existing `result_dim_positions` projection.

## Related

- #488 -- LTM tracking epic (cluster A: read-slice / shape expressiveness)
- #764 -- broadcast/permuted-axis sibling shapes (distinct mechanism: duplicated axes, not reordered/extra ones); closing via the shape-expressiveness phase PR
- `docs/design-plans/2026-06-11-ltm-shape-expressiveness.md` -- the phase design doc

## Discovery context

Characterized by adversarial review during T4 of the LTM shape-expressiveness phase on branch `ltm-shape-phase1`, commit `b7c559c1`.

### LINKED ISSUE #779: ltm: bare-spelled feeder of un-hoisted multi-source reducer gets silent wrong changed-last loop score (per-element partial vs broadcast execution)
## Summary

The BARE-spelled feeder of an un-hoisted multi-source reducer gets a **silently wrong (unwarned, sustained) loop score** from the GH #743 changed-last partial: the partial compiles per-element while the executed equation broadcasts the bare reference.

Fixture:

```
D1 = {a, b}, D2 = {c, d}
matrix[D1, D2]                            -- co-source
frac                                      -- feeder, referenced BARE (unsubscripted)
growth[D1] = SUM(matrix[D1, *] * frac)    -- execution iterates the bare frac over D1
... feedback loops closed through matrix and through frac
```

This is the unfinished **bare-spelling sibling of #743**, whose fix covered the SUBSCRIPTED feeder spelling `frac[D1]` (now fully hoisted by the T5 feeder-sub-slice acceptance, #767).

## Mechanism

1. The bare `frac` reference classifies **all-Reduced** -- the read-slice vocabulary cannot express a bare iterated read (a reference that execution iterates over the owner's A2A dimension without a subscript). Verified byte-identical pre/post T5 at commit `bcb89565` on branch `ltm-shape-phase1`: this is PRE-EXISTING, not a T5 regression.
2. Because the slices disagree, the reducer stays un-hoisted. The matrix-closure loops land on the LOUD degraded path (5 fragment warnings + 0-stubs) -- that half is fine.
3. The frac-closure loop falls to the #743 changed-last fallback partial, `sum(matrix[d1,*] * PREVIOUS(frac))`, which **compiles per-element** -- while the executed growth equation **broadcasts** the bare `frac` across all D1 rows (in the probe, `growth_0 = 10` = the sum of all 2x2 terms).
4. Result: a **sustained, UNWARNED loop score of ~2.92** for the frac-closure loop, internally inconsistent with the four co-source loop scores of +1 each.

## Why it matters

Severity: **low-medium**. This is a silent wrong number (~3x) for a natural spelling of a common shape (`per-row total = SUM(row * shared coefficient)` with the coefficient written bare), violating epic #488's standing "no silent wrong numbers" invariant. The subscripted spelling `frac[D1]` is now fully correct after T5/#767, so users hitting this have an easy workaround they don't know they need.

## Components affected

- `src/simlin-engine/src/ltm_agg.rs` -- read-slice classification (`compute_read_slice` / `combined_read_slice`): no vocabulary for a bare iterated read
- `src/simlin-engine/src/ltm_augment.rs` -- the changed-last chooser accepts the bare-feeder shape and emits a per-element partial that disagrees with the executed broadcast
- `src/simlin-engine/src/db/ltm/link_scores.rs` -- the emitted frac-closure loop score

## Possible approaches

- **Teach the slice vocabulary a bare iterated read** (likely the right fix): classify the bare feeder reference the way execution treats it -- iterated over the owner's A2A dim -- so T5's feeder-sub-slice acceptance covers it and the reducer hoists exactly like the `frac[D1]` spelling. Cf. how `classify_iterated_dim_shape` already treats bare refs as Bare/diagonal.
- **Or have the changed-last chooser decline the bare-feeder shape loudly** (warning + 0-stub), restoring the no-silent-wrong-numbers invariant without the hoist.

## Discovery context

Demonstrated by a reviewer during T5 review of the LTM shape-expressiveness phase (branch `ltm-shape-phase1`, commit `bcb89565`). Verified byte-identical pre/post T5 -- pre-existing, out of scope for that phase.

## Tracking

Part of LTM tracking epic #488 (cluster A) and the array-support umbrella #273. Related: #743 (the subscripted-spelling sibling, fixed -- its changed-last fallback is the partial that goes wrong here), #767 (the feeder-sub-slice hoist extension that fixed the subscripted spelling's co-source closure, closing via the shape-expressiveness phase PR).

### LINKED ISSUE #780: ltm: PartialEquationError link-score skips don't drop dependent loop scores via unscoreable_edges

### LINKED ISSUE #783: ltm: emit_agg_routed_edges enumerates rows via its own AxisPlan, deviating from the design's I4 single-derivation invariant
## Problem

The shape-expressiveness design (`docs/design-plans/2026-06-11-ltm-shape-expressiveness.md`, invariant I4 in Architecture section 2) states:

> the rows a source contributes per result slot are computed ONLY by `read_slice_rows` applied to that source's slice. No consumer re-derives rows from `from_dims` cartesian products.

`emit_agg_routed_edges` (`src/simlin-engine/src/db/analysis.rs`, ~lines 805-1010) violates this: it enumerates agg-routed element-edge rows via its own per-source `AxisPlan` machinery rather than consuming `read_slice_rows`.

## Evidence

During the adversarial review of the Phase-1 proptest extension (commit `d26dfc3e` on branch `ltm-shape-phase1`), a mutation experiment proved the independence: widening `Reduced.subset` to the full extent INSIDE `read_slice_rows` left the production agg-routed edges correct (deterministic subset-edge tests still passed) while only the proptest's expectation side changed. So the link-score emitters consume `read_slice_rows`, but the element-graph edge emitter re-derives rows independently -- two parallel enumerations that currently agree but can drift. That is exactly the two-surface-divergence class the design phase exists to eliminate (and a direct instance of epic #488's standing invariant "Two-surface decisions share one predicate").

Silver lining: because the two derivations are independent, the proptest oracle is accidentally differential for subset routing -- a drift WOULD be caught by the generative tests as long as they cover the drifting shape.

## Why it matters

No behavioral divergence today, but it is a standing-invariant violation in the design's own terms. Any future change to `read_slice_rows` (subranges, permutations, broadcasts -- the whole Phase-1 surface) must be mirrored by hand in `emit_agg_routed_edges`' `AxisPlan` path or the element graph and the link scores silently disagree about which rows an edge covers.

## Component

`src/simlin-engine/src/db/analysis.rs` (`emit_agg_routed_edges`), LTM element-graph construction.

## Suggested direction

Refactor `emit_agg_routed_edges`' row planning to consume `read_slice_rows` per source (or extract one shared row-plan helper that both the element-edge emitter and the link-score emitters consume), preserving byte-identical edge output -- golden-pin the existing element-graph tests before the refactor. Low urgency: no wrong numbers today, and the proptest guards subset/mixed shapes differentially.

## Provenance

Found by the adversarial reviewer of the Phase-1 proptest commit (`d26dfc3e`, branch `ltm-shape-phase1`) via mutation testing, 2026-06-11. Tracked under epic #488 cluster A.

### LINKED ISSUE #785: ltm: square-source aggs (repeated dim in result_dims) pin by dim name across co-source and agg-to-target halves; only the feeder half is defended

### LINKED ISSUE #788: ltm: bare arrayed reducer arg consumed per-slot by execution is hoisted as a whole-array scalar agg (silent wrong link and loop scores)
## Summary

A BARE arrayed argument to a whole-RHS reducer -- `SUM(other)` with `other[D1]` arrayed, inside an A2A equation over D1 -- is consumed PER-SLOT by execution but hoisted by LTM into a SCALAR synthetic aggregate whose value is the WHOLE-ARRAY sum. Every score touching the reducer is silently wrong (unwarned, sustained): phantom cross-element edges, a wrong changed-first freeze, and wildly inflated loop scores.

This is the bare-ARRAYED-arg sibling of #779 (bare SCALAR feeder of an un-hoisted reducer). Same root cause -- the read-slice vocabulary cannot express a bare iterated read -- but #779's loud-decline fix (commit `ef91194d`, branch `ltm-phase1-residuals`) covers only the changed-LAST live-feeder leg. The synthetic-agg halves and the changed-FIRST whole-reducer freeze still score silently wrong.

## Fixture (reproduced at `ef91194d` with a TestProject fixture)

```
D1 = {a, b}
pop[D1]    -- stock
other[D1]  = pop * 0.01
frac[D1]   = pop * 0.005
growth[D1] = SUM(other) * frac   -- flow into pop
```

Execution resolves the bare `other` in the A2A-over-D1 context to the ACTIVE slot: `SUM(other)` = `other[r]` (a sum over one scalar), so `growth[a]` at t0 = 1.0 * 0.5 = 0.5 -- NOT the whole-array sum 2.0 * 0.5.

## What LTM does instead (all unwarned)

1. `enumerate_agg_nodes` hoists `SUM(other)` into a scalar synthetic agg `$:ltm:agg:0 = SUM(other)` whose VALUE is the whole-array sum (2.0): the hoisted fragment evaluates outside the A2A context, so the bare ref reads the full extent.
2. The element graph models cross-element edges `other[a/b] -> $:ltm:agg:0 -> growth[a/b]` that do not exist in execution (execution has only the diagonal `other[r] -> growth[r]`).
3. The changed-first partial for the `frac -> growth` edge freezes `PREVIOUS(SUM(other))` -- again the whole-array sum -- into the partial.

Measured result (ZERO warnings):
- `$:ltm:link_score:frac->growth` = ~100.7 sustained
- `$:ltm:link_score:$:ltm:agg:0->growth[a]` = ~100.7
- loop u1 = ~2512; loops u2/u3 = ~50

This violates epic #488's standing "no silent wrong numbers" invariant.

## Components affected

- `src/simlin-engine/src/ltm_agg.rs` -- `enumerate_agg_nodes` hoists the bare-arg `SUM(other)` as an all-Reduced slice; no vocabulary for a bare iterated read (the per-slot consumption execution actually performs)
- `src/simlin-engine/src/ltm_augment.rs` -- `wrap_non_matching_in_previous` whole-reducer changed-first freeze captures the whole-array sum
- `src/simlin-engine/src/db/ltm/link_scores.rs` -- the emitted agg-routed and feeder link scores

## Possible approaches

- Teach the slice vocabulary a bare iterated read (the fix direction #779 also names): classify the bare arrayed arg the way execution treats it -- the active slot of the owner's A2A dimension -- so the "reducer" is recognized as a per-slot identity read (diagonal element edges, no whole-array agg, per-slot partials).
- Or, short-term, decline the bare-arrayed-arg reducer shape loudly (warning + dropped dependent loop scores per the #758/#780 contract), restoring the no-silent-wrong-numbers invariant without the classification work.

Note the interaction with engine semantics: execution's per-slot resolution of a bare arrayed name inside a reducer is the XMILE A2A reading (a bare arrayed name in an A2A equation denotes the active element). Whatever the engine-side resolution of the related execution issue (see the co-filed engine issue on the bare-ref-with-sliced-co-source x|D1| bug), LTM must model what execution actually computes.

## Discovery context

Empirically confirmed during the adversarial review of commit `ef91194d` ("engine: decline bare-spelled feeder of un-hoisted reducer loudly", the GH #779 fix, branch `ltm-phase1-residuals`), reproduced at that commit with TestProject fixtures. Pre-existing, not a regression from that commit; out of scope for #779's fix.

## Tracking

Part of LTM tracking epic #488 (cluster A). Related: #779 (bare scalar feeder sibling; its fix covers only the changed-last live-feeder leg), #743/#767 (the subscripted-spelling lineage), #738 (closed scalar-target agg compile sibling).

### LINKED ISSUE #789: engine: bare arrayed ref sharing a reducer with a sliced co-source multiplies the result by the bare ref's dimension cardinality
## Summary

A bare arrayed reference that shares a reducer with a SLICED co-source is executed with a redundant extra iteration of its (already-bound) dimension: the reducer's result is multiplied by the cardinality of the bare ref's dimension. This is an ENGINE EXECUTION bug (LTM disabled; no synthetic variables involved), producing silent wrong simulation results for a legal, natural XMILE shape.

## Fixture (reproduced at commit `ef91194d` with a TestProject fixture)

```
D1 = {a, b}, D2 = {c, d}
base[D1] = {a: 1, b: 100}
w[D2]    = {c: 1, d: 2}
matrix[D1, D2] = base[D1] * w[D2]
frac[D1] = {a: 10, b: 1000}
growth[D1] = SUM(matrix[D1, *] * frac)
```

Expected per XMILE A2A semantics: the bare `frac` denotes the active element `frac[r]`; the SUM iterates the wildcard D2 only. So `growth[a]` = (1 + 2) * 10 = **30**.

Engine computes `growth[a]` = **60** = |D1| * rowsum(a) * frac[a], and `growth[b]` = **600000** = 2 * 300 * 1000. The bare `frac` appears to add a redundant iteration of the already-bound D1 to the SUM's iteration space, while both `matrix`'s D1 index and `frac` still resolve to the active slot -- so each row's correct term is counted |D1| times.

## Relationship to GH #779

This is the execution side of GH #779's fixture. #779's issue text described the executed behavior as "broadcasts across all D1 rows (growth_0 = 10 = the sum of all 2x2 terms)" -- on #779's symmetric all-ones fixture that observation is numerically indistinguishable from the x|D1| factor confirmed here (2x2 all-ones: rowsum * |D1| = 2 * 2 * frac... both readings give 10 with frac frozen at 1). The asymmetric fixture above discriminates: the engine does NOT sum all 2x2 terms; it multiplies the correct per-row result by |D1|. Either way the executed value disagrees with XMILE A2A semantics, and #779 (an LTM issue) treats execution as ground truth rather than tracking this engine defect.

Note: the bare-arrayed-ref-ALONE-in-a-reducer shape (`SUM(other)` with no sliced co-source) executes per-slot as expected (no extra factor); the bug requires the sliced co-source alongside the bare ref.

## Why it matters

Silent wrong simulation results (not a compile error, not a diagnostic) for the natural spelling "per-row total = SUM(row * shared per-row coefficient)" with the coefficient written bare. Severity: medium -- silent numeric miscomputation in the core engine.

## Components affected

- `src/simlin-engine/src/compiler/` -- the wildcard-preservation / iteration-context machinery (likely `context.rs` `preserve_wildcards_for_iteration` and/or `subscript.rs` expansion): the bare arrayed ref under an active A2A dimension inside a reducer whose co-source pins that dimension as iterated appears to re-enter the iteration space instead of staying bound to the active slot.

## Possible approach

Trace the lowering of `SUM(matrix[D1,*] * frac)` under the active A2A dimension D1: the reducer's iteration-space derivation presumably unions the bare ref's dims into the iterated set even though D1 is already bound by the A2A context, while the loads themselves stay slot-resolved (hence x|D1|, not a true cross-product). A fix should gate on a regression test with the exact asymmetric fixture above (expected growth = {a: 30, b: 600}) plus the no-co-source control (`SUM(frac)` per-slot) to pin both behaviors.

## Related

- #781 -- mixed-arity multi-source reducer hard compile failure (different shape, same compiler region)
- #365 / #366 (closed) -- earlier preserve_wildcards conflation issues in the same machinery
- #779 -- the LTM-side issue whose fixture exposed this; its fix (commit `ef91194d`) is unaffected

## Discovery context

Empirically confirmed during the adversarial review of commit `ef91194d` ("engine: decline bare-spelled feeder of un-hoisted reducer loudly", GH #779, branch `ltm-phase1-residuals`), reproduced at that commit with TestProject fixtures and LTM disabled. Pre-existing, not a regression from that commit.

### LINKED ISSUE #790: ltm: scalar feeder of a whole-RHS variable-backed reduce emits non-compiling link scores (warned 0-stub loops instead of the #758 drop)
## Summary

The SCALAR feeder of a whole-RHS variable-backed reduce gets link-score variables from `try_scalar_to_arrayed_link_scores` whose fragments FAIL to compile, so every loop through the feeder degrades to a warned constant-0 stub -- instead of either being scored (the machinery exists for the subexpression spelling) or dropped via `unscoreable_edges` (the #758/#780 contract for unscoreable edges).

## Fixture (reproduced at commit `ef91194d` with a TestProject fixture)

```
D1 = {a, b}, D2 = {c, d}
pop[D1]    -- stock
matrix[D1, D2]
scale      = SUM(pop[*]) * 0.005     -- scalar
growth[D1] = SUM(matrix[D1, *] * scale)   -- whole-RHS variable-backed reduce
... feedback closed through pop -> scale -> growth -> pop
```

The user equation compiles and simulates correctly. LTM emits `$:ltm:link_score:scale->growth[a]` / `[b]` via `try_scalar_to_arrayed_link_scores`, whose fragments fail to compile (Assembly warnings, constant-0 slots), and the scale-closure loop scores u1/u2 degrade to warned 0-stubs.

## Why this is a gap, not just a degradation

- The subexpression spelling of the SAME dataflow scores fine: `growth[D1] = 0.1 + SUM(matrix[D1,*] * scale)` routes the scalar feeder through `generate_scalar_feeder_to_agg_equation` and yields a real score (~0.52). So the scoring machinery exists; the whole-RHS variable-backed-reduce target just never reaches it -- `try_scalar_to_arrayed_link_scores` does not handle that target shape.
- When a shape genuinely cannot be scored, the standing contract (#758, and #780 for the remaining gap in it) is ONE loud warning + NO link-score variable + dependent loop scores DROPPED via `unscoreable_edges`. Emitting variables that are known not to compile produces the warned-0-stub middle state both of those issues exist to eliminate: constant-0 series that look like computed values, plus cascading per-loop warning noise.

So this is loud (warned), not silent -- but it is the wrong degradation mode AND an expressiveness gap with a working sibling path.

## Components affected

- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `try_scalar_to_arrayed_link_scores` does not handle the variable-backed-reduce target; the emitted fragments reference the reduce shape in a way that fails assembly. Fix should either route the scalar feeder of a variable-backed reduce through the agg-feeder path (`generate_scalar_feeder_to_agg_equation`, matching the subexpression spelling), or decline the shape into `unscoreable_edges` per the #758 contract.

## Related

- #758 (closed) -- established the loud-skip + loop-drop contract this shape should follow if declined
- #780 -- the PartialEquationError leg of the same contract gap (skips that don't populate `unscoreable_edges`); this issue is the emit-then-fail-compile leg
- #777 -- the arrayed-owner NO-Iterated-axis variable-backed sibling (declined loudly; distinct shape -- this fixture's slice HAS an Iterated D1 axis)
- #752 / #738 (closed) -- earlier variable-backed-reduce and scalar-coefficient agg compile siblings
- Part of LTM tracking epic #488 (cluster A)

## Discovery context

Empirically confirmed during the adversarial review of commit `ef91194d` ("engine: decline bare-spelled feeder of un-hoisted reducer loudly", the GH #779 fix, branch `ltm-phase1-residuals`), reproduced at that commit with TestProject fixtures. Pre-existing, not a regression from that commit; out of scope for #779's fix.

### LINKED ISSUE #791: ltm: I1-declined multi-source reducer with subset-dim arrayed owner falls through to cartesian derivation (silent +1.0 link scores on unread rows)
## Summary

An I1-DECLINED multi-source whole-RHS reducer with an arrayed owner whose dims are a strict SUBSET of the source's dims falls through `try_cross_dimensional_link_scores` (`src/simlin-engine/src/db/ltm/link_scores.rs`) to the legacy CARTESIAN partial-reduce derivation, emitting silently wrong per-`(row, slot)` link scores -- including confident +1.0 constants for source rows the equation NEVER reads.

Shape:

```
Region = {nyc, boston}, D2 = {p, q}
pop[Region, D2]                          -- source (fed by a stock)
w[D2]                                    -- co-source
share[Region] = SUM(pop[nyc,*] * w[*])   -- mismatched co-source slices: [Pinned, Reduced] vs [Reduced]
inflow[Region] = share[Region] * 0.01    -- closes the loop back into pop
```

`accept_source_slices` declines the mismatched co-source slices, so NO variable-backed AggNode is minted (I1-declined). But the legacy cartesian branch's `result_axis_names` containment check passes (`Region` is a subset of `{Region, D2}`), so instead of the loud co-source degradation the design's boundary (c) promises for I1-declined feeders, the edge takes the cartesian partial-reduce derivation.

## Empirically verified behavior (repro fixture at commit `0fcfba58`, branch `ltm-phase1-residuals`; PRE-EXISTING)

1. **Silent wrong LINK scores.** Diagonal per-`(row, slot)` link scores are emitted for EVERY source row including UNREAD rows: `$:ltm:link_score:pop[boston,p]->share[boston]` exists and reads a constant confident **+1.0**, although `share[boston] = SUM(pop[nyc,*]*w[*])` never reads `pop[boston,*]`. The read rows are also wrong: `pop[nyc,p]->share[nyc]` reads 1.0 where the true SUM contribution share is 0.5 (two equal rows) -- the un-pinnable mismatched-arity body (`w[*]`) dooms the changed-first partial to the delta-ratio fallback (|dz/dz| = 1). Zero warnings on this surface.
2. **Only the LOOP surface degrades loudly.** 5 Assembly warnings; `$:ltm:loop_score:u1..u5` fail fragment compile because the conservative cross-product element edges create circuits referencing off-diagonal names like `pop[boston,p]->share[nyc]` that the cartesian derivation never emits, so those loops are warned 0-stubs.

So the loop surface is the (wrong-mode but loud) degradation, while the link-score surface carries unwarned wrong numbers -- violating epic #488's standing "no silent wrong numbers" invariant.

## Why it matters

Severity: medium. This is the pre-T3 "cartesian garbage" class (cf. the #765 / #777 history) SURVIVING for the no-agg multi-source family. #765's fix corrected the slice derivation for agg-backed edges, and T3's gate routes the no-Iterated-axis shapes to the #758 loud skip (#777) -- but when the arrayed owner's dims happen to be a subset of the source's dims, the I1-declined multi-source shape slips past both and lands on the legacy cartesian branch, producing confident constants on unread rows with no warning.

## Components affected

- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `try_cross_dimensional_link_scores`: the legacy cartesian partial-reduce branch's `result_axis_names` containment check admits I1-declined no-agg multi-source edges
- `src/simlin-engine/src/ltm_agg.rs` -- `accept_source_slices` (the I1 decline that leaves the edge agg-less)

## Suggested direction

The cartesian partial-reduce branch should DECLINE (loud skip via `unscoreable_edges`, per the #758/#780 contract) when `classify_reducer`'s body references arrayed deps that are not resolvable per-row, or more generally when no agg backs the edge and the body is not the bare source -- or at minimum, gate the cartesian derivation on a variable-backed agg existing for the edge.

## Related

- #765 (closed) -- value-level sibling for the agg-backed family (Pinned axes in the co-reduced slice); this issue is the surviving no-agg multi-source leg
- #777 -- the no-Iterated-axis I1-declined sibling that IS loud-skipped (fails the containment check); this shape passes it
- #758 (closed) / #780 -- the loud-skip + loop-drop contract the decline should follow
- #488 -- LTM tracking epic (cluster A)

## Discovery context

Empirically confirmed during the adversarial review of commit `0fcfba58` on branch `ltm-phase1-residuals`; pre-existing, not introduced by that commit.

### LINKED ISSUE #792: ltm: per-element-equation (Ast::Arrayed) owner with strict-slice multi-source reducer gets a silent near-zero Bare A2A link-score stand-in
## Summary

A PER-ELEMENT-EQUATION owner (`Equation::Arrayed`, one equation per slot) whose slots hold an I1-declined strict-slice multi-source reducer never reaches `try_cross_dimensional_link_scores`' cartesian arm. Instead the edge falls through to `emit_per_shape_link_scores` with shape **Bare**, which emits ONE arrayed link score whose simulated value is ~-0.0 at every step -- a silent near-zero stand-in for an edge the equation genuinely depends on, with no per-edge warning.

This is the per-element spelling of the GH #791 shape. Neither the pre-791 cartesian garbage nor the new #791 strict-slice loud skip applies here: the gate's `Ast::Arrayed => NotDescribable` arm in `ltm_agg::unhoisted_reducer_source_read` is unobservable for this shape because the edge never gets that far.

## Fixture (verified empirically 2026-06-12 at commit `c1a16df4`, branch `ltm-phase1-residuals`; PRE-EXISTING -- the per-element path is untouched by that commit)

```
Region = {nyc, boston}, D2 = {p, q}
pop[Region, D2]                          -- source (fed by a stock)
w[D2]                                    -- co-source
share[Region]                            -- Equation::Arrayed, EACH slot:
  share[nyc]    = SUM(pop[nyc,*]    * w[*])
  share[boston] = SUM(pop[boston,*] * w[*])
inflow[Region] = share[Region] * 0.01    -- closes the loop back into pop
```

(TestProject fixture `array_with_ranges_direct` shape; sim 0..8, dt 1.)

## Observed behavior

1. One arrayed `$:ltm:link_score:pop->share` (2 slots) is emitted via the Bare per-shape path; its simulated value is **-0.0 at every step**, although `share` genuinely depends on `pop`.
2. Only ONE warning surfaces -- a single loop-score fragment-compile failure (`$:ltm:loop_score:u5`). The other enumerated loops compile and consume the silent ~0 link score, so their loop scores are unwarned wrong (near-zero) numbers.

This violates epic #488's standing "no silent wrong numbers" invariant: a near-zero stand-in score with no per-edge warning.

## Why it matters

Severity: medium. The MDL importer expands single apply-to-all Vensim equations into N identical per-element equations (#651), so this spelling is exactly what real Vensim imports produce -- the per-element variant is arguably MORE reachable from imported models than the A2A spelling #791 fixed.

## Components affected

- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `try_cross_dimensional_link_scores` (per-element owners never reach the cartesian/strict-slice arm) and `emit_per_shape_link_scores` (the Bare fallthrough that mints the near-zero stand-in)
- `src/simlin-engine/src/ltm_agg.rs` -- `unhoisted_reducer_source_read`'s `Ast::Arrayed => NotDescribable` arm is dead for this shape; the strict-slice decline logic needs a per-element-aware entry point

## Suggested direction

Route the per-element-equation owner spelling to the same loud-skip contract #791's fix applies to the A2A spelling: detect the un-hoisted strict-slice multi-source reducer in `Equation::Arrayed` slots and decline via `unscoreable_edges` (warning + dependent loop drop per the #758/#780 contract), instead of letting the edge fall to the Bare per-shape stand-in. Longer term, per-slot pinned slices are statically describable (each slot pins its row), so a per-element hoist may be able to score this shape exactly.

## Related

- #791 -- the A2A/scalar-owner spelling of the same shape (fixed by `c1a16df4`); this is its per-element sibling
- #758 (closed) / #780 -- the loud-skip + loop-drop contract these edges should take
- #651 -- the MDL importer per-element expansion that makes this spelling the import-reachable one
- #488 -- LTM tracking epic (cluster A)

## Discovery context

Empirically confirmed during the adversarial review of commit `c1a16df4` (the GH #791 strict-slice cartesian decline) on branch `ltm-phase1-residuals`; pre-existing, not introduced by that commit.

### LINKED ISSUE #793: ltm: strict-slice reducer read silently unattributed when a sibling hoisted reducer routes the edge through its agg
## Summary

In `emit_link_scores_for_edge` (`src/simlin-engine/src/db/ltm/link_scores.rs`, ~line 2893 at commit `c1a16df4`), when ANY synthetic agg of `to` reads `from`, `routed_aggs` is non-empty and the `(from, to)` edge is scored EXCLUSIVELY via the agg halves (`{from}->$⁚ltm⁚agg⁚{n}` and `$⁚ltm⁚agg⁚{n}->{to}`). `try_cross_dimensional_link_scores` -- and therefore the GH #791 strict-slice gate added by `c1a16df4` -- is never consulted for that edge.

So when the same equation contains a SECOND, I1-DECLINED strict-slice reducer read of the same source, that read contributes to the target's value but gets NO link-score attribution and NO warning. The agg halves are correct for the hoisted reducer ALONE; the strict-slice read's contribution is silently missing from the edge's total attribution.

## Empirical repro (verified 2026-06-12 at `c1a16df4`, branch `ltm-phase1-residuals`)

```
Region = {nyc, boston}, D2 = {p, q}
pop[Region, D2] = stock[Region] * 0.1
w[D2] = 0.5
share[Region] = SUM(pop[nyc, *] * w[*]) + SUM(pop[*, *])
inflow[Region] = share * 0.05
stock integrates inflow
```

- `SUM(pop[*,*])` hoists cleanly into `$⁚ltm⁚agg⁚0`, so `routed_aggs` is non-empty for the `(pop, share)` edge.
- `SUM(pop[nyc,*] * w[*])` is I1-declined (mismatched co-source slices), the exact shape #791's strict-slice gate exists to catch -- but the gate never runs because the edge took the agg-exclusive branch.

Observed:

- ZERO Assembly warnings.
- The only `pop -> share` scores are the agg halves: `pop[r,d] -> $⁚ltm⁚agg⁚0` = 0.25 and `$⁚ltm⁚agg⁚0 -> share[e]` = 0.8 -- both correct for the second reducer ALONE. The missing 0.2 of `share`'s change (the strict-slice read's contribution) is attributed to nothing.
- 8 loop scores (r1-r8) are emitted from this partial attribution, so downstream loop dominance is computed from incomplete numbers with no signal.

## Why it matters

Severity: medium. Same "silent wrong/incomplete number for a strict-slice read" class as GH #791, surviving on the agg-routed leg -- it violates epic #488's standing no-silent-wrong-numbers invariant (an unattributed contribution is indistinguishable from a complete attribution on every surface).

## Root cause

The reference-site IR's per-site routing criterion is too coarse: `route_through_agg = (any agg of `to` reads `from`) && in_reducer` (`src/simlin-engine/src/db/ltm_ir.rs`). The site INSIDE the DECLINED reducer is routed through the OTHER reducer's agg, because the criterion keys on the (from, to) pair plus "inside some reducer", not on WHICH reducer the site sits in / whether that reducer was actually hoisted.

## Possible approaches

- Make the routing criterion site-precise: a site routes `ThroughAgg { agg }` only when it lies inside the specific reducer that minted `agg`; sites inside a declined reducer stay `Direct` and flow to `try_cross_dimensional_link_scores`, where the #791 strict-slice gate (loud skip via `unscoreable_edges` per the #758/#780 contract) applies.
- Alternatively, when `routed_aggs` is non-empty but the edge has residual non-agg-routed sites, still consult the cross-dimensional/strict-slice path for those sites (or at minimum loud-skip the whole edge instead of emitting confidently partial agg halves).

## Components affected

- `src/simlin-engine/src/db/ltm_ir.rs` -- per-site routing criterion (`route_through_agg`)
- `src/simlin-engine/src/db/ltm/link_scores.rs` -- `emit_link_scores_for_edge` (~line 2893): the `routed_aggs`-non-empty branch scores the edge exclusively via agg halves

## Related

- #791 -- the no-agg leg of the same strict-slice class, fixed by `c1a16df4`; this is the surviving agg-routed sibling
- #758 (closed) / #780 -- the loud-skip + loop-drop contract a decline should follow
- #514 (closed) -- the original sliced-reducer hoisting work; the I1 decline family descends from its acceptance gates
- #488 -- LTM tracking epic (no-silent-wrong-numbers invariant)

## Discovery context

Discovered during the adversarial review of commit `c1a16df4` (gh791 strict-slice cartesian decline) on branch `ltm-phase1-residuals`. PRE-EXISTING: the routing logic is untouched by that commit.

### LINKED ISSUE #794: engine: VM panic (out-of-bounds temp slice) on SUM(RANK(<sliced 1-D view>)) in a per-element Arrayed equation
## Summary

A compile-clean model panics the VM at simulation time with an out-of-bounds slice when a per-element (`Equation::Arrayed`) slot equation wraps an array-producing `RANK` over a sliced 1-D view of a 2-D array in a reducer:

```
panicked at src/simlin-engine/src/vm.rs:2946:32:
range end index 6 out of range for slice of length 4
```

This is a process-aborting panic, not an error result, so it also takes down library consumers (libsimlin, pysimlin, WASM hosts). No LTM involved (LTM disabled in the repro).

## Repro

Verified 2026-06-12 at commit `00b6bc0a` (branch `ltm-phase1-residuals`) and byte-identically at the earlier commit `2fc2f2ee` -- pre-existing, NOT introduced by the gh792 commits.

```
Region = {nyc, boston}, D2 = {p, q}
pop[Region, D2] = stock[Region] * 0.1
share[Region]  (Equation::Arrayed):
  share[nyc]    = SUM(RANK(pop[nyc, *], 1))
  share[boston] = SUM(RANK(pop[boston, *], 1))
inflow[Region] = share[Region] * 0.05
stock[Region] integrates inflow, initial 100; sim 0..8 dt 1
```

Built via `TestProject` (`array_with_ranges_direct`), compiled via `compile_project_incremental` (compiles cleanly), then `Vm::new(compiled)` + `vm.run_to_end()` panics as above.

## Analysis

`vm.rs:2946` is the dense-linear fast path of the array-reduce fold:

```rust
for &value in &data[base + start..base + start + size]
```

The SUM reads a temp view (the hoisted RANK result) whose computed `start + size` exceeds the backing temp storage length. Likely cause: the A2A/per-element hoisting of the array-producing RANK builtin (`AssignTemp` + `TempArrayElement`; see the hoisting of `VectorSortOrder`/`Rank` in `compiler/mod.rs`) produces a temp view sized/based inconsistently with what the wrapping reducer's dense-linear fold expects when RANK's argument is a sliced view (`pop[nyc, *]`, a 1-D slice of a 2-D array) inside an `Equation::Arrayed` slot equation.

## Why it matters

- Crash on valid-looking input that compiles cleanly -- severity medium-high.
- Panic (abort), not a recoverable error, crossing into FFI consumers.
- `RANK` inside a reducer in per-element slot equations is reachable from Vensim imports via `VECTOR RANK`.

## Affected component

`src/simlin-engine` -- VM array-reduce fold (`src/simlin-engine/src/vm.rs:2946`) and/or the RANK temp-hoisting path in `src/simlin-engine/src/compiler/mod.rs`.

## Possible approaches

- Audit the temp-view metadata (base/start/size) emitted for hoisted array-valued builtins when the argument is a sliced view in a per-element slot, and make it consistent with the reducer fold's expectations.
- Add a defensive bounds check in the dense-linear fast path that surfaces an internal error instead of panicking.
- Regression test: `SUM(RANK(arr[elem, *], 1))` in an `Equation::Arrayed` slot over a 2-D array.

## Dedup notes

Distinct from #776 (LTM diagonal-only element edges for de-hoisted RANK), #771/#742 (LTM RANK hoist/capture scoring), #781 (fragment-compile failure for mixed-arity reducers), and #789 (wrong values for bare ref sharing a reducer). This is a runtime VM crash, not an LTM-scoring or compile-time issue.

Identified while investigating the gh792 residuals (LTM phase 1) -- the repro fails identically with LTM disabled.

bpowers:
## Summary

- closes the remaining Phase 1 residuals by declining residual unhoisted reducer reads, preserving spelling-specific agg routing, and dropping missing-name loop scores before zero stubs can be compiled
- fixes execution anomalies in reducer temp shape derivation and sliced temp-array view metadata, including the sliced `RANK` VM panic
- keeps the loud-safe contract across engine and FFI diagnostics: unsupported loop scores warn and are not emitted rather than silently producing wrong or stubbed numbers

## Tests

- `cargo test -p simlin-engine --test integration --quiet`
- `cargo test -p simlin-engine --lib --quiet`
- `cargo test -p simlin test_get_errors_surfaces_ltm_fragment_failure_after_ltm_sim --test integration --quiet`
- `git commit` pre-commit hook: dependency policy, docs links, project lint rules, Rust fmt/cbindgen/clippy/tests, TypeScript checks, pysimlin tests

Fixes #780
Fixes #779
Fixes #783
Fixes #778
Fixes #785
Fixes #777
Fixes #791
Fixes #790
Fixes #754
Fixes #792
Fixes #793
Fixes #789
Fixes #788
Fixes #794


bpowers:
@codex review

bpowers:
@codex review

Issue #806, opened 2026-06-21
engine: unguarded panics on dangling module model_name in legacy from_salsa path (module_deps, topo_sort)


bpowers:
## Problem

Two latent unguarded-panic sites on the legacy `Project::from_salsa` dependency-resolution path crash with an "internal compiler error" / HashMap-index panic on user-controllable input (a module with an empty or dangling `model_name`, e.g. a freshly-drawn module or a reference to a deleted model).

### 1. `model.rs:173` — unguarded HashMap index in `module_deps`

In the `is_initial` branch of `module_deps` (`src/simlin-engine/src/model.rs`):

```rust
if ctx.is_initial {
    let model = ctx.models[model_name];   // <- panics if model_name not in ctx.models
```

This is an unguarded `Index` on `ctx.models`. If a `Variable::Module` references a `model_name` that is not present in `ctx.models`, this panics with "no entry found for key" / index panic.

The sibling at `model.rs:241-244` (`module_output_deps`) **already guards this correctly**:

```rust
if !ctx.models.contains_key(model_name) {
    return model_err!(BadModelName, model_name.to_string());
}
let model = ctx.models[model_name];
```

Line 173 should do the same — use `.get()` and return `BadModelName` (or skip gracefully) rather than indexing unconditionally. The fix is a near-mechanical mirror of the guarded twin 70 lines below it.

### 2. `common.rs:2659` — `topo_sort` panics on an unknown ident (lower priority, same class)

```rust
panic!("internal compiler error: unknown ident {}", ident.as_str());
```

`topo_sort` panics when a runlist ident has no dependencies entry. This is reachable from the same legacy `from_salsa` path with a dangling module reference. Lower priority but the same class of latent crash on bad input — should surface as a proper error rather than a `panic!`.

## Why it matters

- **Unguarded panic on user-controllable input**: an empty/dangling `model_name` is something the UI can produce (a module drawn but not yet wired up, or a reference to a model that was deleted). A panic here is an "internal compiler error" crash rather than a clean diagnostic.
- **Inconsistency**: `model.rs:173` is the unguarded twin of an already-guarded sibling (`model.rs:241-244`) in the same file. The guard pattern is established; line 173 simply was not updated to match.
- **WASM**: under `panic = "abort"` (WASM release builds) a panic aborts the entire runtime.

## Currently latent (why it is tracked, not urgently fixed)

[...]

PR #807, opened 2026-06-23
Module editing: fix dangling/cyclic/mis-wired reference bugs found by adversarial audit

bpowers:
## Why

Commit c1c4c954 fixed a WASM panic where a freshly-drawn module (empty `model_name`) crashed unit inference. That was one instance of a broader **dangling / empty / cyclic / mis-wired module-reference** bug class. An adversarial audit of the module-editing feature (engine patch/compile pipeline, the JSON/FFI boundary, MCP, and the diagram controller) surfaced five more production-reachable issues plus one severe lurking bug; this PR fixes all of them, each with tests that reproduce the failure end to end.

## Fixes

**`engine: keep module references consistent under edits`** — module wiring stores a parent `src` and a module-qualified `module·port` `dst`, which the patch layer was not keeping consistent across structural edits:
- `apply_delete_variable` pruned deleted flows from stocks and groups but left module references whose `src` named the deleted variable, so deleting a wired-in parent variable left a dangling dependency and the whole project failed to compile with a confusing "missing variable" error. It now prunes those references too.
- `apply_rename_variable` rewrote module references only within the renamed variable's own model. Renaming a *child* model's input port left every parent module's `dst` pointing at the old name, silently feeding the renamed port its default value (wrong numbers, no error). The rename now retargets the qualified `dst` suffix of every module instance that targets the renamed model.
- `canonicalize_module` canonicalized only the module ident, so a non-canonical `src`/`dst` arriving via the FFI `apply_patch` (pysimlin `upsert_module`) disagreed with the canonical idents every consumer compares against. It now canonicalizes the endpoints, mirroring `canonicalize_stock`.

**`engine: reject cyclic module graphs instead of panicking`** — nothing checked that the module-instance graph is acyclic, so a self-referential or mutually-recursive module drove the recursive salsa queries (`model_module_map`, `compute_layout`) into a dependency-graph cycle, which salsa turns into an unrecoverable panic — reached from MCP `edit_model`/`read_model` and from importing a hand-authored recursive model, aborting WASM. A new `project_module_cycle` query detects the cycle with flat reads + an iterative DFS (no recursive salsa) and the three compile / diagnostic / analysis entry points fail cleanly with a `CircularDependency` error instead. This is the **production-reachable cousin** of the `from_salsa` panic class tracked in #806 — #806's own dead legacy sites are not reachable from production and are left as tracked; the live path is what this guards.

**`engine: warn on unresolved module input wiring`** — `build_module_inputs` silently drops a reference whose `dst` does not match an existing child input, so the port reads its default and the simulation is quietly wrong with no error. The legacy compile path returned `BadModuleInputDst`/`BadModuleInputSrc`; the salsa path had dropped the check. A new diagnostic pass re-adds it as a **Warning** (partial-result philosophy — a mis-wired input should not block the rest of the model), deliberately conservative to avoid false positives (empty placeholder rows skipped, only existing target models checked, qualified/synthetic `src`s left alone).

**`diagram: persist module input dst in the engine-qualified form`** (the lurking bug) — the wiring UI dropdown yields a *bare* child port name, and `ModuleDetails` persisted that bare value straight into the reference `dst`. But the engine wires inputs by the canonical `module·port` form and silently drops a `dst` that lacks the prefix, so **every module input wired through the editor failed to connect**. XMILE import already produces the qualified form, so this was an editor-only divergence from the datamodel contract. The fix qualifies the port on persist and strips it on display (pure `qualifyDst`/`unqualifyDst` helpers), tolerating a legacy `.` separator on read.

**`core: cover module serialization edge cases`** — the only Module round-trip test exercised a fully-populated module. Added cases for the empty-`modelName` freshly-drawn state (the c1c4c954 panic shape; the empty name is a required Rust `String`), placeholder references with empty `src`/`dst`, and a module-qualified `dst` whose middot must survive the JSON round trip.

## Non-obvious decisions

- **Cycle detection lives in the engine, not the frontend.** The diagram's `wouldCreateCycle` only filters the model-reference dropdown and is bypassed by MCP and file-import, so the engine must own the guard for every entry point.
- **The wiring check is a Warning, not a hard error**, matching the engine's partial-result philosophy and avoiding breaking existing models with benign dangling references; the structural fixes (delete-prune, rename-retarget, editor qualification) close the ways these arise in the first place.
- **`dst` is qualified in the editor, not bare.** The canonical datamodel/XMILE representation is `module·port`; making the editor conform (rather than making the engine accept bare `dst`) keeps a single representation across import/export/edit.

## Testing

Every fix ships with a test that reproduces the original failure: the engine patch tests compile and simulate a wired module (asserting the value flows, or cleanly drops to the port default); the cycle tests assert a clean error across compile, diagnostics, and analysis (and that valid nesting does not false-positive); the wiring-diagnostic tests pin bare/dangling `dst` and `src` warnings while a correct wiring and empty rows stay clean; the diagram tests cover the qualify/unqualify helpers and the UI persist/display round trip; and the core tests cover the serialization edge cases. Each commit was gated by the full pre-commit suite (Rust fmt/clippy/test, TypeScript lint/build/tsc/test, pysimlin).

Related: #806 (the production-reachable cousin of that dangling-reference panic class is fixed here; #806's specific dead `from_salsa` sites remain tracked).

### DIFF
--- a/src/core/tests/datamodel.test.ts
+++ b/src/core/tests/datamodel.test.ts
@@ -305,6 +305,82 @@ describe('Module', () => {
     expect(restored.modelName).toBe('SectorModel');
     expect(restored.references.length).toBe(1);
   });
+
+  it('roundtrips a freshly-drawn module (empty modelName, no references)', () => {
+    // The editor creates a module with an EMPTY modelName until the user
+    // assigns a target model -- the exact state that panicked the engine in
+    // c1c4c954. The empty modelName must survive serialization (the Rust
+    // `model_name` is a required String, so a dropped field is an FFI hard
+    // error) and deserialize back to empty.
+    const mod: Module = {
+      type: 'module',
+      ident: 'new_module',
+      modelName: '',
+      documentation: '',
+      units: '',
+      references: [],
+      data: undefined,
+      errors: undefined,
+      unitErrors: undefined,
+      uid: undefined,
+    };
+
+    const json = moduleToJson(mod);
+    expect(json.modelName).toBe('');
+    expect(json.references ?? []).toEqual([]);
+
+    const restored = moduleFromJson(json);
+    expect(restored.modelName).toBe('');
+    expect(restored.references).toEqual([]);
+  });
+
+  it('preserves placeholder references with empty src/dst', () => {
+    // The wiring UI persists partially-filled rows (empty src or dst) as the
+    // user builds up inputs; these must round-trip verbatim, not be coalesced
+    // or dropped.
+    const mod: Module = {
+      type: 'module',
+      ident: 'sector',
+      modelName: 'SectorModel',
+      documentation: '',
+      units: '',
+      references: [
+        { src: '', dst: '' },
+        { src: 'food', dst: '' },
+      ],
+      data: undefined,
+      errors: undefined,
+      unitErrors: undefined,
+      uid: 5,
+    };
+
+    const restored = moduleFromJson(moduleToJson(mod));
+    expect(restored.references).toEqual([
+      { src: '', dst: '' },
+      { src: 'food', dst: '' },
+    ]);
+  });
+
+  it('roundtrips a module-qualified dst reference', () => {
+    // The canonical reference dst is the module-qualified `{moduleIdent}·{port}`
+    // form (the engine strips the prefix to wire the input); the middot must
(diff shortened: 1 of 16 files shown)


bpowers:
@codex review

bpowers:
@codex review

Issue #695, opened 2026-06-03
deploy: .gcloudignore doesn't import .gitignore; gcloud app deploy uploads untracked files and trips GAE 10k-file cap


bpowers:
## Summary

`.gcloudignore` does not import `.gitignore`. There is no `#!include:.gitignore` directive — line 6 `/.gitignore` merely excludes the ignore file itself. As a result, `gcloud app deploy` (run by `pnpm deploy:web`) uploads **every file present in the working tree that is not explicitly listed in `.gcloudignore`**, completely independent of git tracking status.

This trips **GAE Standard's hard 10,000-file deploy cap**: `pnpm deploy:web` does a full `pnpm clean` + `pnpm build` and then aborts at the `gcloud app deploy` step, after the expensive clean+build has already run.

## Evidence (measured on a working tree with `git status --porcelain` reporting ZERO untracked files)

```
$ git status --porcelain --ignored=no | grep -c '^??'
0
$ gcloud meta list-files-for-upload . | wc -l
144737            # > 14x the 10,000-file GAE Standard cap
```

Top-level breakdown of the upload set:

```
132863 third_party        # gitignored (.gitignore:19 `/third_party`), NOT in .gcloudignore
 10534 .venv-pysimlin     # gitignored (.gitignore:79 `.venv*`), NOT in .gcloudignore
  1115 env                # not gitignored, not in .gcloudignore
    96 docs
    31 .hypothesis
    29 .claude
    26 .agents
   ...
```

Two compounding root causes, both confirmed:

1. **`.gcloudignore` is independent of `.gitignore`.** `third_party/` and `.venv-pysimlin/` are *gitignored* (so `git status` is clean), yet both are uploaded because `.gcloudignore` has no matching rule and no `#!include:.gitignore`.
2. **The `/target/` rule is root-anchored.** `.gcloudignore` line 13 `/target/` only excludes the repo-root `target/`. Nested cargo target dirs escape it — `third_party/codex/codex-rs/target` exists, and `108536` of the uploaded paths contain `/target/`.

An operator's real working tree is worse: the finding that surfaced this measured ~177k untracked files (third_party ≈ 143k, .venv-pysimlin ≈ 10.5k, src/pysimlin/.venv ≈ 7.3k, .venv ≈ 5k, env ≈ 1.1k, plus playwright-report/, test-results/, simlin/, src/pysimlin/{venv,build,dist}).

## Why it matters

[...]

PR #809, opened 2026-06-27
deploy: add staged server bundle to shrink the GAE install

bpowers:
## Why

`pnpm deploy:web` runs `gcloud app deploy` from the **workspace root**, so the App Engine instance's `pnpm install` walks `pnpm-workspace.yaml` and installs *every* workspace package's dependency closure — React, slate, radix, rspress, vite, jest, `@rsbuild/*` — **~590 MB / 1171 packages**, none of which the Express server needs at runtime.

App Engine standard **always reinstalls from the deployed `package.json` + lockfile** and has no vendored-`node_modules` escape hatch (confirmed against Google's docs), so the only lever that shrinks the install is the deployed manifest. The pnpm10 `NODE_ENV=production` devDep bug ([buildpacks#591](https://github.com/GoogleCloudPlatform/buildpacks/issues/591)) is secondary: even a perfect `--prod` install of the *whole workspace* would still pull React/slate/radix/rspress/vite, because those are runtime deps of other packages. The real cost is installing the whole workspace at all.

## What

Adds **`pnpm deploy:web:staged`**, which deploys from a self-contained `deploy-staging/` whose `package.json` is exactly the server's prod closure — **~80 MB / 230 packages**.

| | install on instance | packages | upload payload |
|---|---|---|---|
| `deploy:web` (today) | 590 MB | 1171 | whole repo |
| `deploy:web:staged` | **80 MB** | **230** | **28.7 MB** |

### How it works
- `scripts/deploy-web-staged.sh` mirrors `deploy-web.sh` (validate → clean → build → `deploy:assemble` → `verify-deploy-build.sh`), then runs `scripts/build-deploy-staging.mjs` and `gcloud app deploy deploy-staging/app.yaml`.
- `@simlin/core` and `@simlin/engine` are **not published to npm**, so a `pnpm deploy` version-rewrite (`workspace:*` → `2.0.0`) would 404 on the instance. Instead they are vendored into `deploy-staging/vendor/` and referenced via `file:` deps; core's transitive `@simlin/engine` ref is rewritten too. The wasm keeps loading unchanged (the engine stays a real package dir; `wasm.node.js` resolves `core/libsimlin.wasm` via `__dirname`).
- Only the **full** server-side `libsimlin.wasm` is vendored; the browser variants are deliberately omitted (plain Node `require` never selects the `module`/browser export condition), keeping the 4.8 MB slim wasm off the server image.
- A `pnpm-lock.yaml` is generated for the minimal manifest with peer settings pinned (`--config.auto-install-peers=true --config.strict-peer-dependencies=false`) so the lockfile doesn't depend on the operator's ambient `.npmrc`.
- The pure transforms live in `scripts/deploy-staging-manifests.mjs` (unit-tested via `pnpm test:scripts`); the builder self-verifies the artifact (full wasm with `simlin_project_render_png`, substituted `index.html`, no residual `workspace:` refs) and guards against an `app.yaml` `entrypoint` that wouldn't resolve in the flat layout. Unknown workspace deps and `peerDependencies` fail loud rather than ship a broken manifest.

## Testing (local, end-to-end)
- `pnpm test:scripts`: 33 unit tests green.
- Built + assembled, ran the builder, then **simulated the GAE instance**: `pnpm install --frozen-lockfile --prod` in a copy *outside* the repo (no workspace/`.npmrc` ancestor) installs cleanly from the generated lockfile (74–80 MB / 230 pkgs); the full wasm materializes at the `__dirname` path; all server `require`s (incl. `@simlin/core/common`) resolve.
- `node lib/index.js` boots (`WASM module initialized successfully`) and serves `/` → 200, `/api/user` → 401, `/static/js/sd-component.js` → 200, `/robots.txt` → 200.
- Adversarially reviewed twice; no material findings outstanding.

## Not yet verified / why `deploy:web` stays the default
The literal `gcloud app deploy` upload and the **nodejs24 buildpack** behavior can't be tested locally: whether it honors `packageManager: pnpm@10.6.0` via corepack and accepts the frozen `lockfileVersion`. Run `pnpm deploy:web:staged --no-promote`, confirm the instance ran a successful frozen `pnpm install`, then the post-deploy smoke test — before this replaces `deploy:web`. `deploy:web` remains the proven fallback (untouched).

Relates to tech-debt #39 (and incidentally puts the upload well under GAE's 10k-file cap, #695, once this becomes the default).


bpowers:
Addressed the Codex P2 (lockfile re-resolution) in 66b8ffe6.

The staging `pnpm-lock.yaml` is now **seeded from the committed root lockfile** (its `packages`/`snapshots` resolution cache with `importers` dropped) before `pnpm install --lockfile-only`, so pnpm reuses the already-locked, CI-tested versions and only resolves the new `file:` vendored packages, then prunes to the staging closure. A downgrade test confirmed it pins rather than re-resolves (a seeded `helmet@8.1.0` is kept, not bumped to the newest `8.2.0`). A post-generation guard (`untestedPackages`) additionally fails the build if any registry version in the staging lockfile is absent from the root lockfile, so an untested version can never ship. Verified end-to-end: an isolated `pnpm install --frozen-lockfile --prod` from the seeded lockfile still boots the server and serves all routes.

bpowers:
@codex review

bpowers:
Addressed in the latest commit: `/deploy-staging/` is now excluded in the root `.gcloudignore` so a fallback `pnpm deploy:web` from the repo root won't re-upload the leftover staged artifact (or its app.yaml copy). The staged deploy uploads from `deploy-staging/` with its own `.gcloudignore`, so the entry is inert for it.

@codex review

Issue #830, opened 2026-07-02
Flow endpoint detach off a stock has no live preview (only valve moves)


bpowers:
## Problem

In `src/diagram/drawing/Canvas.tsx` `deriveRenderState`, dragging a flow endpoint (arrowhead/source) that is currently attached to a **stock** does not live-update the flow path. Only the flow valve moves during the drag; the flow line and arrowhead stay pinned to the stock until pointer-up, at which point the geometry snaps to a newly-created cloud at the release position (created by `flow-attach.ts` `reattachEndpoint`'s `newCloud` branch).

This is the same class of bug as the recently-fixed cloud-endpoint case: for an along-axis (parallel) drag, the endpoint drag routes through `applyGroupMovement` -> `UpdateFlow` (`drawing/Flow.tsx`), which treats the drag as a valve slide rather than a reroute, so the path preview is stale.

## Scope / why it wasn't fixed with the cloud case

The just-landed fix added a branch in `deriveRenderState` (Canvas.tsx around line 924-961) that calls `growEndpointDrag` (from `src/diagram/flow-attach.ts`) -> `UpdateCloudAndFlow` so the flow line, arrowhead, and valve all track the dragged endpoint. That branch is gated on `endpointEl?.type === 'cloud'`; a **stock** endpoint deliberately falls back to the old `applyGroupMovement`/`UpdateFlow` behavior. A live detach preview needs to synthesize a cloud that follows the cursor, which was out of scope for the reported cloud-only bug.

## Why it matters

UX polish / consistency in the diagram editor. Detaching a flow endpoint off a stock feels broken relative to dragging a cloud endpoint: the user drags but the flow path doesn't follow, then it jumps on release. Low-to-medium priority.

## Component(s) affected

- `src/diagram/drawing/Canvas.tsx` (`deriveRenderState`, the cloud-only endpoint-drag branch)
- `src/diagram/flow-attach.ts` (`growEndpointDrag`, `reattachEndpoint`)

## Possible approach

Extend the `growEndpointDrag`/`UpdateCloudAndFlow`-based live-preview path to stock endpoints by synthesizing a transient cloud at the dragged endpoint (mirroring `reattachEndpoint`'s `newCloud` branch) so the flow line tracks the cursor during a detach drag, matching the committed geometry and the cloud-drag feel.

## Discovered

While fixing the cloud-endpoint live-update bug (the fix that added the `growEndpointDrag` branch in `deriveRenderState`).

Issue #829, opened 2026-07-02
diagram: units editor accepts Enter, inserting literal newlines into the units string


bpowers:
## Problem

The units field in the variable details panel is semantically a single-line value, but it is rendered as a plain Slate editor with no single-line normalization and no `onKeyDown` handler:

- `src/diagram/VariableDetails.tsx`: the units `<Editable>` (around line 534) has only `renderLeaf`, `placeholder`, `spellCheck`, and `onBlur` -- unlike the equation and notes editors, which deliberately handle newline chords. Pressing plain Enter (or Shift+Enter, which Slate handles by default) inserts a new block/newline.
- `stringFromDescendants` -> `plainSerialize` (`src/diagram/drawing/common.ts`) joins Slate blocks with `'\n'`, so on blur/save a literal `\n` is committed into the variable's units string.
- `src/diagram/ModuleDetails.tsx` has the same pattern for its units editor (`plainSerialize(unitsContents)` at line ~109, no keydown handling).

## Why it matters

Newlines inside a units expression are almost certainly meaningless or invalid to the engine's unit parser, so a stray Enter keypress silently corrupts the units string and can surface as a confusing unit error (or be persisted into the saved model). It's also a UX inconsistency: the sibling equation and documentation editors have explicit, commented newline-chord handling, while the single-line units field has none.

## Possible approaches

- Add an `onKeyDown` to the units `<Editable>` that intercepts Enter (and Shift/Cmd/Ctrl+Enter): either make Enter commit the edit (blur/save) or make it a no-op.
- And/or add a `withSingleLine` Slate normalizer (`normalizeNode` merging blocks / stripping `\n` in `insertText`) so pasted multi-line content is also flattened -- keydown alone doesn't cover paste.
- Alternatively, if multi-line units are intentionally tolerated, document that decision and verify the engine's unit parser treats `\n` as ordinary whitespace.

Applies to both `VariableDetails.tsx` and `ModuleDetails.tsx` units fields.

## Discovery context

Identified while reviewing keyboard handling in `VariableDetails.tsx` (the equation editor's `isNewlineChord` handling made the units editor's complete absence of Enter handling stand out).

PR #831, opened 2026-07-02
diagram: editor bug burndown (caret mapping, Cmd+Enter, dblclick rename, connector sync, cloud drag)

bpowers:
Five user-reported editor bugs, fixed one commit each. Each fix was implemented and then adversarially reviewed (independent review pass verifying root cause, running tests against pre-fix code, and probing for regressions) before landing.

## 1. Caret placement on rendered fractions (04e588a6)

Clicking the fraction bar or denominator of a KaTeX-rendered division placed the caret at a garbage offset. Root cause: the click handler resolved the annotated source span via `Element.closest` (ancestor-only), but KaTeX renders fraction operands inside unannotated `.vlist`/`.frac-line` layout chrome, so those clicks resolved to the whole-fraction span and interpolated across the entire equation. Now all annotated spans are gathered and the smallest box containing the click wins (pure `chooseSpanForClick`), generalizing to any nested expression.

**Design note**: a dead-center fraction-bar click maps into an adjacent operand rather than exactly at the `/` -- the bar carries no annotation and is geometrically ambiguous, so that is the principled behavior.

## 2. Cmd+Enter inserts a newline (f714c2c1)

Cmd+Enter did nothing in the equation/documentation editors: modifier+Enter chords produce no `beforeinput` event, and Slate's legacy keydown fallback (`is-hotkey`) requires unnamed modifiers to be absent. New `isNewlineChord` routes Cmd+Enter through `Editor.insertSoftBreak` -- the identical operation Shift+Enter triggers.

**Design notes**: Ctrl+Enter is treated the same so the gesture works on non-mac keyboards (nothing else binds it in those fields). The units editor (semantically single-line) and the canvas name editor (Enter/Ctrl+Enter commit there) deliberately do not get the chord. Discovered along the way: the units editor already accepts plain-Enter newlines into a single-line field -- filed as #829 rather than scope-creeping here.

## 3. Reliable double-click rename (1942bdf5)

Two root causes. Deterministic: a label double-click on an *already-selected* variable routed into the deferred-single-select path whose name-edit resolution only ran on pointer-up -- unreachable for a terminal `dblclick` event, so the editor never opened (the interaction union's `deferredIsText` field was provably dead code, now removed). Flaky: labels had no click-vs-drag threshold, so 1-2px of double-click wobble started a label drag and captured the pointer, suppressing the native `dblclick`. Labels now use the same 5-screen-pixel threshold as the rest of the canvas.

## 4. Connector/equation sync warnings (3791f798)

New feature: variables now show the existing warning-dot (plus explanatory text in the details panel) when the sketch and equations disagree -- an equation references a variable with no connector drawn from it, or a connector's source is no longer used by the equation.

**Design decisions**:
- Dependency truth comes from the engine's existing `Model.getIncomingLinks` (no new FFI). Using `getLinks`/causal edges instead would be wrong: those add structural flow-to-stock integration edges (which no connector represents) and omit stock initial-equation deps.
- Warning-level and non-blocking: never affects simulatability, never pins the equation editor open.
- Missing-connector only flags when the referenced variable is actually drawable on the view; modules are exempt as targets (input wiring lives in the references table) and as connector sources; aliases resolve one hop; stocks are checked against their initial equation.
- A freshly drawn connector warns until the equation references its source -- that is out-of-sync by definition, and the dot doubles as a "finish wiring this" nudge.
- The annotation is display-only; new tests pin that serialization never emits it, so saved projects are untouched.
- Known inefficiency (recompute on viewport-only settles) documented as tech-debt item 67 with fix options.

## 5. Live flow updates while dragging a cloud (524a4fe8)

Dragging an existing flow's cloud endpoint moved the valve live but left the line/arrowhead stale until release, which then snapped. The preview used `UpdateFlow` (valve-centric) while the commit used `UpdateCloudAndFlow`. Fixed by unifying on the flow-creation codepath, per the reported expectation: `growInCreationFlow` is generalized into `growEndpointDrag`, and the canvas previews existing-endpoint drags through it -- preview and commit now share geometry, so release is snap-free by construction.

**Design notes**: `UpdateFlow`'s valve/Z-routing math is deliberately not unified (it still serves stock-endpoint and stock-move drags, and past attempts to merge the two broke routing). Dragging a *stock-attached* endpoint retains the old stale-until-release preview -- scoped out and filed as #830.

## Verification

Every commit passed the full pre-commit gamut (Rust fmt/clippy/tests, TS lint/tsc/build/tests, pysimlin). Diagram suite grew from 1322 to 1376 tests; each fix's new tests were verified to fail against the pre-fix code. Browser verification of the caret fix was done against the real KaTeX DOM pipeline; the dev-server API returns 401 to the automation browser session, so the remaining fixes rely on the gesture-suite tests driving real pointer/keyboard events through the full component tree -- worth a quick manual pass on /bpowers/population (fraction clicks, Cmd+Enter, double-click rename, connector warnings, cloud drags).

Generated with [Claude Code](https://claude.com/claude-code)
Session: https://claude.ai/code/session_01591tEkjitjUfd5wrtSmr6u


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

Issue #833, opened 2026-07-03
diagram: applySimSpecChange full-payload read-modify-write can silently revert a concurrent sim-specs commit (lost update)


bpowers:
## Problem

`applySimSpecChange` in `src/diagram/Editor.tsx` (currently around lines 1166-1190) performs a full-payload read-modify-write on sim specs: it reads the CURRENT `project.simSpecs` from the controller snapshot and builds a complete `setSimSpecs` payload (all six fields) with the one updated field substituted.

`ProjectController.applyPatch` calls are not queued/serialized against each other. If a second sim-specs commit fires while the first one's engine round-trip + snapshot refresh is still in flight, the second payload is built from a stale snapshot: it carries the first field's OLD value and silently reverts it (classic lost update).

## Why it matters

Correctness: a user edit can be silently discarded with no error and no visible indication, leaving the model with a stale sim-spec value (e.g. reverted `dt` or stop time).

## Severity / reachability

Currently only reachable in a sub-second window (WASM `applyPatch` + serialize round-trip) that a human tabbing between fields realistically cannot hit -- which is why this is tracked rather than hotfixed. The window would widen if patch application ever becomes slower or genuinely async (e.g. worker offload), so it should not be left to rot.

## History

- This predates the issue-#55 commit-on-settle change: the old per-keystroke handlers had the identical read-modify-write.
- The #55 work (commit c011e15e) made the race strictly less likely (far fewer patches) but did not fix it; that commit message notes the race is "now tracked separately" -- this issue is that tracking.

## Component

`src/diagram` -- `Editor.tsx` (`applySimSpecChange`), `ProjectController.applyPatch`.

## Possible approaches

1. Send a partial `setSimSpecs` patch containing only the changed field, so concurrent payloads compose instead of overwriting each other.
2. Serialize `applyPatch` calls in `ProjectController` (queue them so each read-modify-write observes the previous patch's result).

## Context

Identified during the interactive-editing burndown (branch `interactive-editing-burndown`, PR forthcoming), flagged by the adversarial reviewer of the #55 change and confirmed pre-existing.

Issue #733, opened 2026-06-08
epic: backlog burndown -- tracking all open work


bpowers:
## Purpose

A single, pinned index of **all known open work**, so anyone (human or agent) can find a bite-sized thing to pick up without re-surveying the whole tracker. This is the backlog-burndown drive: chip away, tick boxes, keep the count falling.

This is a **master index**. Two large clusters have their own dedicated trackers and are *not* re-listed here:

- **Loops that Matter (LTM)** -- 35 open issues curated in #488 (pinned). Filter: `gh issue list --label ltm --state open`.
- **Dormant feature epics** -- big-rock features that are designs unto themselves: #25 (array support), #35 (unit checking), #46 (module editing), #57 (fork & merge models).

Everything else is grouped below by code area. Each row links a standalone issue.

## How to use this tracking issue

- **Looking for something to pick up?** Browse a section that matches your interest and grab an unchecked box. `bug`/`hygiene` items tend to be self-contained; `enhancement` items are larger.
- **Filtering on the command line:** every issue now carries area labels -- `gh issue list --label engine --state open`, `--label frontend`, `--label backend`, `--label security`, `--label layout`, `--label ci`, `--label docs`, `--label mcp`, `--label compat`, `--label hygiene`.
- **Closed something?** Tick the box here in the same PR (or a fast-follow). The box is the human-readable progress signal; the underlying issue state is the source of truth.
- **Filed something new?** Add a `- [ ] #N -- short title` row to the right section (or to #488 if it is LTM).

## Recently triaged (2026-07-28)

- Full sweep of every open issue against this index. **71 untracked issues were added.** The largest new cluster is the Stella conveyor/queue engine (16 issues), which now has its own section; the second largest is build/lint tooling.
- Ticked 11 rows whose issues had been closed since the last sweep: #287, #568, #585, #635, #642, #688, #694, #695, #707, #708, #709.
- The old "Deploy / ops / CI / build / test hygiene / CLI" bucket had grown to ~40 rows, which defeats the browse-and-pick purpose. Split into **Build / lint / type-check tooling**, **CI / test hygiene**, and **Deploy / ops / release / CLI**.
- LTM-labeled issues #782, #977 and #984 were routed to #488 per the convention below (and #754/#776 ticked there), so they are deliberately absent here.
- #78, #90 and #91 predate this epic and were missed by the original index; they are now filed under the matching sections.

## Recently triaged (2026-06-08)

[...]

Issue #110, opened 2025-08-26
Flows are invisible on creation, if created left or up


filipesmedeiros:
Hey!

If a person creates a flow, but initially moves it up or left, the flow will be invisible and jittery. I haven't been able to fix this issue after trying for a few hours, but I do have some pointers for you, @bpowers :)

First, here is a video of the problem: 

https://github.com/user-attachments/assets/1946a5c0-c0ea-46e2-9a0a-21907ff25e80

Second, here are the problematic lines I have found.

Inside this loop:

https://github.com/bpowers/simlin/blob/526b3b992bbd9fe8a2322d4318d178755bcbba61/src/diagram/drawing/Flow.tsx#L508-L531

The value of `θ` is always *0 or 90*, which leads me to believe it is never 270 or 180 (left and up).

This seems to be because the points returned in

https://github.com/bpowers/simlin/blob/526b3b992bbd9fe8a2322d4318d178755bcbba61/src/diagram/drawing/Flow.tsx#L482

are always greater than or equal to the original drag. In other words, if you start a flow at point (10, 10), even if you drag left, the second point will be (10, 10).

I don't know why this is the case, but hopefully my contribution will be enough to fix the problem :)

bpowers:
Verified fixed at HEAD (interactive-editing burndown pass, 2026-07-02).

The degenerate-axis recovery in adjustFlows (src/diagram/drawing/Flow.tsx) now classifies the in-creation flow's axis by the larger-magnitude component of the source->target delta (Math.abs comparison), which handles the negative direction on both axes; regression tests cover LEFTWARD and UPWARD creation drags in src/diagram/tests/flow-attach.test.ts ('the sink follows a LEFTWARD/UPWARD cursor (not degenerate)') and src/diagram/tests/flow-routing.test.ts (UpdateCloudAndFlow degenerate flow creation, all four directions).

Also browser-verified against the dev server: creating a flow and dragging left renders the live preview (line, valve, arrowhead) tracking the cursor leftward; same for upward. Both commit correctly. No invisible/jittery behavior remains.

PR #834, opened 2026-07-03
diagram: interactive-editing burndown (Cancel, cloud drags, offscreen open, sim specs, selection state)

### LINKED ISSUE #52: if diagram is completely(ish) offscreen on current viewport (on didMount) center it
this is complicated a little bit by react's abstractions, but should be possible

### LINKED ISSUE #53: when moving a cloud, the flow label jumps around
this is mostly cosmetic, but it bugs me

### LINKED ISSUE #55: every character change in sim specs shouldn't result in a model save
There is the UI aspect to consider: should we have a "cancel/save" set of buttons, or automatically sync after editing has stopped for a few seconds (or user clicks out of sidebar)

### LINKED ISSUE #86: 'Cancel' not prevent equation from being edited
I was fiddling with the logistics model and inspected one of the Auxiliary variables.  I clicked into the equation field of the inspector and typed random characters (I was wondering if it was editable).  I then clicked "Cancel" next to the save button, but my changes were already saved and the model had errors.  

EXPECTED BEHAVIOR:

I would have expected "Cancel" to have prevented my changes from being saved.  "Save" button is right next to it, which I expect would save my edits.

### LINKED ISSUE #529: diagram: selection-mutating Editor handlers bypass handleSelection, skipping showDetails reset and onSelectionChanged
## Summary

`src/diagram/Editor.tsx` has a single canonical selection-change handler, `handleSelection(selection: ReadonlySet<UID>)` (around line 699), which does three things:

1. `setState({ selection, flowStillBeingCreated: false, variableDetailsActiveTab: 0 })`
2. When the new selection is **empty**, also `setState({ showDetails: undefined })` -- resets the variable-details panel state.
3. Notifies the optional `onSelectionChanged?: (idents: string[]) => void` prop via a deferred `setTimeout(..., 0)` that reads `getSelectionIdents()` after React commits the new selection.

Several other handlers mutate `state.selection` directly via `setState`/`setStateForView` instead of routing through `handleSelection`, so they skip steps 2 and/or 3:

- **`handleSelectionDelete`** (around line 747, clears at ~line 816-818): `this.setState({ selection: new Set<number>() })`. Leaves `showDetails` at whatever it was (e.g. `'variable'`) even though there is no selection, and does **not** notify `onSelectionChanged` that the selection was cleared by the delete.
- **`handleFlowAttach`** (around line 835, clears at ~line 1369): sets `selection: new Set<number>()` -- no `showDetails` reset, no `onSelectionChanged`.
- **`handleSelectionMove`** (around line 1373) and the undo/redo path (around line 1814) -- set `showDetails: undefined` but do not notify `onSelectionChanged`.
- **`handleNavigateBack`** / **`handleNavigateToLevel`** (around lines 1823 / 1848) -- restore the parent's selection and *do* set `showDetails: undefined`, but do not notify `onSelectionChanged`.

## Why it matters

- **Stale UI state**: After `handleSelectionDelete`, `state.showDetails` can remain `'variable'` with an empty selection -- an inconsistent state. Currently masked because `getVariableDetails()` also checks `getNamedSelectedElement()` (returns undefined for an empty selection), so the panel doesn't render -- but the state is wrong and a future change to that render guard would surface it.
- **Broken host contract**: `onSelectionChanged` is documented (see `src/diagram/CLAUDE.md`) as firing "after each selection change" and is consumed by `simlin-serve`'s `EditorHost` to forward selection state to backend listeners. Hosts that subscribe are **not** told the selection was cleared when the user deletes the selected element(s), or when navigating in/out of modules. `src/app`'s `HostedWebEditor` does not subscribe, so the web app is unaffected, but the contract is violated for serve and any other subscriber.

## Component(s) affected

- `src/diagram` -- `Editor.tsx` (`handleSelection`, `handleSelectionDelete`, `handleFlowAttach`, `handleSelectionMove`, `handleNavigateBack`, `handleNavigateToLevel`, undo/redo handler)
- Downstream: `src/simlin-serve` `EditorHost` (subscriber to `onSelectionChanged`)

## Possible approach

Route all `state.selection` mutations through a single helper (or through `handleSelection` itself) so the `showDetails`-reset-on-empty and the deferred `onSelectionChanged` notification happen consistently. Handlers that also need to set view state can compose the selection helper with their own `setState`. At minimum, `handleSelectionDelete` should clear `showDetails` and notify `onSelectionChanged` when it empties the selection.

Care needed around the existing `selectionDeferralTimer` bookkeeping (instance field cancelled in `componentWillUnmount`, superseded by newer selections) so consolidation doesn't introduce duplicate or lost notifications.

## How discovered

Noticed while fixing an unrelated bug in `src/diagram/Editor.tsx` -- pre-existing inconsistency between `handleSelectionDelete` and `handleSelection`.

### LINKED ISSUE #832: diagram: adjustFlows valve-mirroring is non-idempotent for off-center valves on the min-coordinate endpoint
## Summary

The 2-point branch of `adjustFlows` in `src/diagram/drawing/Flow.tsx` (the block commented "Re-place the valve by mirroring it around the segment", ~lines 697-722) repositions an off-center valve by **mirroring** it across the segment midpoint rather than **preserving its fractional position** along the segment. This makes the operation non-idempotent: when the dragged endpoint is the *smaller-coordinate* end of the flow, an off-center valve jumps to its mirror position on the very first drag frame -- even for a zero or sub-pixel move.

## Concrete repro

For a horizontal flow with:
- source cloud at `x = 100`
- stock sink edge at `x = 277.5`
- valve at `x = 200`

Calling `growEndpointDrag(flow, isSource=true, {x:0, y:0})` (or any tiny delta) returns the valve at `x = 177.5` instead of `200` -- a ~22.5px jump the instant the user grabs the source cloud.

The mirror case (sink cloud at the *larger* coordinate) is idempotent and correctly stays at `200`, so the defect is specific to the moved-endpoint-is-min-coordinate geometry (a SOURCE cloud on the left of a horizontal flow, or the TOP of a vertical flow).

## Root cause

The formula computes `base = min(otherEnd, movedEnd)` then reconstructs the valve as `base + |fraction * d|`. When `base` lands on the moved (cloud) side rather than the anchor (`otherEnd`) side, the absolute-offset reconstruction reflects the valve across the segment midpoint instead of preserving its fractional position.

The mirroring was **intentionally** added so an off-center valve stays between the two ends when a drag *flips* the segment orientation (the moved end crossing past `otherEnd`; the guard/comment references the #818-adjacent NaN case). But it is wrong for the common **non-flipping** case: it should preserve the valve's fraction along the segment, not reflect it.

## Impact

User-visible valve/label jump at the *start* of dragging a source cloud (or a top-anchored vertical flow's endpoint). Minor but real. Correctness/UX polish, not a crash.

## Component

`src/diagram/drawing/Flow.tsx` -- `adjustFlows` 2-point branch (shared by `growEndpointDrag` / `growInCreationFlow` in `flow-attach.ts`).

## Suggested fix direction

Make the 2-point valve reposition **preserve the fractional position along the segment** (like `preserveValveFraction` does for multi-segment flows) rather than mirroring, while:
- keeping the divide-by-zero guard from #818 (`origStock.x - otherEnd.x || 1`), and
- keeping the flipped-orientation case (moved end crossing past `otherEnd`) working.

Warrants its own change with dedicated tests, since `adjustFlows` also carries the along-axis and #818 NaN-guard behavior.

## How discovered

Identified while fixing #53 (flow valve/label teleports when dragging a cloud perpendicular to form an L). This is a **separate, pre-existing** bug in the same file, out of scope for #53: #53's scope is the perpendicular-reroute teleport (a different code path), and touching the shared `adjustFlows` valve formula risks the along-axis and #818 NaN-guard behavior, warranting its own change.

bpowers:
Seven fixes for users interactively creating and editing models, picked from the #733 backlog index plus three issues discovered (and fixed) along the way. Each fix was implemented by a dedicated agent, adversarially reviewed by an independent agent, and iterated until the review had no material findings; each landed as its own commit.

Fixes #86
Fixes #53
Fixes #52
Fixes #55
Fixes #529
Fixes #832

## What changed

- **Cancel/Escape discard equation edits (#86, `59c780fd`).** All three details-panel editors save on blur, and clicking Cancel blurred first — so Cancel *saved* the edit it was discarding. Blur now commits only when focus leaves the panel; Cancel/Save/Delete preventDefault on pointer-down; Escape discards on all three fields; the discard resets the live Slate documents and clears their undo history so Cmd+Z can't resurrect the abandoned text (found by the adversarial review). The load-bearing canvas-blur-commit is regression-tested. Browser-verified on the real macOS mouse path.
- **Flow valve/label stability during cloud drags (#53 `b8031d52`, #832 `91718548`, follow-up `d4199189`).** Browser-reproduced: dragging a cloud perpendicular so the flow bends into an L teleported the valve+label onto the stock corner (~100px, every frame) and committed it there. Root cause was closest-segment clamping against the fixed pre-drag valve; the valve now rides the segment the drag is growing. The review of that fix surfaced two more defects in the same code, both fixed here rather than deferred: grabbing a *source* cloud reflected an off-center valve ~22px on the first frame (#832, the old mirror math was only idempotent for sink drags — replaced with signed fraction preservation), and a diagonal drag that traveled along-axis before bending snapped the cloud back ~100-165px at the bend (the reroute discarded accumulated parallel travel — it now applies the full delta). Continuity is pinned by step-sweep tests (same-topology steps <12px, topology-change hop <15px, measured 8px).
- **Center the diagram when a saved view is offscreen (#52, `a780ee96`).** Once per mount, after the first real svg measurement, idle-only, never in embedded mode; commits through the no-undo viewport path. The adversarial review caught that the first-draft visibility metric would have yanked every large/zoomed-in model to center on open (fraction-of-bbox can never reach the threshold when the bbox dwarfs the viewport); the metric now normalizes by the smaller of bbox and viewport area, so a viewport full of diagram reads fully visible.
- **Selection-emptying edits no longer strand the details panel (#529, `370200b5`).** Most of #529 was already fixed by the React refactor (`onSelectionChanged` fires from a post-commit effect for every path). The residue: delete/create/flow-attach/link-attach (and a no-match search) left `showDetails: 'variable'` over an empty selection. They now route through one pure helper. Decision: the model-level errors panel is deliberately exempt — deleting a variable mid-error-triage keeps the error list open; empty-canvas click/Escape keep their broader dismiss-everything behavior.
- **Sim-specs edits commit on settle (#55, `c011e15e`).** Every keystroke in start/stop/dt/time-units fired an engine patch, an undo entry (typing "1900" evicted most of the 5-deep undo buffer), and a save — and clearing a field patched `Number('') === 0` into the model. Fields now hold a local draft and commit once on blur/Enter through a pure resolver that rejects empty/non-finite values and non-positive dt; Escape reverts the field without dismissing the drawer. Browser-verified.

## Issues handled without code

- **#110** (flows invisible when created dragging left/up): verified already fixed at HEAD by the recent flow-tool work — regression tests exist for both axes, and I browser-verified live creation drags in all directions. Closed with evidence.
- **#833** (filed): `applySimSpecChange`'s full-payload read-modify-write can lose a concurrent sim-specs commit. Pre-existing, sub-second window, not human-reachable today; tracked instead of hotfixed.

## Verification

Every fix has fail-then-pass unit coverage (+~1,050 net test lines; the diagram suite grew from 1,395 to 1,475 tests). #86, #53, #110, and #55 were additionally verified in a real browser against the dev stack; #52's mount path is covered by integration tests with real resize sequencing plus the reviewer's code-trace (synthetic pointer events can't drive the canvas viewport, so no browser E2E for that one). Pre-commit hooks (Rust + TS + Python) passed for every commit.

One known cosmetic behavior deliberately left alone: the perpendicular catch-up hop when a straight flow first bends (the flow is axis-pinned below the bend threshold, so the arrowhead catches up at the bend). It predates this work, is strictly smaller than before, and smoothing it is a reroute-threshold redesign rather than a bug fix.

### DIFF
--- a/src/diagram/Editor.tsx
+++ b/src/diagram/Editor.tsx
@@ -48,6 +48,7 @@ import { FlowIcon } from './FlowIcon';
 import { LinkIcon } from './LinkIcon';
 import { ModuleIcon } from './ModuleIcon';
 import { ModelPropertiesDrawer } from './ModelPropertiesDrawer';
+import type { SimSpecField } from './sim-spec-draft';
 import { renderSvgToString } from './render-common';
 import { Status } from './Status';
 import { StockIcon } from './StockIcon';
@@ -1197,24 +1198,26 @@ export const Editor = React.memo(function Editor(props: EditorProps): React.Reac
     await applyPatchAndRefresh(patch, 'sim specs');
   };
 
-  const handleStartTimeChange = React.useCallback(async (event: React.ChangeEvent<HTMLInputElement>): Promise<void> => {
-    const value = Number(event.target.value);
-    await applySimSpecChange({ startTime: value });
-  }, []);
-
-  const handleStopTimeChange = React.useCallback(async (event: React.ChangeEvent<HTMLInputElement>): Promise<void> => {
-    const value = Number(event.target.value);
-    await applySimSpecChange({ endTime: value });
-  }, []);
-
-  const handleDtChange = React.useCallback(async (event: React.ChangeEvent<HTMLInputElement>): Promise<void> => {
-    const value = Number(event.target.value);
-    await applySimSpecChange({ dt: `${value}` });
-  }, []);
-
-  const handleTimeUnitsChange = React.useCallback(async (event: React.ChangeEvent<HTMLInputElement>): Promise<void> => {
-    const value = event.target.value;
-    await applySimSpecChange({ timeUnits: value });
+  // The drawer holds a draft while a sim-specs field is focused and calls this
+  // exactly once when the field settles (blur/Enter) with a validated, changed
+  // value -- so a single edit is one engine patch and one undo entry, instead
+  // of one per keystroke (issue #55). The drawer already rejected garbage/empty
+  // input, so we only route the field to its `applySimSpecChange` update.
+  const handleSimSpecCommit = React.useCallback((field: SimSpecField, value: number | string): void => {
+    switch (field) {
+      case 'startTime':
+        void applySimSpecChange({ startTime: value as number });
+        break;
+      case 'stopTime':
+        void applySimSpecChange({ endTime: value as number });
+        break;
+      case 'dt':
+        void applySimSpecChange({ dt: `${value as number}` });
+        break;
+      case 'timeUnits':
+        void applySimSpecChange({ timeUnits: value as string });
+        break;
+    }
   }, []);
 
   const handleDownloadXmile = React.useCallback(async (): Promise<void> => {
@@ -1276,10 +1279,7 @@ export const Editor = React.memo(function Editor(props: EditorProps): React.Reac
         stopTime={simSpec.stop}
         dt={dt}
         timeUnits={simSpec.timeUnits || ''}
-        onStartTimeChange={handleStartTimeChange}
-        onStopTimeChange={handleStopTimeChange}
-        onDtChange={handleDtChange}
-        onTimeUnitsChange={handleTimeUnitsChange}
+        onSimSpecCommit={handleSimSpecCommit}
         onDownloadXmile={handleDownloadXmile}
         onDelete={onDelete}
       />
--- a/src/diagram/Editor.tsx
+++ b/src/diagram/Editor.tsx
@@ -167,6 +167,52 @@ interface EditorState {
   variableDetailsActiveTab: number;
 }
 
+// Enforces the Editor's selection invariant for the element-mutation paths
+// (delete, create, flow/link attach): when such a path leaves the selection
+// EMPTY, the selection-tied variable-details panel must close and the
+// variable-details tab must reset. Every one of those setState sites composes
+// this (spread into its patch) so the rule holds uniformly instead of each call
+// site re-deriving it and drifting apart.
+//
+// The errors panel is EXEMPT. showDetails is a single field that is either
+// 'variable' (a panel bound to the selected variable), 'errors' (the
+// model-level error list, independent of any selection), or undefined. Only the
+// 'variable' case is tied to the selection, so an emptied selection clears
+// showDetails ONLY when it is currently 'variable' -- deleting a variable while
+// triaging the error list must not dismiss that list. currentShowDetails is
+// therefore an input.
+//
+// A non-empty selection is returned unchanged: the caller keeps ownership of
+// showDetails for the paths that intentionally open a panel. handleSelection's
+// empty-click/Escape dismiss and the navigation handlers deliberately close
+// EVERYTHING (variable AND errors), so they do NOT route through here.
+//
+// The bug this closes: the delete/create/attach handlers mutated `selection`
+// directly and skipped the variable-panel reset, leaving showDetails at
+// 'variable' with nothing selected -- masked only because getDetails() also
+// guards on a named selected element, but fragile (a later single-select would
+// pop the stale panel open, and any change to that render guard would surface a
+// broken panel over an empty selection).
+export function selectionStatePatch(
+  selection: ReadonlySet<UID>,
+  currentShowDetails: 'variable' | 'errors' | undefined,
+): {
+  selection: ReadonlySet<UID>;
+  showDetails?: 'variable' | 'errors' | undefined;
+  variableDetailsActiveTab?: number;
+} {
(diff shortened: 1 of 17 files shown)


bpowers:
@codex review

Issue #841, opened 2026-07-05
vdf: overlap-path descriptor peel likely drops a real stock on SimService/Base.vdf (c Identified Oil Reserve)


bpowers:
## Problem

On `third_party/uib_sd/spring_2008/SimService_BUENO/SimService/Model Files/Base.vdf` (and its `ctxt0001` sibling), the OVERLAPPING-descriptor peel path in `src/simlin-engine/src/vdf/record_results.rs::identify_descriptor_records` (mirrored in `tools/vdf_xray.py`) classifies the record for **`c Identified Oil Reserve`** as a graphical-function descriptor and drops its series.

Details:
- Its span is OT[17..19): a 2-element stock-coded span with real data blocks, f11=17 with n_lookups=18.
- The span overlaps `AGE SPECIFIC FERTILITY DISTRIBUTION FUNCTION` (OT[18..100), an 82-wide stock-coded span).
- The lexical lookup-def-name test is ambiguous for the pair, so the highest-f10 fallback decides -- and it picks `c Identified Oil Reserve` as the descriptor.

This is very likely a false positive: the file's other `c Identified * Reserve` siblings (Coal / Gas / tar sands) are all real 2-element stocks that are correctly kept.

## Why it matters

Correctness: a real stock's simulation series is silently dropped from decoded results, which corrupts any comparison or import based on this file.

## Components affected

- `src/simlin-engine/src/vdf/record_results.rs` (`identify_descriptor_records`, the highest-f10 fallback)
- `tools/vdf_xray.py` (same function, kept in parity)
- `docs/design/vdf.md` -- "Name-to-OT mapping" step 2 and "Appendix: the owner/descriptor discriminator"

## Context

Identified while hardening the STANDALONE lookup-only descriptor drop (GH #840 follow-up work on branch `vdf-parity-fixes`). The behavior is pre-existing: identical before and after that branch's changes. The standalone-drop hardening in that branch fixed the four STANDALONE false positives (Agriculture Employment, Alaska Oil Discovered/Undiscovered, Atmos UOcean Temp) but deliberately did not touch the overlap path.

## Possible approaches

- Strengthen the overlap-path discriminator so a span with real data blocks and stock-coded siblings (e.g. matching `c Identified * Reserve` family members that are kept) is not peeled as a descriptor.
- Prefer structural evidence (data-block presence, span width vs n_lookups consistency) over the highest-f10 fallback when the lexical lookup-def-name test is ambiguous.

[...]

PR #843, opened 2026-07-05
vdf: fix decoders, add Rust/Python parity harness, expose load_vdf in pysimlin

### LINKED ISSUE #839: engine: VDF record-key map off-by-one on stale name-table entries mislabels result series (risk2.vdf)
## Problem

The Rust VDF parser mislabels result columns on files whose section-2 name table contains stale/deleted entries. Confirmed on `test/bobby/vdf/econ/risk2.vdf` (and `risk.vdf`).

`VdfFile::record_name_key_to_name_index` in `src/simlin-engine/src/vdf.rs` (around line 1468) re-walks the section-2 name table to build the record-f[2]-key -> name-index map. It increments `name_idx` for EVERY u16-length-prefixed entry, including stale/deleted entries with non-printable payloads. But `parse_name_table_extended` (which builds `VdfFile::names`) SKIPS non-printable stale entries. Two consequences:

1. After each stale entry, the key map is off by one relative to the `names` vec -- every subsequent record resolves to the wrong name.
2. The loop condition `name_idx < names.len()` terminates early, dropping the trailing names' keys entirely (some records resolve to `None` and their series are silently dropped).

## Empirical verification (2026-07-05)

Both algorithms were replicated in Python against the fixtures:

- `risk2.vdf` has 3 stale entries; 60 of its section-1 records resolve f[2] to the WRONG name (e.g. rec[45] should be "perceived inflation rate" but Rust resolves "inflation elasticity of risky behavior"; each subsequent name is shifted by one). `to_results_via_records` emits ~60 mislabeled series.
- `risk.vdf` has 1 stale entry; 3 records affected, including one whose key resolves to `None`, dropping that series.
- The Python twin `tools/vdf_xray.py::build_record_name_key_to_name_index` is CORRECT: it reuses `_parse_name_table_entries`, which returns only printable entries, each with its true string offset.

## Why it matters

Correctness: `to_results_via_records` silently emits series under the wrong variable names, so any comparison against Vensim reference output using these fixtures is comparing the wrong columns. Mislabeled data is worse than missing data.

## Components affected

- `src/simlin-engine/src/vdf.rs` -- `record_name_key_to_name_index` (~line 1468)
- `src/simlin-engine/src/vdf/section5_dims.rs` -- `build_name_key_to_name` duplicates the same buggy walk a third time and needs the same fix

## Fix direction

Refactor the Rust side to parse name-table entries once into `(name, string_offset)` pairs (mirroring Python's `_parse_name_table_entries`) and derive both `names` and the key map from that single pass, eliminating the duplicated walks.

## Test gap

Existing tests miss this because `test_record_name_key_maps_to_section2_string_start` only uses fixtures without stale entries, and the `to_results` tests on the econ siblings only assert non-emptiness. New tests should assert correct name resolution on `risk.vdf`/`risk2.vdf` (fixtures with stale entries).

## Context

Identified 2026-07-05 during VDF parser investigation by cross-checking the Rust implementation against `tools/vdf_xray.py`.

### LINKED ISSUE #840: vdf: Rust and Python decoders have drifted with no differential parity test
## Problem

The two VDF decoders -- Rust (`src/simlin-engine/src/vdf.rs` and submodules) and Python (`tools/vdf_xray.py`) -- have drifted apart, and nothing keeps them aligned: `tools/test_vdf_xray.py` has 71 tests but none compare the Python extraction against the Rust extraction. The xray tool is nominally the reference/oracle for the Rust parser, so silent drift undermines its purpose.

Specific drift found in a 2026-07-05 audit:

1. **Slot-table location: xray still uses the deprecated heuristic.** `tools/vdf_xray.py` locates the slot table with the backward-scan heuristic `find_slot_table` (its docstring claims "we have not yet identified that field", which is stale) even though the deterministic decode -- section-1 field1 word pointer + `block1[7]` count + `0x00430000` terminator -- was identified, documented in `docs/design/vdf.md`, implemented in Rust (`vdf::slot_table_from_header`), and pinned corpus-wide by `tests/integration/vdf_structural_invariants.rs::test_slot_count_matches_block1_word7_exactly`. The xray's `--validate` even warns when field1 doesn't match its own heuristic table.

2. **Ghost stock-slot series for bare graphical functions.** Python `identify_descriptor_records` lacks the standalone lookup-only descriptor drop that the Rust side has (`src/simlin-engine/src/vdf/record_results.rs`, `standalone_lookup_only_descriptors`), so `--extract` still emits ghost stock-slot series for bare graphical functions on Ref.vdf-class files.

3. **Sensitivity/optimization magic `7f f7 17 53` handling disagrees.** Python's `parse_vdf` accepts it; Rust `VdfFile::parse` rejects it, even though `docs/design/vdf.md` says such files parse with the ordinary rules up to the `header[0x68]` payload. (Adjacent to tech-debt entry 33, which tracks decoding the extra tail payload -- this item is about the two parsers disagreeing on whether the file parses at all.)

4. **~1500 lines of superseded legacy reconstruction code in vdf_xray.py.** The owner-block / slot-alignment-scoring / DP non-overlap-selection / gap-filling-assignment stack (roughly `analyze_slot_table_offsets` through `map_names_to_owner_blocks`) was superseded by the direct record-map path (`decoded_record_spans` + `identify_descriptor_records` + `extract_named_results_with_diagnostics`) and is now used only by exploratory display flags. It should be deleted or quarantined.

5. **Stale doc references in committed code.** Four call sites cite a `docs/design/vdf.md` "appendix" ("Claims about the owner/descriptor discriminator") that does not exist in the doc:
   - `src/simlin-engine/src/vdf/record_results.rs:243`
   - `tools/vdf_xray.py:2991` and `:3220`
   - `tools/test_vdf_xray.py:1480`

   Three cite a deleted temp file `/tmp/vdf_ref_dims.md` whose evidence should be folded into `docs/design/vdf.md`:
   - `src/simlin-engine/src/vdf/section5_dims.rs:24`
   - `tests/integration/vdf_multidim.rs:703` and `:791`

## Why it matters

- **Broken oracle.** The corpus-precision workflow treats xray as the reference for the Rust parser; when the two silently diverge (items 1-3), disagreements stop meaning "one of them has a bug" and start meaning "who knows".
- **Wrong extractions.** Item 2 produces concretely wrong `--extract` output (ghost series) on real files today.
- **Maintainability.** Items 4-5 make the tool harder to modify safely and send readers chasing documentation that does not exist.
- **No regression guard.** Without a differential/parity test (Python extraction vs Rust extraction over the corpus), every future fix on one side re-opens the drift.

## Components affected

- `tools/vdf_xray.py`, `tools/test_vdf_xray.py`
- `src/simlin-engine/src/vdf/` (`record_results.rs`, `section5_dims.rs`)
- `tests/integration/vdf_structural_invariants.rs`, `tests/integration/vdf_multidim.rs`
- `docs/design/vdf.md`

## Possible approaches

1. Add a differential test that runs both extractors over the shared corpus and asserts the result sets and series values agree (a CLI shim on the Rust side, or a JSON dump compared by the Python tests).
2. Port the deterministic slot-table decode into `vdf_xray.py`, retire `find_slot_table` (or keep it only as a `--validate` cross-check), and fix its stale docstring.
3. Port `standalone_lookup_only_descriptors` into `identify_descriptor_records`.
4. Align `7f f7 17 53` handling with `docs/design/vdf.md` (accept in Rust up to the `header[0x68]` payload, or document why not).
5. Delete or quarantine the legacy reconstruction stack behind an explicitly-labeled legacy module/flag.
6. Fix the stale doc references: restore or write the owner/descriptor-discriminator appendix in `docs/design/vdf.md`, and fold the `/tmp/vdf_ref_dims.md` evidence into the doc.

## Related

- #501 -- dataset (0x41) parity gap (Python stubs what Rust implements); same drift theme, dataset-specific.
- `docs/tech-debt.md` entry 19 ("Rust VDF Parser Boundary Parity") -- the earlier drift direction (Rust adopting Python boundary handling).
- `docs/tech-debt.md` entry 33 ("VDF 0x53 Result-Family Tail Is Undecoded") -- the undecoded tail payload behind item 3.
- #500 (closed) -- a previous xray-vs-Rust disagreement caused by exactly this kind of drift.

## Discovery context

Identified during a 2026-07-05 audit of the VDF decoders.

### LINKED ISSUE #842: vdf: class-0x05 blocks on their own short time grid decode as garbage (zambaqui corpus)
## Problem

Both VDF readers -- Rust (`src/simlin-engine/src/vdf.rs`, `VdfFile::extract_block_series_sampled` / `block_bitmap_layout`) and the Python inspector (`tools/vdf_xray.py`, `VdfFile._block_bitmap_layout` / `extract_block_series`) -- decode a data block's bitmap width by testing exactly two candidate widths: the saved grid (header `0x78`) and the full block grid (header `0x7C`).

The `third_party/uib_sd/zambaqui` corpus (both `0x52` run files and `0x53` sensitivity/optimization files) contains class-0x05 (input/data) blocks stored on their **own** short time grid, independent of both header grids. Neither candidate width's popcount matches the block's u16 count, so both readers fall back to the run-grid bitmap and silently decode garbage.

Canonical example: `gdp deflator` in `third_party/uib_sd/zambaqui/opt-1.vdf`, OT[708], block at `0x5bc38`:
- u16 count = 26 values on a 71-point run
- its bitmap appears to be ceil(26/8) = 4 bytes with all 26 bits set (`ffffff03`), followed by plausible f32 data (first value ~0.62, final value 2.08698 matching the section-6 final-values array)
- both candidate widths give popcount 50 != 26, so the readers fall back to the 9-byte run-grid bitmap and decode a garbage series (starting -3.77e-29, ending 0.0), with no error

## Scope

- At least 23 zambaqui files affected (0x52 and 0x53), with 1-14 owner-coded OT slots each: `old runs/Current*.vdf` and `old runs/base-*.vdf` have 6-14 bad class-0x05 slots; most others have exactly 1 (the `gdp deflator`). Verified by comparing extracted last values against the section-6 final-values array.
- No file under `test/` is affected; the parity/oracle suites there pass with zero exceptions.
- The two readers agree with each other (the differential parity harness in `src/simlin-engine/tests/integration/vdf_parity.rs` holds), so this is a shared decode gap, not reader drift.

## Current mitigation

`src/simlin-engine/tests/integration/vdf_sensitivity.rs` structurally skips such blocks via a popcount gate (`block_bitmap_is_decodable`), bounded to <= 2 skips per file, and documents the limitation. `docs/design/vdf.md` ("Sensitivity / optimization format") mentions it.

## Why it matters

Correctness: affected input/data series decode silently as garbage rather than failing loudly, which corrupts any comparison or import built on those files.

## Components affected

- `src/simlin-engine/src/vdf.rs` (`extract_block_series_sampled`, `block_bitmap_layout`)
- `tools/vdf_xray.py` (`_block_bitmap_layout`, `extract_block_series`)
- `docs/design/vdf.md`

## Possible approach

Extend the per-block bitmap-width discriminator with a third candidate: ceil(count/8), i.e. an own-grid block whose bitmap covers exactly its count of points. Also needs a story for mapping those points onto absolute times -- there may be a paired time block, or the block may carry its own grid origin. The "old runs" files with 6-14 bad slots suggest several distinct own-grid layouts, so this needs format investigation before implementation.

## Context

Identified while building the VDF differential parity harness and 0x53 sensitivity coverage (follow-on to #840; related tech-debt entries 19 and 33 in docs/tech-debt.md cover the two-candidate discriminator's origin and the undecoded 0x53 tail, respectively -- neither covers this own-grid gap).

bpowers:
Fixes #839
Fixes #840
Fixes #842

This branch closes out the VDF-decoder audit: two correctness bugs in the Rust reader, the Rust/Python drift cluster, a newly-discovered false positive in the descriptor gates, a new format discovery (the data-grid header word), and a pysimlin import surface. Six commits, each independently reviewed.

## Why

The two VDF decoders (simlin-engine and tools/vdf_xray.py) decode the same undocumented format but had no test keeping them aligned, and each had drifted ahead of the other in different places. Every bug below would have been caught by the differential parity harness this branch adds, which is the structural fix; the individual bugs are the symptoms.

## What changed

- **#839**: the section-2 name table was walked three times by code that disagreed about stale/deleted entries, mislabeling 60 result series on econ/risk2.vdf. The table is now parsed once into (name, offset) entries that all consumers derive from.
- **Decode gaps (tech-debt 19)**: per-block bitmap-width discrimination (saved-suffix runs like econ/risk.vdf mix widths; previously two blocks per file decoded to garbage) and raw-0/class-0x11 missing-data slots (455 on Ref.vdf, previously constant-0.0 series, now NaN). The C-LEARN comparator skips reference-missing columns and pins their count at exactly 434 so a future decode regression trips the pin instead of hiding in the skip.
- **#840**: vdf_xray.py adopts the deterministic slot-table decode, gains the standalone lookup-only descriptor drop, and loses its superseded reconstruction stack (~1,100 lines); stale references to a nonexistent vdf.md appendix and a deleted /tmp analysis file are fixed by writing the real appendix. A differential parity test extracts every run-file VDF under test/ (plus the zambaqui corpus) through both readers and compares name sets and full series bit-for-bit (~5s). Both parsers now accept the 0x53 sensitivity-run magic, validated against the section-6 final-values oracle.
- **Descriptor-gate hardening** (found by review during this work): the standalone drop false-positived on real scalar stocks whenever a model's lookup count exceeded its stocks-first OT prefix (four real stock series silently dropped on third_party SimService fixtures). The gates now also require consumer corroboration (the forward link must start a distinct same-length span) and a per-file coherence veto, surfaced through diagnostics rather than silently. Extending the same gates to the overlap-path f[10] fallback was investigated and rejected with recorded evidence (#841 stays open: scalar members of shared-consumer families legitimately fail corroboration on Ref.vdf, so a fix needs a new discriminating signal).
- **#842**: header word 0x74 is the external data file's time-point count; exogenous-data blocks (class 0x05/0x06/0x0c) are the data file's sparse blocks embedded verbatim, bitmapped over that grid. All 456 such blocks across the corpora now reconcile and pass the final-value oracle (previously silent garbage in both readers, including the groupon fixtures under test/). The data grid's own time values are not recoverable from a run file, so placement onto the saved axis is a documented span/zero-order-hold approximation; blocks reconciling with no candidate now NaN-fill loudly instead of decoding garbage.
- **pysimlin**: simlin.load_vdf(path) returns a Run.results-shaped DataFrame (run, sensitivity, and dataset containers auto-detected), backed by a small orthogonal libsimlin results object. The engine's low-level readers were made total (bounds-checked to Err/NaN) because the workspace ships panic=abort, with a truncation/corruption sweep test enforcing it; parse-level validation still fails first, so this strictly converts crashes into errors.

## Non-obvious decisions

- The parity harness canonicalizes Python-emitted names in Rust with the production rule, so the mapping cannot become a drifting reimplementation; serde_json's float_roundtrip feature is a dev-dependency because the default float parser introduced 1-ULP phantom divergences.
- The data-grid bitmap candidate is deliberately tried last: over a thousand ordinary blocks coincidentally popcount-match the narrower data width, while no exogenous block matches a wider distinct width (documented with the residual risk in vdf.md).
- docs/tech-debt.md entry 19 is resolved; the duplicate entry also numbered 19 moved to 68 because entries 20-67 are cross-referenced by number elsewhere.


bpowers:
@codex review

bpowers:
@codex review

PR #844, opened 2026-07-06
vdf: resolve residual record-span overlap (SimService ghost columns)

### LINKED ISSUE #841: vdf: overlap-path descriptor peel likely drops a real stock on SimService/Base.vdf (c Identified Oil Reserve)

bpowers:
Fixes #841

## Problem

On 2007-era VDFs (`third_party/uib_sd/spring_2008/SimService_BUENO/.../Base.vdf` and its `ctxt0001` sibling), the overlap-path descriptor peel exits with 84 record spans still claiming overlapping OT slots, and the readers resolved those conflicts silently: the Rust reader by alphabetical first-claim-wins (emitting 82 ghost `AGE SPECIFIC FERTILITY DISTRIBUTION FUNCTION[k]` columns carrying other variables' data while dropping ~45 real series, including the issue's headline `c Identified Oil Reserve`), the Python xray by emitting duplicate columns for contested slots. The root cause is a file phenomenon the format doc did not describe: records for variables not saved in the run carry stale `f[11]` slot indices that land on saved variables' OT slots, and most fail the `f11 < n_lookups` descriptor-candidacy gate, so the peel could neither remove them nor detect that it had failed. The two SimService files are the only ones in the 141-file tracked corpus that exhibit residual overlap, and they sat outside the parity corpus, so the Rust/Python divergence was invisible to CI.

## Fix (two commits)

**Correctness floor (`vdf: honest-drop residual record-span overlap`).** Residual overlap components are computed as data after the peel; conflicted spans are dropped rather than silently mislabeled, surfaced on per-file diagnostics (`VdfFile::residual_overlap_diagnostics()`, xray `NamedResultsDiagnostics.residual_overlap`). The parity corpus now includes `spring_2008`, which also surfaced a pre-existing dimension-recovery divergence (Rust bailed when record-anchor and section-5 counts disagree; it now falls back to the same record-group recovery Python uses — byte-identical on Ref.vdf, no-op on every other corpus file).

**Recovery (`vdf: recover real owners from residual overlap components`).** Residual components are re-resolved instead of blanket-dropped: phase-1 peels that overlap a component are undone (a peel that failed to resolve its component was evidence-free), lexically lookupish names are peeled without the candidacy gate, and remaining conflicts are adjudicated by an ordering oracle built on Vensim's alphabetical OT allocation — bracketing by nearest uncontested owners, run-boundary detection via inverted brackets, recovered same-run anchors, iterated to a fixpoint. The oracle is gated per-file: it only runs when the file's own uncontested owners measure at least 0.95 adjacent-pair alphabetical consistency (SimService: 0.9964; corpus median 0.9869); anything inconclusive falls back to the honest drop. On both fixtures this drops the 41 stale-`f[11]` ghost records and recovers every real owner — `Population`'s 164 elements, both `c Identified Oil Reserve` series (51000 -> 8783.97, 7900 -> 2958.07), and the four contested scalars — leaving empty residual diagnostics.

## Non-obvious decisions

- **No model-guided crutch**: the fix is entirely model-free, like Vensim's own reader. The ordering invariant was previously a footnote in `docs/design/vdf.md` ("not a rule a reader needs"); it is now load-bearing but only under a measured per-file gate, never assumed universal.
- **No scalar objective can adjudicate these conflicts**: the fixtures contain both orientations (a ghost wide span over ~40 real narrow owners AND a real 164-wide `Population` under 36 internally-alphabetical ghost claimants), which is why the design uses anchor consistency rather than majority/coverage/conflict-degree rules. `Ref.vdf` contains both shapes at small scale inside components the existing peel already resolves — those paths are untouched.
- **Provable blast radius**: recovery triggers only on residual components, which exist only in the two SimService files; the other 139 corpus files are structurally unaffected (verified by corpus survey and the parity harness, which pins Rust and Python bit-identical including the un-peel snapshot semantics).

## Validation

- Full pre-commit battery green on both commits (fmt, clippy, cargo tests, TS, pysimlin).
- `parity_python_and_rust_vdf_extraction_agree` green with the extended corpus (name sets both directions + bitwise values).
- Fixture tests in both readers pin the exact 41-name dropped set and 1235-column total on both SimService files; synthetic unit tests pin the oracle's decision logic (run-boundary trap, internally-ordered-ghost trap, chained-descriptor un-peel scope, inconclusive-to-drop, gate abstention).
- C-LEARN comparator (`simulates_clearn`, release) green on the final tree.
- Both commits were adversarially reviewed by independent sessions; all material and minor findings addressed.

### DIFF
--- a/docs/design/vdf.md
+++ b/docs/design/vdf.md
@@ -197,7 +197,11 @@ records is a subset of the owner records' value set, and the section-6 lookup
 record carries no back-pointer. A reader that has the model knows the descriptor
 set; a model-free reader recognises it from the lookup-def names (see
 "Name-to-OT mapping"). Once the descriptor records are set aside, the remaining
-owner spans form a clean, non-overlapping OT partition.
+owner spans form a non-overlapping OT partition **on every file whose overlaps
+the descriptor peel fully resolves** -- the whole tracked corpus except the
+two 2007-era SimService `Base.vdf` files, where some records survive the peel
+still in owner-vs-owner conflict. That residual case, and the stale-`f[11]`
+record phenomenon behind it, is described in "Residual OT-overlap" below.
 
 ### Slot table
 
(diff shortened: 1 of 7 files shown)


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

Issue #350, opened 2026-02-27
MDL writer EXCEPT round-trip changes behavior for excluded elements


bpowers:
## Problem

When writing an `Equation::Arrayed` with `default_equation` set (from Vensim's `:EXCEPT:` syntax), the MDL writer emits `name[Dim...]=default_eq` which applies the default equation to ALL dimension elements. Elements that were originally excluded via `:EXCEPT:` (and intentionally absent from the elements list, defaulting to 0 in simulation) incorrectly receive the default equation value after MDL round-tripping.

### Example

- **Original MDL**: `x[DimA] :EXCEPT: [A2]= 5` (meaning A1=5, A2 defaults to 0)
- **Current writer output**: `x[DimA]= 5` followed by `x[A1]= 5`
- **After re-import**: A1=5, A2=5 (behavior change -- A2 should still be 0)

## Why it matters

This is a correctness bug. An MDL file written by Simlin and re-imported (by Simlin or by Vensim) will produce different simulation results for any model that uses `:EXCEPT:` syntax to exclude elements from a default equation. The excluded elements silently change from 0 to the default equation value, which can produce subtly wrong simulation output without any error or warning.

## Component(s) affected

- `src/simlin-engine/src/mdl/writer.rs` (lines 988-991)

## Possible approaches

1. **Pass dimension definitions to the MDL writer** so it can determine which elements are excluded and emit proper `:EXCEPT:` syntax in the output (e.g., `x[DimA] :EXCEPT: [A2]= 5`).

2. **Only write element-specific equations without the dimension-level default** -- instead of emitting a dimension-level equation, emit individual element equations only for elements that appear in the elements list (e.g., `x[A1]= 5`). This avoids the need for `:EXCEPT:` syntax in the output but may produce more verbose MDL.

Approach 1 is more faithful to the original MDL representation. Approach 2 is simpler but changes the structure of the output.

## Context

Identified during review of MDL full compatibility work on branch `mdl-full-compat`.

## Related issues

- #341 (protobuf support for EXCEPT equations) -- covers serialization format, not writer correctness
- #342 (XMILE support for EXCEPT equations) -- covers serialization format, not writer correctness

bpowers:
Fixed in commit 9a721854 — MDL writer now emits :EXCEPT: syntax preserving original semantics.

Issue #863, opened 2026-07-07
engine: print_eqn emits stacked unary negatives (--x) that the equation parser rejects


bpowers:
## Problem

`crate::ast::print_eqn` and the equation parser (`Expr0::new`, `LexerType::Equation`) disagree on stacked unary negatives, so `print_eqn` can emit XMILE equation text that does not round-trip through the parser.

Minimal repro AST:

```
Expr0::Op1(Negative, Op1(Negative, App("pi", [])))
```

`print_eqn` serializes this as `--pi()`, but `Expr0::new` **rejects** `--pi()`. The printer is therefore not a total inverse of the parser: for a doubly-negated sub-expression it produces text the parser will not accept.

This is a property of the shared AST printer + equation parser -- **independent of the MDL writer** (it is not about `mdl_bare_keyword` / `PI()` rendering as in #850; the `pi()` in the repro is incidental, any inner expression triggers it).

## Why it matters

Defense-in-depth / correctness of the printer<->parser contract. `print_eqn` output is expected to re-parse to an equivalent AST. Any consumer that serializes an AST via `print_eqn` and re-parses (round-trip harnesses, the MDL writer's `equation_to_mdl` re-parse fallback, potential future tooling) can be tripped by such an AST.

## Impact / reachability

**Effectively unreachable from real models today.** Every stored datamodel equation originated from the parser, which never produces stacked `--` (a leading `--` is rejected at parse time), so no real equation's AST can contain `Op1(Negative, Op1(Negative, _))`. The concern is a latent printer/parser asymmetry, not a live data-corruption bug.

- On the MDL-writer side specifically, `equation_to_mdl` falls back to raw text when it cannot re-parse. In principle that path could leak unparseable XMILE into MDL output for such an equation -- but since the offending AST cannot arise from a real datamodel equation, this is not reachable in production.

## How it was discovered

Surfaced by the MDL-writer property tests (branch \`mdl-writer-hardening\`, \`src/simlin-engine/src/mdl/writer_proptest.rs\`). The proptest currently works around it with a \`prop_filter\` that restricts generated ASTs to those whose \`print_eqn\` output re-parses.

## Component(s)

- \`src/simlin-engine/src/ast.rs\` (\`print_eqn\`)
- the equation parser (\`Expr0::new\`, \`LexerType::Equation\`)

## Possible approaches

- Make \`print_eqn\` parenthesize or space stacked unary negatives, e.g. emit \`-(-x)\` instead of \`--x\`; **or**
[...]

Issue #862, opened 2026-07-07
engine: MDL writer units field re-import hard-fails on non-unit-expression text


bpowers:
## Description

The MDL writer's `units` field is a typed **unit expression**, not free text, but the app lets a modeler type arbitrary text into a units field, and that text survives to the `datamodel`. When a variable's `units` string is not a valid unit expression, the writer serializes it verbatim (after `#849` free-text sanitization) and the exported MDL then **hard-fails `parse_mdl` on re-import**: the whole model fails to load with a units-grammar error (e.g. `expected unit name, found /`), rather than emitting an `ExportWarning` or being sanitized to a valid/empty unit expression.

Triggering inputs include:

- a bare number as units, e.g. `"3.14"`
- a stray comma
- the `/` that the `#849` sanitizer substitutes for a dangling `|` (`"|"` -> `"/"`)

### Relationship to #849 (this is a distinct gap)

`#849`'s sanitizer keeps the units field **structurally safe** -- it never spawns phantom variables or drops the real one, verified by the passing `units_free_text_is_structurally_safe` proptest. But structural safety is not the same as validity: the sanitizer does not (and cannot) make arbitrary text a valid unit expression. So a units string can pass `#849` and still make the exported model fail to load. This is a separate failure mode (hard load failure on re-import) from `#849`'s silent structural corruption (phantom variables from `|`/`~`).

## Why it matters

A `|` (or bare number) entered as units in the app survives to the datamodel, and an MDL export -> re-import round trip then fails to load the model **with no actionable warning**. The user gets an opaque units-grammar parse error instead of either a clean model or an `ExportWarning` naming the offending variable/field.

## Components affected

- `src/simlin-engine/src/mdl/writer.rs` -- `sanitize_free_text` (the units `|` -> `/` substitution) and the units-emission path
- `parse_mdl` (units-expression grammar) on the re-import side

## Possible approaches (policy decision for the writer owner)

1. Validate the sanitized units against the unit-expression grammar in the writer; when it fails, drop to empty units and emit an `ExportWarning` (consistent with the `#856` warnings channel).
2. Make the units-field `|` substitute a **space** (units-insignificant) instead of `/`, and treat bare-number / dangling-operator units as unrepresentable-with-warning.

Note: the units `|` -> `/` choice in `sanitize_free_text` was made deliberately for comment-field trailing-space idempotence, so any change should be **field-specific** (units vs. comment), not global.

## Context

[...]

Issue #861, opened 2026-07-07
engine: MDL export emits per-element graphical-function warnings one-per-element (should aggregate per variable)


bpowers:
## Problem

The arrayed-element graphical-function warnings in `src/simlin-engine/src/mdl/writer.rs` (the Discrete / un-representable-Extrapolate `ExportWarning`s emitted in the arrayed-element loop, around the `write_arrayed_element_entries` path) are emitted **one per element**. A large arrayed variable whose per-element GFs are all Discrete or Extrapolate would emit one `ExportWarning` per element -- potentially thousands of near-identical stderr lines from `simlin-cli`.

The exact-duplicate collapse added at the end of `write_project` (dedups identical messages) does **not** help here: each per-element warning names a distinct element, so the messages are not exact duplicates and are not collapsed.

## Why it matters

Developer experience / stderr verbosity only. Not a correctness or corruption issue. Rare in practice -- the current corpus has zero live per-element extrapolate/discrete GFs -- but a single large arrayed variable of this shape would flood the CLI output.

## Component

`src/simlin-engine/src/mdl/writer.rs` (arrayed-element GF warning emission in the `write_arrayed_element_entries` path).

## Possible approach

Aggregate the per-element GF warnings into **one warning per variable**, e.g. "graphical function for N of variable X's elements uses discrete interpolation; exported clamped", instead of one warning per element. This complements (does not replace) the existing exact-duplicate collapse in `write_project`.

## Relationship to other issues

- Cross-references #856 (the MDL-export lossiness contract + `ExportWarning` channel umbrella), which introduced the warnings mechanism and the exact-duplicate collapse this item refines.
- Distinct from #854 (the underlying semantic drop of `GraphicalFunction.kind` on export). This issue is purely about the *verbosity* of the warnings that surface that degradation, not the degradation itself.

## Discovery context

Identified during review of the MDL writer warning-dedup work (GH #856, branch `mdl-writer-hardening`).

PR #864, opened 2026-07-07
engine: harden the MDL writer (re-parse + semantic-loss + lossiness contract)

### LINKED ISSUE #846: engine: MDL writer corrupts quoted identifiers containing underscores, can drop quoting entirely
## Description

`format_mdl_ident` in `src/simlin-engine/src/mdl/writer.rs` (around line 89) applies `underbar_to_space` to the whole identifier BEFORE checking whether it is a quoted identifier. In Vensim MDL, quoted names are literal -- underscores inside quotes are part of the name -- so this rewrites the interior of quoted identifiers and silently renames variables on export.

Two concrete failure modes (verified via a `simlin-cli` convert round-trip):

1. **Interior corruption**: MDL `"rate_of_change!" = 5` round-trips as `"rate of change!"` -- a different identifier in Vensim.
2. **Quoting dropped entirely**: `"a_b c"` round-trips as the UNQUOTED `a b c`. After the underscore-to-space conversion, `needs_mdl_quoting` no longer sees any character requiring quotes, so the quotes are stripped.

## Why it matters

Silent rename on export is a correctness bug: a model exported to MDL and opened in Vensim (or re-imported) has different variable identities than the source model. Case 2 additionally changes the lexical class of the name (quoted literal vs. normal ident), which can collide with or diverge from other names.

Part of the space-vs-underscore ambiguity is baked in by simlin's canonicalization (spaces and underscores collapse), so a perfect fix may need canonicalization-aware handling. But at minimum the writer should not rewrite the interior of identifiers that are already quoted: detect the quoted form first, and pass the quoted body through literally.

## Components affected

- `src/simlin-engine/src/mdl/writer.rs` -- `format_mdl_ident` (ordering of `underbar_to_space` vs. quote detection) and `needs_mdl_quoting`
- `src/simlin-engine/src/mdl/` reader -- for verifying the round-trip contract

## Possible approaches

1. In `format_mdl_ident`, check for (or strip) the surrounding quotes before applying `underbar_to_space`; emit the quoted body verbatim.
2. Ensure `needs_mdl_quoting` is evaluated against the original (pre-conversion) name so that names which required quoting keep their quotes.
3. Add re-parsing round-trip tests: the existing escape tests in `writer_tests.rs` are string-level only and never re-parse the writer output. This overlaps with #335, which asks for a special-character-identifier round-trip integration test; that test should cover quoted identifiers containing underscores.

## Context

Discovered during MDL writer investigation; verified with `simlin-cli` convert round-trips. Related: #335 (round-trip integration test gap for special-character identifiers).

Severity: medium (silent rename on export).

### LINKED ISSUE #847: engine: MDL writer prints wildcard subscripts as '*' instead of 'Dim!' -- exported .mdl fails re-parse
## Problem

The MDL writer emits XMILE wildcard syntax inside subscript lists, producing `.mdl` files that simlin's own MDL parser rejects.

Vensim `SUM(a[DimA!])` imports as `sum(a[*])` (the bang-subscript reduction is canonicalized to an XMILE wildcard). On export, `MdlPrintVisitor::walk_index` in `src/simlin-engine/src/mdl/writer.rs` (~line 627) prints `IndexExpr0::Wildcard` as `*`:

```rust
fn walk_index(&mut self, expr: &IndexExpr0) -> String {
    match expr {
        IndexExpr0::Wildcard(_) => "*".to_string(),
        ...
```

But MDL subscript lists require the dimension-bang form (`DimA!`). Re-parsing the exported file fails with:

```
expected symbol in subscript list, found *
```

## Root cause

The expression printer is context-free: it has no dimension table, so it cannot reconstruct which dimension the wildcard ranges over. Fixing this requires threading per-variable dimension context into the writer so `a[*]` on a variable arrayed over `DimA` prints as `a[DimA!]`. (Multi-dimensional cases need positional mapping of each wildcard slot to the corresponding dimension.)

## Impact

A corpus sweep (parse -> project_to_mdl -> re-parse over all `test/**/*.mdl` under 200KB) shows ~17 fixtures fail re-parse from this alone, out of 29/240 total re-parse failures (the remainder includes the zero-arg builtin class tracked separately). Affected fixtures include:

- `test/sdeverywhere/models/sum/sum.mdl`
- `test/sdeverywhere/models/vector/vector.mdl`
- `test/sdeverywhere/models/except/except.mdl`
- `test/test-models/tests/subscript_aggregation`
- `test/metasd/scientific-revolution/scirev7.mdl`

Severity: critical for MDL export of any arrayed model using bang-subscript reductions (SUM/PROD/VMIN/VMAX/etc.) -- the exported file is unusable in Vensim and unreadable by simlin itself.

## Component

`src/simlin-engine/src/mdl/writer.rs` (`MdlPrintVisitor::walk_index`, `project_to_mdl` equation path)

## Possible approach

Pass the variable's resolved dimensions into the print visitor (similar context is already available where equations are written per-variable), and map each wildcard subscript position to `format_mdl_ident(dim_name) + "!"`. A corpus-wide parse -> export -> re-parse sweep would make a good regression gate.

Related: the second re-parse failure class (zero-arg builtins like `PI()`) is filed separately.

Context: identified during an MDL writer round-trip corpus sweep.

### LINKED ISSUE #849: engine: MDL writer emits doc/units/group free-text unsanitized, silently corrupting exported models
## Description

The MDL writer emits variable documentation, units strings, group names/docs, and unit-equivalence settings lines **verbatim**, with no sanitization or error when the content contains MDL structural characters. `write_units_and_comment` (`src/simlin-engine/src/mdl/writer.rs:1354`) writes the units and doc raw between the `~` separators and the `|` entry terminator:

```rust
fn write_units_and_comment(buf: &mut String, units: &Option<String>, doc: &str) {
    buf.push_str("\n\t~\t");
    if let Some(u) = units {
        buf.push_str(u);
    }
    buf.push_str("\n\t~\t");
    buf.push_str(doc);
    buf.push_str("\n\t|");
}
```

### Repro

An XMILE aux with documentation `cost is 5 | maybe 6 ~ per year` exports to MDL without error. Re-parsing the exported file yields **4 variables instead of 2**: the embedded `|` terminates the equation entry early, and phantom variables (`dmnl`, `maybe_6` with equation `TIME`) are invented from the trailing text. No error is raised at export or import time -- the model is silently corrupted.

### Affected free-text sinks (all unsanitized)

- Variable `doc` and `units` via `write_units_and_comment` (writer.rs:1354; called from ~6 sites including the settings-section unit entries at line 1598) -- content containing `|` or `~` breaks entry framing
- Group names/docs in the `********` banner emitted by `write_equations_section` (writer.rs:2874) -- a group name containing the banner/terminator syntax corrupts the sketch/group structure
- Unit-equivalence "22:" settings lines (writer.rs:3233) -- a unit alias containing a comma splits into extra equivalences
- A doc containing the literal `\\\---///` sketch-section terminator would end the equations section prematurely

Only view titles are sanitized today (`sanitize_view_title_for_mdl`, writer.rs:1968). Identifier quoting was fixed separately (#330/#335), but that covers variable *names* only, not these free-text fields.

## Why it matters

**Severity: critical -- silent model corruption on export.** A user exporting to Vensim gets a structurally different model (phantom variables, truncated docs, invented equations) with no diagnostic anywhere in the pipeline. This violates the writer's basic contract that the exported file re-parses to an equivalent model, and it is data-dependent: models authored in XMILE (where `|` and `~` are legal in docs/units) are the trigger.

## Components affected

- `src/simlin-engine/src/mdl/writer.rs` -- `write_units_and_comment`, `write_equations_section` (group banners), the settings-section writers (unit-equivalence `22:` lines, unit entries)

## Possible approach

A single sanitize-or-error choke point for all free-text fields written into MDL structural positions, analogous to `sanitize_view_title_for_mdl`: either escape/strip the structural characters (`~`, `|`, the `\\\---///` terminator, commas in settings lists) with a documented lossy transformation, or reject the export with a clear error naming the offending variable/field. Every free-text emission site should route through it. Round-trip tests should include docs/units containing each structural character.

## Context

Discovered during review of the MDL writer's handling of free-text fields; repro verified by exporting an XMILE model with `|`/`~` in an aux doc and re-parsing the output (2 variables became 4, no error raised).

### LINKED ISSUE #850: engine: MDL writer emits zero-arg builtin calls like PI() that the MDL parser rejects (Vensim has no PI builtin)
## Problem

The MDL writer prints zero-argument builtin calls such as XMILE `pi()` as `PI()`. This output is doubly wrong:

1. simlin's own MDL parser rejects it on re-parse: `symbol call requires at least one argument`.
2. Vensim has no `PI` builtin at all, so the exported file is invalid Vensim MDL regardless.

In `src/simlin-engine/src/mdl/writer.rs`, `mdl_bare_keyword` (~line 181) only handles `time`, `dt`/`time_step`, `starttime`/`initial_time`, and `endtime`/`stoptime`/`final_time`; `pi` (and any other zero-arg builtin) falls through to the generic uppercase-function path and prints as `NAME()`.

## Impact

Part of the 29/240 re-parse failures found by a corpus sweep (parse -> project_to_mdl -> re-parse over all `test/**/*.mdl` under 200KB). Affected fixtures include:

- `test/metasd/industrial-dynamics/IDch15/IDch15d.mdl`
- `test/metasd/thyroid/thyroid-2008-d.mdl`

## Component

`src/simlin-engine/src/mdl/writer.rs` (`MdlPrintVisitor::walk`, `mdl_bare_keyword`)

## Possible approach

Emit a numeric literal for `pi()` (e.g. `3.14159265358979`), matching what Vensim models actually contain. More generally, audit all zero-arg XMILE builtins reachable through the writer and give each an explicit MDL rendering (bare keyword, literal, or a documented reject) instead of the generic `NAME()` fallback. The corpus re-parse sweep from #847 would gate this too.

Related: #847 tracks the other (larger) re-parse failure class, wildcard subscripts printed as `*` instead of `Dim!`.

Context: identified during an MDL writer round-trip corpus sweep.

### LINKED ISSUE #852: engine: MDL writer exports 1-arg INITIAL as invalid ACTIVE INITIAL -- name-only inverse of many-to-one function renames (roundtrip harness is blind to the class)
## Problem

The MDL writer exports XMILE `init(x)` as `ACTIVE INITIAL(x)` unconditionally, producing invalid Vensim syntax: `ACTIVE INITIAL` requires 2 arguments, and a 1-arg `init` should export as `INITIAL`.

**Repro:** an MDL model containing `x = INITIAL(y)` (e.g. `test/sdeverywhere/models/initial/initial.mdl`, which is in the corpus) round-trips through import + export to `x = ACTIVE INITIAL(y)` -- wrong function name AND wrong arity for Vensim.

## Root cause: many-to-one import mapping with a name-only inverse

Import direction (`src/simlin-engine/src/mdl/xmile_compat.rs`, `format_function_name`, ~line 558) maps **three** Vensim functions to one XMILE builtin:

```rust
"active initial" => "INIT".to_string(),
"initial" => "INIT".to_string(),
"reinitial" => "INIT".to_string(),
```

Export direction (`src/simlin-engine/src/mdl/writer.rs:203`, `xmile_to_mdl_function_name`) inverts by name only -- the function does not receive the argument list, so it cannot dispatch on arity:

```rust
"init" => "ACTIVE INITIAL".to_owned(),
```

## Why the roundtrip test is blind to this

`mdl_to_mdl_roundtrip` re-parses the exported MDL and compares datamodels. Since `INITIAL`, `ACTIVE INITIAL`, and `REINITIAL` all collapse back to `init` on re-parse, the corrupted export compares equal to the original. The harness structurally cannot catch any many-to-one rename collapse -- it validates simlin-internal fixpoint, not Vensim-syntax validity.

## Same bug class elsewhere in the inverse table

Every many-to-one entry in `format_function_name` needs an arity/shape-aware inverse. Known instances:

1. **`init` (this issue):** 1-arg -> `INITIAL`, 2-arg -> `ACTIVE INITIAL`.
2. **`normal` -> `RANDOM NORMAL`:** XMILE `normal(mean, sd)` (2-arg, optional seed as 3rd) exports as `RANDOM NORMAL(mean, sd)`, but Vensim's `RANDOM NORMAL` requires 5 args (min, max, mean, sd, seed). `reorder_args` in writer.rs only reorders when `args.len() >= 5`, so the 2-arg form passes through as invalid Vensim. Needs synthesized min/max (or `NORMAL` where appropriate).
3. ~~`zidz`/`xidz` -> `safediv`~~ **(correction: already handled)** -- the writer special-cases arity for this pair (`writer.rs:663`: `safediv` with 3+ args emits `XIDZ`, otherwise `ZIDZ`), and a CLI round-trip of `XIDZ(1, 2, 3)` confirms it survives unchanged. Listed here as the *model* for the fix: `init` and `normal` need the same call-site arity dispatch.
4. Audit the remaining collapsed pairs for the same shape: `smooth`/`smoothi` -> `SMTH1`, `smooth3`/`smooth3i` -> `SMTH3`, `delay1`/`delay1i` -> `DELAY1`, `delay3`/`delay3i` -> `DELAY3` (the `i` variants differ by arity -- verify the writer's output arity is valid for the non-`i` Vensim name it picks), `lookup extrapolate` -> `LOOKUP` (extrapolation semantics silently dropped), `vmax`/`vmin` -> `MAX`/`MIN`, `reinitial` -> `init` (re-exported as `ACTIVE INITIAL`).

## Why it matters

- Correctness: exported `.mdl` files fail to parse or simulate in Vensim (high severity -- MDL export is a headline interop feature).
- Test gap: the only roundtrip harness cannot detect this entire bug class, so regressions land silently.

## Suggested approach

- Plumb argument count into `xmile_to_mdl_function_name` (or dispatch at the `App` call site in the writer) and pick the Vensim name by arity: `init` 1-arg -> `INITIAL`, 2-arg -> `ACTIVE INITIAL`; `safediv` 2-arg -> `ZIDZ`, 3-arg -> `XIDZ`; handle `normal`'s arity forms explicitly.
- Add a Vensim-syntax-validity check to the roundtrip harness (per-function arity table for emitted MDL builtins) and/or golden-file tests for exported MDL text, so name collapses cannot silently round-trip.

Identified during review of the MDL writer's function-name inverse table (writer.rs `xmile_to_mdl_function_name` vs xmile_compat.rs `format_function_name`).

### LINKED ISSUE #853: engine: MDL writer silently rebinds user variables named like Vensim control/builtin keywords (Time Step, Time, PI, ...)
## Summary

The MDL writer silently rebinds references to user variables whose names collide with Vensim control variables or bare builtin keywords (`Time Step`, `Time`, `Initial Time`, `Final Time`, `SAVEPER`). The definition line is emitted with the user's name, but every *reference* to the variable is printed as the builtin keyword, so re-importing the exported MDL silently changes simulation results.

## Repro

An XMILE model with:

- aux `Time Step` = `5`
- aux `if` = `Time_Step * 2`

exports to MDL where:

- the definition line reads `Time Step = 5` (user variable, kept), but
- the reference in `if`'s equation is routed through `mdl_bare_keyword` (`src/simlin-engine/src/mdl/writer.rs:181`, used in the expression printer's zero-arg `App` branch around line 648) and printed as the builtin `TIME STEP`.

On re-import, the equation reads the simulation dt (`DT * 2`) instead of the user's variable -- silently changing results. The exported file also contains two conflicting `TIME STEP` definitions: the user's and the `.Control` section's.

The same class of corruption applies to user variables named `Time`, `Initial Time`, `Final Time`, and `SAVEPER`.

## Related import-direction sibling: PI

A user variable literally named `PI` (e.g. `test/metasd/industrial-dynamics/IDch15/IDch15d.mdl` has `PI=3.14159`) has its references rewritten to the engine builtin `pi()` on import, which then exports as unparseable `PI()`. Same root confusion between user-variable names and zero-arity builtin/keyword names, in the opposite direction.

## Root cause

The MDL expression printer is a context-free printer today: it has no access to the model's variable-name set, so it cannot distinguish "zero-arg builtin call that should print as a bare keyword" from "reference to a user variable that happens to share the keyword's canonical name". (Upstream, the XMILE-side parse likely resolves the colliding name as the zero-arity builtin in the first place, so the corruption may begin before the writer -- see #681 for the same shadowing class with `inf`.)

## Why it matters

- **Severity: high -- silent semantic corruption.** Simulation results change on an export/import round trip with no diagnostic.
- Exported MDL can also be structurally invalid (duplicate `TIME STEP` definitions; unparseable `PI()`).

## Components affected

- `src/simlin-engine/src/mdl/writer.rs` (`mdl_bare_keyword`, expression printer `walk`)
- XMILE-side equation parsing / builtin resolution (zero-arity builtin shadowing of user variables)
- MDL import (PI direction)

## Possible approaches

- Thread the model's variable-name set into the MDL expression printer so keyword-vs-variable is decided per name; quote or rename colliding user variables on export (and keep the reference spelling consistent with the definition line).
- On the import/parse side, prefer a declared user variable over a zero-arity builtin when a name collides (mirroring whatever resolution #681 lands on for `inf`).
- Add round-trip tests for user variables named `Time Step`, `Time`, `Initial Time`, `Final Time`, `SAVEPER`, and `PI`.

## Related issues

- #552 -- import-direction sibling: macro formal parameter named like a control variable is silently dropped on MDL import (distinct mechanism).
- #681 -- variable named `inf` (or other zero-arity builtin name) collides with the builtin and silently reads +infinity (same shadowing class, XMILE side).

## Discovery context

Identified while auditing MDL export fidelity: the writer's bare-keyword mapping was observed rebinding references without consulting the model's variable names.

### LINKED ISSUE #854: engine: MDL writer silently drops GraphicalFunction.kind (extrapolate/discrete lookups exported as plain continuous)
## Problem

The MDL writer ignores `GraphicalFunction.kind` when exporting lookups, so any non-continuous graphical function is written back as a plain continuous lookup -- a silent semantic change.

Concretely:

- The MDL importer (`src/simlin-engine/src/mdl/convert/variables.rs:1434-1439`) produces `GraphicalFunctionKind::Extrapolate` when `table.extrapolate` is set or the lookup is used via `LOOKUP EXTRAPOLATE` / `TABXL` anywhere in the model.
- The MDL writer (`src/simlin-engine/src/mdl/writer.rs`, `write_lookup_body` at line 784 and `write_lookup` at line 816, plus their call sites at lines 1256/1266/1324/1333) only reads `x_points`/`y_points`/`x_scale`/`y_scale`. `gf.kind` is never consulted anywhere in writer.rs.

Consequences on a round trip (or on XMILE -> MDL export of a discrete/extrapolate GF):

- **Extrapolate** lookups become clamping lookups: behavior outside the table's x-range silently changes (Vensim's default LOOKUP clamps at the endpoints; `LOOKUP EXTRAPOLATE` extends the end segments linearly).
- **Discrete** GFs (XMILE `type="discrete"`) become interpolating lookups: behavior *between* points silently changes from stepwise to linear interpolation.

## Why it matters

Silent simulation-semantics change on export (medium severity). A model that simulates one way in Simlin produces different results after export to Vensim, with no error or warning.

## Component

`src/simlin-engine/src/mdl/writer.rs` (and the round-trip contract with `src/simlin-engine/src/mdl/convert/variables.rs`).

## Possible approach

- For `GraphicalFunctionKind::Extrapolate`, emit `LOOKUP EXTRAPOLATE(<table>, <input>)` call syntax at WITH LOOKUP call sites (and/or preserve whatever original spelling produced the extrapolate flag). Note the importer sets Extrapolate on the table when *any* use site is `LOOKUP EXTRAPOLATE`/`TABXL`, so the writer may need to emit extrapolating call syntax at the lookup's use sites rather than in the table definition itself.
- For `GraphicalFunctionKind::Discrete`, Vensim has no direct WITH LOOKUP equivalent; the writer should at minimum error or emit a loud warning instead of silently changing interpolation semantics.
- Add MDL round-trip tests covering extrapolate and discrete kinds (import LOOKUP EXTRAPOLATE -> write -> re-import -> assert kind preserved; XMILE discrete GF -> MDL export -> expect error/warning).

## Related (distinct) known gaps

- #393 -- MDL writer does not reconstruct GET DIRECT calls for arrayed variables.
- The arrayed/per-element GF write path similarly ignores per-element comments and `Equation::Arrayed` + `compat.active_initial` / `data_source` combinations.

## Discovery context

Identified during code review of the MDL writer's lookup path.

### LINKED ISSUE #856: engine: MDL export has no lossiness contract -- callers can't learn an export was degraded
## Problem

`project_to_mdl` in `src/simlin-engine/src/mdl/writer.rs` mixes three incompatible failure modes with no unifying contract and no channel to surface degradation to callers:

- **hard error** for some unsupported constructs (e.g. `Variable::Module`, mod.rs:88-94)
- **silent drop** for many fidelity/semantic fields (see inventory below)
- **best-effort garbling** for others (an unparseable equation falls through `equation_to_mdl`, printing the XMILE text with underscores converted to spaces, writer.rs:758)

A caller such as `simlin-cli` gets back a `String` and cannot tell whether the export was faithful, silently lossy, or actively corrupted. There is no documentation of which category each field falls into.

## Inventory of silent drops (field-by-field audit, 2026-07)

**Semantic (changes simulation results):**
- `compat.non_negative` -- never referenced by the writer; the exported model can go negative in Vensim where the XMILE source clamps. (Enforcement in the compiler/VM is separately tracked in #545; this is the export-side drop.)

**Fidelity (structure/metadata loss):**
- `Model.loop_metadata` (LOOPSCORE pins)
- `ModelGroup.parent` sector nesting (importer populates it, convert/variables.rs:206) and `run_enabled`
- per-element comments in arrayed equations (writer.rs:1078, 1317)
- one-to-many element-level dimension mappings (writer.rs:1658-1683 falls back to plain name mapping; documented in-code)
- `StockFlow.zoom` / `view_box`
- `compat.visibility`
- `ai_state`, `Project.ai_information`
- `Dimension.parent` indexed subdimensions (XMILE-only)

## Why it matters

Correctness (`non_negative` changes results), data loss (fidelity fields vanish on round-trip), and developer experience (no signal that an export degraded). This is the design umbrella above the individual per-field bugs.

## Proposal

1. **Document the contract**: for each field/construct, specify whether export errors, warns, or silently drops -- and why.
2. **Add a warnings channel**: change `project_to_mdl` to return `(String, Vec<ExportWarning>)` (or add a warnings out-param) so callers like `simlin-cli` can print "exported with N fidelity warnings" and enumerate them.
3. **Escalate `non_negative`**: hard-error or `MAX(0, ...)`-wrap on export, since silently dropping it changes simulation semantics.

## Scope / relationship to other issues

This is the design/tech-debt umbrella. Individual semantic items are tracked separately: GraphicalFunction.kind #854, free-text sanitization #849, quoted-underscore idents #846, Module-in-macro drop #566, EXCEPT defaults, stock compat, `non_negative` enforcement #545, view-element fidelity #811. This issue tracks the missing *contract + warnings mechanism* itself, not any single field.

## Discovery

Identified during a field-by-field audit of `src/simlin-engine/src/mdl/writer.rs` (2026-07).

### LINKED ISSUE #857: engine: MDL writer ignores compat on stocks and arrayed vars, silently dropping ACTIVE INITIAL / GET DIRECT
## Problem

The MDL writer (`src/simlin-engine/src/mdl/writer.rs`) drops `compat` initialization metadata in two places, silently changing simulation initialization semantics on MDL round-trip.

### Gap 1: `write_stock_variable` never reads `stock.compat`

`write_stock_variable` (writer.rs:974-1031) does not consult `stock.compat` at all. `datamodel::Stock` carries a `compat: Compat` field, but none of the three equation arms (`Scalar`, `ApplyToAll`, `Arrayed`) pass it through:

- A stock imported with an `ACTIVE INITIAL(...)` initial value loses the `ACTIVE INITIAL` wrapper on export (its `compat.active_initial` is never applied).
- A stock with a `GET DIRECT ...` data-source initial loses the reconstructed `GET DIRECT` call on export (`compat.data_source` is never consulted).

By contrast, the aux/flow path (`write_variable`, writer.rs:887-916) handles both cases: `compat_get_direct_equation(compat)` reconstructs the `GET DIRECT` expression, and `wrap_active_initial(eqn, compat)` re-wraps `ACTIVE INITIAL`. Stocks get neither.

### Gap 2: `wrap_active_initial` / GET DIRECT not applied to `Equation::Arrayed`

Even in the aux/flow path, `wrap_active_initial` and `compat_get_direct_equation` are only applied to `Equation::Scalar` (writer.rs:895) and `Equation::ApplyToAll` (writer.rs:902). The `Equation::Arrayed` arm (writer.rs:913-915) calls `write_arrayed_entries` with neither `compat` nor the data-source equation, so an arrayed aux/flow with `compat.active_initial` silently loses its `ACTIVE INITIAL` wrapper. (The arrayed stock arm, `write_arrayed_stock_entries` at writer.rs:1066, has the same omission as part of Gap 1.)

## Why it matters

`ACTIVE INITIAL` and `GET DIRECT` change how a variable is *initialized* at t=start. Dropping them on round-trip produces an MDL file that initializes differently from the source model -- a silent semantic loss, not a cosmetic diff. Severity: medium-high, because it corrupts real imported models on export with no error.

## Components affected

- `src/simlin-engine/src/mdl/writer.rs` (`write_stock_variable`, `write_arrayed_stock_entries`, the `Equation::Arrayed` arm of `write_variable` / `write_arrayed_entries`)

## Possible approaches

- Thread `stock.compat` into `write_stock_variable` and apply `compat_get_direct_equation` / `wrap_active_initial` in each arm, mirroring the aux/flow path.
- Extend the arrayed writers (`write_arrayed_entries` and `write_arrayed_stock_entries`) to accept `compat` and apply the `ACTIVE INITIAL` wrapper (and, where in scope, the GET DIRECT reconstruction) to arrayed equations.
- Add round-trip fixtures: a stock with an `ACTIVE INITIAL` initial, a stock with a `GET DIRECT` initial, and an arrayed aux/flow with `ACTIVE INITIAL`, asserting the wrapper/call survives export. The existing `mdl_to_mdl_roundtrip` datamodel-fixpoint harness may not catch this if import collapses the forms; a golden-text assertion on the emitted MDL is more reliable.

## Related but distinct

- #393 tracks GET DIRECT reconstruction for **arrayed** variables specifically (aux/flow element-by-element expansion), not stocks and not `active_initial`.
- #852 tracks the `init` -> `ACTIVE INITIAL` function-name **arity** inversion (wrong Vensim name/arity), a separate concern from `compat` metadata being dropped entirely.

## Context

Identified during review of the MDL writer's compat/initialization handling.

### LINKED ISSUE #858: engine: MDL writer silently drops Equation::Arrayed default_equation (EXCEPT), losing equations for default-only elements
## Problem

The MDL writer silently drops `Equation::Arrayed.default_equation` (the EXCEPT default). For a dimension element that is covered *only* by the default equation (i.e. not present in the enumerated `elements` list), the exported `.mdl` contains **no equation at all** for that element -- silently changing simulation values or producing an incomplete model.

`write_arrayed_entries` (`src/simlin-engine/src/mdl/writer.rs:1297`) and `write_arrayed_stock_entries` (`:1066`) both bind `_default_equation` / receive the default but never re-emit a `name[Dim] :EXCEPT: [...]` default entry -- only the enumerated `elements` are written (via `write_arrayed_element_entries`).

The importer DOES populate `default_equation` from Vensim `:EXCEPT:` syntax (`src/simlin-engine/src/mdl/convert/variables.rs:489,562-567`), so an MDL->MDL round-trip of any Vensim model using EXCEPT loses the default equation entirely: elements covered only by the default get NO equation in the exported file.

### Concrete example

- **Original MDL**: `x[DimA] :EXCEPT: [A2] = 5` with `DimA = {A1, A2, A3}` (meaning A1=5, A3=5, A2 defaults to 0)
- **Datamodel after import**: `Equation::Arrayed { default_equation: Some("5"), elements: [(A2, "0")], has_except_default: true }` (only excepted elements enumerated; the rest come from the default)
- **Writer output today**: only `x[A2] = 0` -- A1 and A3 get **no equation**, silently changing their simulated value from 5 to (missing/0).

## Why it matters

Silent semantic loss on export of *any* EXCEPT-using model. No error or warning is produced; the round-tripped model simply computes different results (or is incomplete). The MDL writer is a supported export path.

Note the existing code comment at `writer.rs:1290-1296` frames the current behavior as conservative ("emitting `name[Dim]=default` can incorrectly apply the default to omitted EXCEPT members"). That is true of the *naive* re-emit, but the actual effect of the current code is **dropped equations**, not merely conservative formatting.

## Relationship to #350 (regression / reopen candidate)

Closed #350 described the *opposite* failure mode: the writer emitted `name[Dim...]=default`, which applied the default to elements that EXCEPT had intentionally excluded. Its resolution took approach 2 from that issue ("only write element-specific equations without the dimension-level default"), which fixed the over-application but introduced *this* under-application: elements covered only by the default now get nothing. So this is effectively a regression introduced by the #350 fix -- consider reopening #350 or tracking here as its follow-up.

## Component(s) affected

- `src/simlin-engine/src/mdl/writer.rs` (`write_arrayed_entries` :1297, `write_arrayed_stock_entries` :1066, `write_arrayed_element_entries`)

## Possible approaches

The fix needs, at the writer call site, either:

1. **Dimension-membership info** to reconstruct the EXCEPT list -- emit `name[Dim] :EXCEPT: [<enumerated elements>] = default` for the default-covered members while keeping the explicit element entries, or
2. **The importer's original except metadata** -- `Equation::Arrayed.has_except_default` (added in #356) already records that the array is EXCEPT-originated; carry enough of the original EXCEPT structure through to reconstruct faithful `:EXCEPT:` output.

Approach 1 (reconstructing proper `:EXCEPT:` syntax) is the faithful representation #350 originally recommended as its approach 1.

## Testing

Related test-harness gaps that let this ship undetected are tracked in #851 -- in particular #851(a): `assert_semantic_equivalence` compares all `Equation::Arrayed` values as the literal string `"<arrayed>"`, so per-element/default equation drift is never caught. A fix here should add a round-trip test that exercises an EXCEPT-with-default fixture and asserts default-covered elements survive.

## Context

Identified during MDL writer bug investigation, 2026-07-06.

bpowers:
## Why

The MDL writer (`project_to_mdl`, the export path behind `simlin-cli convert --to mdl`) was the subject of a pre-ship adversarial audit. That audit found two silent-corruption classes and several semantic-loss bugs: a context-free expression printer that emitted unparseable subscripts, raw free text that could terminate a construct early and invent phantom variables on re-import, and a handful of constructs (graphical-function kind, EXCEPT defaults, stock/arrayed initials) that were silently dropped. About 12% of the model corpus exported to files that would not re-parse.

This PR hardens the writer to the point where the corpus re-parses cleanly and the known semantic-loss bugs are fixed, and it adds the test scaffolding (a two-way corpus ratchet plus property tests) that keeps those wins from regressing. The only production caller is the CLI converter, so these are export-quality fixes, not user-file-destruction fixes.

## What, in five reviewed steps

Each step was implemented by a fresh sub-agent under strict TDD and then adversarially reviewed by a different fresh sub-agent; findings were folded in before commit.

**1. Corpus round-trip ratchet gates** (`tests/integration/mdl_roundtrip.rs`). Two-way allowlists over every `.mdl` fixture: `corpus_reparse_ratchet` (writer output must re-parse) and `writer_output_idempotence_ratchet` (write is a fixpoint). An unlisted failure is a regression; a listed-but-now-passing entry is also a failure, so the allowlists can only shrink. Coverage-floor guards prevent a vacuous pass. This landed first (green) so every subsequent step had an objective, non-regressing scoreboard.

**2. Context-aware expression printer.** A `WriterContext` (built per model) threads the variable-ident set, each arrayed variable's declared dimensions, the standalone-extrapolating-lookup idents, and each named dimension's element list into the printer. This resolves what a context-free printer cannot: wildcard subscript recovery `a[*]` -> `a[DimA!]` (#847), the `pi` literal (#850), keyword-shadow resolution for user variables named like builtins (#853), INITIAL vs ACTIVE INITIAL arity at the call site (#852), and quoted-ident preservation (#846). Emptied the re-parse-failure allowlist from 27 to 0.

**3. Free-text sanitization choke point** (`sanitize_free_text`). Every modeler-authored free-text sink (units, documentation, group name/doc, `22:` tokens) routes through one function. Policy: `|` -> `/` (a non-whitespace substitute so a field-final `|` cannot become an idempotence-breaking trailing space), the four section-terminator runs -> space, per-field separators -> space, line endings normalized losslessly to LF. Fixes the confirmed `|`-in-doc phantom-variable corruption (#849) and cut the idempotence allowlist from 103 to 80.

**4. Semantic-loss fixes + a lossiness contract.** `project_to_mdl_with_warnings` returns `(String, Vec<ExportWarning>)` and is the new primary entry point; `project_to_mdl` stays a thin wrapper that discards warnings so existing callers are unchanged. The CLI prints warnings to stderr. Warnings never change the emitted text, so they do not affect the ratchets (#856). On top of the channel: stock and arrayed `active_initial` / GET DIRECT initials are preserved (#857); a graphical function's Extrapolate kind is preserved by rewriting its `LOOKUP` call sites to `TABXL` -- the same marking the importer reads (#854); and an `Equation::Arrayed` with an EXCEPT default materializes the declared-but-missing elements with the default value instead of silently dropping them to 0 (#858, reopening the loss the #350 element-only fix introduced). The EXCEPT reconstruction was verified with a simulation-equivalence check (`except.mdl` now yields `s[A1]=14/s[A2]=14/s[A3]=13`, identical to the original run).

**5. Property tests + docs.** Four properties (`writer_proptest.rs`): free text never corrupts the variable set, the same text in the typed units field is structurally safe, equation write is a byte-equal re-parse fixpoint (a grammar aimed at the printer fixes), and whole-model write is idempotent. The review confirmed each is non-vacuous and non-tautological (a reverted fix trips the property). `mdl/CLAUDE.md` gains a Writing section and an updated Known Gaps list.

## Decisions worth calling out

- **Warnings are a side channel, not a signature break for everyone.** Keeping `project_to_mdl -> Result<String>` as a wrapper avoided churning ~40 test callsites while still giving the one production caller (the CLI) a way to surface degradation.
- **GF extrapolate is marked at the call site, not the definition.** MDL has no definition-level extrapolate flag; the importer only infers it from a `TABXL` call. So the faithful inverse is to rewrite the call, which is what this does. An inline `WITH LOOKUP` Extrapolate GF (no call site to mark) warns instead.
- **EXCEPT is reconstructed by materialization, not by re-emitting `:EXCEPT:` syntax.** The reader materializes EXCEPT coverage into explicit elements, so re-emitting the sugar is non-idempotent (the explicit set grows each write). Materializing the default-covered elements is both value-correct and a re-import fixpoint. It declines with a warning when dimension membership is unavailable or the default references its own dimensions.
- **The units field gets an honest, weaker guarantee.** Units is a typed unit expression, not prose; sanitization keeps it structurally safe (never spawns phantom vars) but cannot make arbitrary text a valid unit expression. That asymmetry is documented and pinned by a separate property.

## Follow-ups filed (not in this PR)

- #859 -- importer non-determinism (a first-import source of round-trip variance).
- #861 -- per-element arrayed GF warnings are unbounded (stderr verbosity, not corruption).
- #862 -- app-entered garbage units hard-fail re-import with no ExportWarning (pre-existing, surfaced by the units property).
- #863 -- `print_eqn` emits `--x` for stacked unary negatives, which the equation parser rejects (a shared AST/parser asymmetry, unreachable from real models).

## Closes

Fixes #846
Fixes #847
Fixes #849
Fixes #850
Fixes #852
Fixes #853
Fixes #854
Fixes #856
Fixes #857
Fixes #858

Substantially addresses #851 (the all-corpus re-parse gate, the idempotence ratchet, and the property tests all land here); the general simulation-equivalence oracle it also proposes remains open. #848's free-text non-idempotence class is fixed; its other non-idempotence classes (time-unit drift, `TIME BASE` rewriting, control-variable substitution) are enumerated in the writer's Known Gaps and remain in the idempotence allowlist.


bpowers:
@codex review

Issue #899, opened 2026-07-08
build: TypeScript test files are never type-checked (~55 latent type errors)


bpowers:
## Problem

No test file in the repo is type-checked.

`tsconfig.base.json` sets `"isolatedModules": true`, which made ts-jest run transpile-only for every package, so type errors under `tests/` were silently ignored. The migration to Rstest (swc-based) does not change this -- swc also strips types without checking them -- so nothing was *lost* in that migration, but the hole is now explicit and worth closing.

Four packages (`core`, `engine`, `diagram`, `app`) explicitly exclude tests from their tsc program:

- `src/core/tsconfig.json` -> `"exclude": ["lib", "lib.module", "tests"]`
- `src/engine/tsconfig.json` -> `"exclude": [..., "tests", ...]` (and `@simlin/engine` is not even in the root `tsc` script's `--filter` list)
- `src/diagram/tsconfig.json` -> `"exclude": ["lib", "lib.browser", "tests"]`
- `src/app/tsconfig.json` -> `"exclude": [..., "../**/tests/**", "../**/*.test.ts", ...]`

Two packages (`@simlin/server`, `@simlin/serve-web`) *do* type-check their tests today, but only incidentally: their tsconfig `include` globs happen to cover the test files, so `pnpm tsc` picks them up.

## Proof the hole is real

All of these currently compile clean, because nobody looks:

- `src/diagram/tests/editor-applyPatch.test.ts:11` imports `JsonProjectPatch` from `'../json-types'` -- **that module does not exist**. The import survives because it is type-only and gets elided.
- `src/diagram/tests/build-selection-map.test.ts:9` imports `UID` from `'@simlin/core/common'`; `UID` is actually declared in `@simlin/core/datamodel`.
- `src/diagram/tests/editor-details-remount.test.ts:33` (and `editor-selection-changed`, `editor-selection-invariant`): `Module '"@simlin/core/datamodel"' declares 'JsonProject' locally, but it is not exported`.
- `src/engine/tests/cleanup.test.ts:41`: the fake `EngineBackend` is missing `projectRenderPng` -- the interface drifted and the fake never followed.

## Why it matters

[...]

Issue #895, opened 2026-07-08
build: eslint.config.shared.js mutates module-level shared objects (cross-call clobbering, fragile index lookup)


bpowers:
## Problem

`eslint.config.shared.js` defines `baseConfig`, `reactConfig`, and `jestConfig` as **module-level object literals**. `createConfig()` pushes those same objects (by reference) into the array it returns, then mutates one of them in place:

```js
const baseConfig = { /* module-level, shared */ };

const createConfig = (options = {}) => {
  const configs = [];
  if (options.ignorePatterns) {
    options.ignorePatterns.push('eslint.*.js');   // (b) mutates the CALLER's array
    configs.push({ ignores: options.ignorePatterns });
  }
  configs.push(baseConfig);                        // by reference
  if (options.react) configs.push(reactConfig);    // by reference
  configs.push(jestConfig);                        // by reference

  if (options.project) {
    // (a) mutates the module-level baseConfig, and (c) finds it by index arithmetic
    configs[configs.length - (options.react ? 3 : 2)]
      .languageOptions.parserOptions.project = options.project;
  }
  return configs;
};
```

Three distinct defects follow.

### (a) Two `createConfig()` calls in one process clobber each other

Because every call returns the *same* `baseConfig` object, the last call's `project` wins for all of them. Verified by loading the real module with its dependencies stubbed:

```
call A requested ./core.json    -> A reports: "./server.json"
call B requested ./server.json  -> B reports: "./server.json"
A.base === B.base (same object): true
=> LAST CALL WINS FOR BOTH: true
```

This is **latent, not currently firing**: ESLint loads one flat config per package in its own process, so `createConfig()` is called once per process today. It becomes a live bug the moment anything calls it twice -- a root config that composes packages, a programmatic `ESLint` API run over the monorepo, a lint-staged setup, or a test that imports two package configs. The failure mode is silent and awful: one package type-checks against another package's tsconfig.

### (b) It mutates the caller's `ignorePatterns` array

`options.ignorePatterns.push('eslint.*.js')` writes into the array the caller passed in. Repeated calls accumulate duplicates:

```
caller's array after one call:  ["lib/","eslint.*.js"]
caller's array after two calls: ["lib/","eslint.*.js","eslint.*.js"]
```

Harmless today only because each consumer passes a fresh array literal.

### (c) It locates `baseConfig` by fragile index arithmetic

`configs[configs.length - (options.react ? 3 : 2)]` identifies the base config positionally.

[...]

Issue #523, opened 2026-05-12
app: jest.config.js hardcodes a pnpm virtual-store hash for wouter; breaks on every React bump


bpowers:
## Problem

`src/app/jest.config.js` maps the ESM-only `wouter` package to its real source via `moduleNameMapper`, but does so by hardcoding a full pnpm virtual-store path that embeds the React peer-dependency version hash:

```js
'^wouter$': '<rootDir>/../../node_modules/.pnpm/wouter@3.9.0_react@19.2.4/node_modules/wouter/src/index.js',
'^wouter/(.*)$': '<rootDir>/../../node_modules/.pnpm/wouter@3.9.0_react@19.2.4/node_modules/wouter/src/$1',
```

pnpm's virtual store directory name for `wouter` includes the resolved peer (`wouter@3.9.0_react@<react-version>`), so any bump to React rewrites that path. `pnpm-lock.yaml` was recently bumped so React resolves to `19.2.6`, and the installed package now lives at `node_modules/.pnpm/wouter@3.9.0_react@19.2.6/node_modules/wouter/...`. The hardcoded `..._react@19.2.4` path no longer exists.

## Impact

- Every test in `src/app/tests/app.test.tsx` fails to load with a Jest `createNoMappedModuleFoundError` at `import { memoryLocation } from 'wouter/memory-location';`. Repro: `pnpm --filter @simlin/app run test`.
- Any future suite in `@simlin/app` that imports `wouter` (or a `wouter/*` subpath) will hit the same failure.
- The `@simlin/app` portion of the `pnpm test` run (pre-commit hook step 7 + CI) is red on the branch that bumped React.
- Same fragility recurs on any React version change going forward -- it's a latent landmine, not a one-off.

## Component(s) affected

- `src/app` test config (`src/app/jest.config.js`)

## Why it matters

Pinning a pnpm virtual-store path -- which by design encodes peer-dep version hashes -- is inherently fragile. The mapping exists for a legitimate reason (per the comment in `jest.config.js`: `wouter` is ESM-only / `"type":"module"`, and CommonJS ts-jest can't resolve its `package.json` `"exports"` conditions), but the path should not be hash-pinned.

## Possible approaches

- Resolve `wouter` without the hash, e.g. `require.resolve('wouter')` and `require.resolve('wouter/' + '$1')`-style resolution in the config (Jest `moduleNameMapper` values can be computed in the JS config file).
[...]

PR #900, opened 2026-07-08
build: TypeScript 7, Rsbuild 2, and Jest -> Rstest

### LINKED ISSUE #635: server: @simlin/server Jest suite intermittently fails one test under parallel pre-commit load
## Summary

The `@simlin/server` Jest suite (`src/server`) intermittently fails **one** test when run under the parallel load of the pre-commit hook. A single pre-commit run reported `1 failed, 72 passed` for `@simlin/server`, but re-running the suite standalone passed cleanly `73/73` (confirmed twice). This is a pre-existing flake: the change that triggered the failed pre-commit run was a Rust-only change to `src/simlin-engine/src/layout/metrics.rs`, completely unrelated to the server, so the test itself was almost certainly the cause rather than the diff under test.

## Why it matters

This undermines the reliability of the pre-commit gate (`scripts/pre-commit`). The hook runs many checks in parallel (Rust fmt/clippy/tests, WASM build, TS lint/typecheck/tests, Python tests -- see root `CLAUDE.md` "Pre-commit Hooks"), and under that contention one server test flips red. The consequence is a spurious red on a branch that is actually green: a developer can be blocked (or worse, conditioned to distrust the gate) by a failure that has nothing to do with their change. It is a developer-experience and CI-trust problem, not a product correctness bug, but it erodes the value of the canonical gate.

This is the same *class* of problem as the recently-filed #629 (pre-commit Rust pipeline spuriously failing on a cold cache due to parallel `clippy --all-features` + capped `cargo test` contending on the package-cache lock), but in a different component: that one is the Rust pipeline / cargo package-cache lock; this one is the `@simlin/server` Jest suite. It is also distinct from #474 (an order-dependent/flaky *Rust engine* test) and from the pysimlin Hypothesis health-check flake tracked in `docs/tech-debt.md` item 19.

## Component affected

- `src/server` -- the `@simlin/server` Jest test suite (7 test files; see `src/server/CLAUDE.md`).
- Surfaces through `scripts/pre-commit` (and by extension CI, if the same suite runs there under parallel load).

## How it reproduces / what's known

- Symptom: under parallel pre-commit load, `1 failed, 72 passed`; standalone re-run passes `73/73` (reproduced-clean twice in isolation).
- The specific failing test has **not** been identified yet -- the parallel run did not surface (or the surfacing was not captured) which of the 73 tests flaked.
- Likely root-cause families for a "fails under load, passes in isolation" flake: test-isolation / shared mutable fixture state, a hard-coded port or other shared OS resource, a timer/`setTimeout`-driven async race, or resource contention (CPU starvation pushing an implicit timeout over the edge when the box is busy with the rest of the pre-commit run).

## Possible approaches for resolution

1. **Identify the flaky test.** Run the server suite repeatedly under load to reproduce. Useful levers:
   - Run with reduced/maximal worker parallelism to bracket the behavior, e.g. `pnpm --filter @simlin/server test -- --runInBand` (serial) vs. the default parallel run, and a high `--maxWorkers` while the machine is otherwise busy.
   - Loop the suite (e.g. a shell `for` loop, or Jest with a repeat) while a CPU/IO load generator runs in parallel to mimic the rest of the pre-commit hook.
   - Capture the failing test name and its error/stack the first time it trips.
2. **Find the root cause** once the test is known: look for shared module-level fixture state not reset between tests, a fixed port / temp path collision, a real timer vs. fake timers, an unawaited promise, or an implicit timeout that's too tight under contention.
3. **Make it deterministic:** isolate the shared state (per-test fixtures / `beforeEach` reset), bind to an ephemeral port (`:0`) or unique temp paths, use fake timers / explicit `await`s instead of wall-clock waits, and/or relax overly tight timeouts. Prefer fixing the root cause over papering it with retries.

## Context

Discovered during the layout-quality-eval work (branch `layout-quality-eval`). The triggering pre-commit run was for an unrelated Rust-only change to `src/simlin-engine/src/layout/metrics.rs`; the server flake is therefore pre-existing and not caused by that change.

bpowers:
Modernizes the TypeScript toolchain in three commits: TypeScript 7, Rsbuild 2, and Jest -> Rstest. They ship together because they are entangled -- ts-jest cannot run on TypeScript 7, and picking a replacement runner constrains which bundler the repo carries.

No open issue is resolved by this PR, so there is no closing keyword. See "Does not close #635" below.

## 1. `build: migrate to TypeScript 7`

TypeScript 7 is the Go-native compiler rewrite. It ships **no programmatic compiler API** (deferred to 7.1) and no `tsserver`, so anything that embeds the compiler -- ts-jest, `@rsbuild/plugin-type-check`, typescript-eslint -- still needs a TS6-era package.

The repo therefore carries two entries per package, which is the layout the TypeScript team recommends for exactly this transition:

- `"typescript": "npm:@typescript/typescript6@^6.0.2"` -- the compat shim. Aliased as `typescript` so peer-dependency ranges still see `6.0.2`; its binary is renamed `tsc6` to avoid colliding with TS7's.
- `"@typescript/native": "npm:typescript@^7.0.2"` -- supplies the native `tsc` that `pnpm tsc` and the builds actually run.

Most of the breaking changes landed in **TypeScript 6**, not 7. The ones that bit us:

- `"moduleResolution": "node"`/`"node10"` were removed. `"bundler"` is the closest analogue (extensionless relative imports plus `exports` support), and pairing it with `"module": "commonjs"` only became legal in TS6.
- `"baseUrl"` was removed. Without it, `paths` targets resolve relative to the config file that *declares* them (even through `extends`), so every target is now `./`-prefixed. TS7 rejects bare non-relative targets outright (TS5090).
- `"types"` now defaults to `[]` rather than "every installed `@types` package". Each package names the globals-only packages it actually needs. `Symbol.dispose` used to arrive implicitly via `@types/node`; the engine now names `esnext.disposable`, the lib that owns it.
- `@types/resize-observer-browser` is gone -- `lib.dom` has declared `ResizeObserver` since TS 4.2. A triple-slash `/// <reference types="resize-observer-browser" />` in `Canvas.tsx` was keeping it alive.

**Emit is unchanged.** I diffed all 446 emitted files between TS 5.9 and TS 7: three `.js` files differ textually and all three are AST-identical (re-printed and compared). Nineteen `.d.ts` files change how `React.memo` components are spelled -- a TS6 change -- and the new type is assignable to the old (verified with a standalone assignability probe).

**Gotcha for reviewers:** a stale `.tsbuildinfo` written by TS 5.9 makes TS7 report hundreds of phantom errors. Buildinfo lives at `<outDir>/<configName>.tsbuildinfo`; `src/engine`'s `clean` script now removes it. TS6-written buildinfo *is* read correctly by TS7 -- only the 5.9 artifacts are poison. If you see a wall of errors after checking this branch out, `rm` the buildinfo.

## 2. `app: upgrade to Rsbuild 2`

`@rsbuild/core` is now pure ESM and requires Node 20.19+/22.12+ (exactly where `require(esm)` became available).

Two real behavior fixes fell out of reading the changelog rather than trusting the config:

- **The dev-server proxy was silently matching everything.** Rsbuild 2 bundles http-proxy-middleware 4, which renamed `context` to `pathFilter`. Neither the old key nor `logLevel` errors -- they are ignored -- and an array-form proxy entry with no `pathFilter` matches every request. Verified counterfactually against a dead backend: with `context:`, `GET /` returns 504 (the SPA itself was being proxied); with `pathFilter:`, SPA routes return 200 HTML and all seven API prefixes return 504.
- **`PORT` never worked.** `startDevServer()` has never accepted a `port` option (checked against both 1.x and 2.x type definitions). The port now comes from `server.port`, where it belongs.

Also `performance.chunkSplit: {strategy: 'all-in-one'}` -> top-level `splitChunks: false`. The web-component build keeps its `LimitChunkCountPlugin({maxChunks: 1})`, which is a cross-origin embed contract, not a size tweak.

## 3. `test: migrate from Jest to Rstest`

Rstest compiles with the same Rspack/swc toolchain the app already bundles with, so the runner shares one dependency tree instead of dragging in a parallel one. (Vitest was the other candidate; it hard-depends on Vite, which would have entrenched a second bundler -- see the `simlin-serve` follow-up below.)

**-260 packages** from the lockfile. The only jest-family package left is a transitive `jest-worker`.

**Test counts are unchanged, package for package:** core 100, engine 442 + 1 skipped, server 139, app 99, diagram 1500, serve-web 113.

Non-obvious bits, all documented in `docs/dev/typescript.md`:

- **Globals are off.** Every test imports what it uses from `@rstest/core`. That retires `@types/jest` and the shared eslint config's `jest: 'readonly'` block.
- `Mock<T>`'s type argument is the whole *function signature*, not the return type. `@types/jest` typed `jest.fn()` loosely enough to hide argument mismatches -- serve-web's `ws.test.ts` had been passing `<void, [WsMessage]>` all along, and it type-checks now.
- `jest.requireActual` is synchronous; `rs.importActual` is not, and Rstest rejects async mock factories outright. The replacement is the `import ... with { rstest: 'importActual' }` attribute, which hoists along with the factory. `src/server` cannot use it (import attributes need an ES module target, and that program emits CommonJS while type-checking its own tests), so `static-config.test.ts` uses `rs.mock('fs', { spy: true })`.
- Rstest has no `@jest-environment` docblock. The 14 diagram tests that opted into a Node environment are now an explicit `projects` entry -- `project-controller.ts` documents a "zero React, zero DOM" invariant that an ambient jsdom global would quietly erode. The other 39 docblocks existed only to carry the pragma, so those files' block-comment copyright headers return to the canonical line-comment form.
- `@testing-library/react` self-registers `afterEach(cleanup)` only when `afterEach` is a global, and `@testing-library/dom`'s `waitFor` only pumps fake timers when a global `jest` object exists (`helpers.js` gates on `typeof jest !== 'undefined'`, then calls `jest.advanceTimersByTime`). Both are wired up explicitly in a per-package `tests/setup-testing-library.ts`; the timer shim is inert under real timers.

Two Jest workarounds are **deleted rather than ported**:

- The wouter stub and its `moduleNameMapper` existed because CommonJS Jest mishandled wouter's ESM-only `exports`. Rspack resolves it natively, which also retires the `require.resolve()` dance that guarded #523.
- The CSS-module Proxy stubs (which echoed each property name back as its class name) give way to compiling the real stylesheets with `output.cssModules.localIdentName: '[local]'`, so tests that query `.eqnEditor` still match -- against real CSS now.

### One behavior change worth a close look

jsdom goes 26 -> 29, which implements `PointerEvent`. `dialog.test.tsx` asserted that a bare `pointerdown` dismisses a Radix dialog. That was only true because `@testing-library/dom` fell back to a plain `Event` whose `button` was `undefined`; given a real `PointerEvent` (`button: 0`), Radix waits for the matching `click`, as browsers do. The test now fires the full pointerdown/pointerup/click sequence, which is what a real pointer produces. **The test was passing for the wrong reason.**

## Discovered along the way

- **#899 -- test files are never type-checked.** `isolatedModules: true` in `tsconfig.base.json` made ts-jest transpile-only for every package, so this migration loses nothing, but the hole is now explicit: 55 latent type errors, including `src/diagram/tests/editor-applyPatch.test.ts` importing from `../json-types`, **a module that does not exist**. The issue carries a working `tsconfig.tests.json` and the full error list. (It also subsumes tech-debt entry 57, whose "fix" left the real bug in place -- because nothing type-checks it.)
- **Does not close #635** (`server: @simlin/server Jest suite intermittently fails one test under parallel pre-commit load`). The Jest suite named in that issue is gone, but its failing test was never identified and the likely root causes (shared fixture state, tight timeouts under contention) are runner-agnostic. It should be retitled rather than closed.
- **Partially addresses #895.** Deleting the jest globals block removed the last config that made `createConfig`'s index arithmetic (`configs[configs.length - (options.react ? 3 : 2)]`) necessary, so defect (c) is gone -- `options.project` is now assigned through a direct `baseConfig` reference. Defects (a) (module-level object shared across calls) and (b) (mutating the caller's `ignorePatterns` array) are untouched, so **#895 stays open**.
- `src/app/scripts/test.js` was an orphaned create-react-app launcher that `require`d jest and that nothing invoked. Deleted.
- **#901** filed for porting `simlin-serve/web` off Vite onto Rsbuild, so the repo carries one bundler.

## Verification

- `pnpm test`: 2393 passed, 1 skipped, exit 0 -- matching Jest package for package.
- `pnpm tsc`, `pnpm lint`, `scripts/lint-project.sh`, `check-copyright.py`: clean.
- Full pre-commit hook (Rust fmt/clippy/test, TS lint/build/tsc/test, pysimlin): passed.


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

PR #970, opened 2026-07-17
app: use popup OAuth flow so Google/Apple sign-in works on Safari

bpowers:
Google sign-in on app.simlin.com worked in Chrome but silently failed on iOS Safari: after returning from Google, the user was dropped back on the login screen with no error.

Root cause: `signInWithRedirect`'s result pickup loads an `auth.simlin.com` iframe that must read storage the sign-in helper wrote while it was the top-level page. WebKit partitions iframe storage per-origin, while Chromium partitions per-site, so the same-site custom authDomain (`auth.simlin.com` under `app.simlin.com`) only rescues Chromium-family browsers. On Safari the iframe sees an empty partition and `getRedirectResult` reports null with no error anywhere (firebase/firebase-js-sdk#7824; WebKit's per-origin partitioning is confirmed in https://webkit.org/blog/14403/updates-to-storage-policy/ and https://developer.apple.com/forums/thread/725074).

Fix: use `signInWithPopup`, which delivers the result to the opener via postMessage and needs no storage shared with the auth domain, so it works under per-origin partitioning. `signInWithRedirect` remains only as a fallback when a popup blocker rejects the popup (the provider buttons are direct user gestures, so blocking is rare); a user closing the popup is treated as a cancel, not an error. A side effect is that OAuth should now also complete on cross-site canary hosts, which the redirect flow structurally never could -- docs/dev/deploy.md is updated to reflect the new flow and the remaining canary caveat.

Verified end-to-end against the dev stack (auth emulator): popup opens with `authType=signInViaPopup`, sign-in completes, the `/session` exchange succeeds, and the new-user dialog renders. Real-device Safari verification still needs to happen on app.simlin.com after this deploys.

https://claude.ai/code/session_01EQoRGRGfkerDZapRpyrgAJ


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

Issue #984, opened 2026-07-20
ltm: the ceteris-paribus wrap leaves a LOOKUP table argument's runtime index live in every link-score partial


bpowers:
## Problem

`wrap_non_matching_in_previous` (`src/simlin-engine/src/ltm_augment.rs`, the `LOOKUP` / `LOOKUP_FORWARD` / `LOOKUP_BACKWARD` arm around line 685) holds the TABLE argument (child 0) **verbatim** and only wraps the index argument(s).

Holding the table *reference* verbatim is correct and deliberate: a graphical-function table is static data with no value slot, so `lookup(PREVIOUS(table), ...)` cannot compile, and wrapping it silently zeroes the whole link-score fragment. That is the failure mode behind WRLD3's identically-zero table-mediated link scores, and the in-code comment says so.

The defect is that the arm holds the **whole argument** verbatim, **including any subscript index expressions inside it** -- and those are genuine runtime value reads:

```
LOOKUP(g[idx], x)                  -- `idx` is a variable read
LOOKUP(pop[Region, other], x)      -- likewise
```

`compiler::codegen::extract_table_info` (`src/simlin-engine/src/compiler/codegen.rs:610`) evaluates such an index to select the table element at runtime: its `Expr::Subscript` arm computes an element offset from the index expressions (`offset = sum(index_i * stride_i)`, built out of the live index `Expr`s), while its `Expr::Var` / `Expr::StaticSubscript` arms resolve a static one to a constant.

So for a target whose equation contains `LOOKUP(g[idx], x)`, **every** link-score partial of that target varies with the *current-step* value of `idx`, even though ceteris paribus requires everything except the live source occurrence to be frozen. The score misattributes `idx`'s influence to whatever source is being isolated.

## Scope

This affects **all** link-score partials, not just the `PerElement` ones. For the non-`PerElement` callers the wrap's `ctx.pin` is `None`, so the table argument is returned completely untouched (the `None => a` branch) -- no pin descent, no freeze, nothing.

It predates branch `roundtrips-track-a2`.

## How it was found

[...]

Issue #983, opened 2026-07-20
ltm: emit_source_to_agg_link_scores re-parses and re-lowers an agg equation just to reach classify_reducer


bpowers:
## Problem

`emit_source_to_agg_link_scores` (`src/simlin-engine/src/db/ltm/link_scores.rs`, around line 2860) re-parses AND re-lowers an aggregate node's *printed equation text* through the entire front-end pipeline, purely to recover facts it already had.

The call site:

```rust
let agg_eqn = if agg.result_dims.is_empty() {
    LtmEquation::scalar(agg.equation_text.clone())
} else {
    LtmEquation::apply_to_all(agg.result_dims.clone(), agg.equation_text.clone())
};
let Some(agg_var) = reconstruct_ltm_var_lowered(db, &agg.name, &agg_eqn, model, project) else {
    return;
};
let Some(classified) = crate::ltm_augment::classify_reducer(&agg_var, from) else {
    return;
};
```

`reconstruct_ltm_var_lowered` (`src/simlin-engine/src/db/ltm/parse.rs:126`) runs the full boilerplate: `parse_ltm_equation_for_model_with_ids` (text -> `Expr0`, with module-ident and model-var-name scoping and implicit-module instantiation), then `crate::model::lower_variable` (`Expr0` -> `Expr2`) against a freshly built `ScopeStage0`. All of that so `classify_reducer` (`src/simlin-engine/src/ltm_augment.rs:4335`) can walk the resulting `Expr2` and hand back a `ClassifiedReducer`, of which the emitter consumes:

- `classified.kind` (`ReducerKind`) -- checked against `ReducerKind::Constant`, and passed to the partial generators;
- `classified.name` (`"SUM"` / `"MEAN"` / `"STDDEV"` / `"RANK"` / ...);
- `classified.body` (the reducer's array-argument AST), fed into `reducer_body_ctx_parts` / `ReducerBodyCtx`.

All three are already known upstream:

- `ltm_agg::enumerate_agg_nodes` (`src/simlin-engine/src/ltm_agg.rs:593`) is what *decided* this hoist in the first place, and it classified the reducer to do so.
- `ltm_agg::reducer_kind` (`src/simlin-engine/src/ltm_agg.rs:221`) is the authoritative classifier; `classify_reducer` is documented as a thin reader over the same fact.
[...]

Issue #982, opened 2026-07-20
ltm: two different reducer sets answer "is this reference inside a reducer?" (SIZE/RANK disagree between the #779 decline and OccurrenceSite::in_reducer)


bpowers:
## Summary

Two adjacent LTM decisions each answer "is this reference inside an array reducer?", and each answers it with a **different reducer set**. They disagree on exactly two builtins, `SIZE` and `RANK`. Neither choice is a bug where it lives -- both are documented and defended in their own rustdoc -- but they are two implementations of one question, which is the same drift surface #965 exists to remove.

## The two sets

1. **`ltm_augment::references_bare_source_inside_reducer`** (`src/simlin-engine/src/ltm_augment.rs:1404`) -- the GH #779 bare-reducer-feeder decline. It marks a child in-reducer via `ltm_agg::reducer_collapses_to_scalar`, which **includes `SIZE`** and **excludes `RANK`** (`ltm_agg.rs:211`: `reducer_kind_from_name(name, arity).is_some() && name != "rank"`).

2. **`db::ltm_ir`'s occurrence walker** -- sets `OccurrenceSite::in_reducer` (`src/simlin-engine/src/db/ltm_ir.rs:1276`) via `ltm_agg::builtin_routes_through_agg` (`ltm_agg.rs:259`: `reducer_is_hoistable(builtin) || matches!(builtin, BuiltinFn::Rank(_, _))`), which **excludes `SIZE`** (`SIZE` is `ReducerKind::Constant`, never hoisted) and **includes `RANK`** (array-valued, but still aggregate-routed per GH #776).

So on `SIZE` and `RANK` the two answers are exactly inverted.

## Concrete consequence

`OccurrenceSite::in_reducer` is the natural IR-side signal for the #779 decline -- it is the same fact, already computed by the one occurrence walk the typed-IR work (#965) established. Consuming it there would be the right simplification, and it is exactly what the occurrence IR is for. But it is **not behavior-neutral today**:

- **`RANK` is the blocker.** A bare arrayed source read inside `RANK(...)` currently classifies as *not* in a scalar reducer, so it is scored (via the GH #742 arrayed-capture path). If the decline consumed `in_reducer`, that same site would flip to "loudly declined as a bare reducer feeder" -- a real behavior change, with a real user-visible score/diagnostic difference.
[...]

Issue #981, opened 2026-07-20
ltm: a NaN literal in a scored equation defeats salsa backdating for typed LtmEquation (avoidable fragment recompiles)


bpowers:
## Problem

`LtmSyntheticVar::equation` is a typed `LtmEquation` whose arms carry a parsed `Expr0` AST (`src/simlin-engine/src/db/ltm/equation.rs`). `Expr0::Const(String, f64, Loc)` (`src/simlin-engine/src/ast/expr0.rs:76`) holds an `f64`, and `LtmArm` / `LtmEquation` get **derived** `PartialEq` and `salsa::Update`:

```rust
#[derive(Clone, PartialEq, salsa::Update)]
pub struct LtmArm { /* text: String, expr: Option<Expr0>, ... */ }
```

`nan` is a lexer keyword (`("nan", Nan)` in the KEYWORDS table, `src/simlin-engine/src/lexer/mod.rs:86`), so a generated LTM equation can legitimately embed `f64::NAN`. Since `NaN != NaN`, two arms built from **identical text** compare unequal.

Consequence: salsa cannot backdate the affected query result. `link_score_equation_text_shaped` re-runs and reports a changed value on every unrelated edit, so the expensive dependent `compile_ltm_var_fragment` (`src/simlin-engine/src/db/ltm/compile.rs`) re-executes even though the equation is semantically and textually identical. This is a performance/incrementality regression, not a wrong answer.

It also falsifies a claim in commit `dea42188` ("engine: store LTM synthetic equations as typed ASTs"), whose message states: "Salsa incrementality is preserved: the shaped text query backdates on the PartialEq-equal typed equation, so unrelated edits reuse the expensive fragment (pinned by the existing per-link caching test)." That is true for ordinary equations and false for NaN-bearing ones -- worth noting so the next reader does not trust that sentence unconditionally.

## Measured (verified by running code, not by inspection)

A throwaway probe constructing values through the public API (since removed, nothing committed):

- `LtmArm::new("NaN")` parses successfully (`expr.is_some() == true`).
- Two `LtmArm::new("NaN")` compare **unequal** (`a == b` is `false`).
- Two `LtmEquation::scalar("1 + NaN")` compare **unequal**.
- Control: two `LtmEquation::scalar("1 + 2")` compare **equal**.
[...]

PR #985, opened 2026-07-20
engine: delete the LTM input-side round trip and the last Expr0 classifiers

bpowers:
Part 2 of Track A, the follow-on to #980 (now merged as `24759a7e`; this branch has been rebased
onto `main`). #980 deleted LTM's `AST -> text -> AST` round trip on the
*output* side (generated equations stored as typed ASTs) and killed the classify-twice contract
that produced silent-zero link scores. This deletes the round trip on the **input** side: the
ceteris-paribus wrap no longer parses the target equation's printed text.

Part of #965. Does **not** close it — see "What remains" below.

## The result worth leading with

`expr2_to_string` is literally `print_eqn(expr2_to_expr0(e))`. So every one of the eight
`Expr0::new` sites in `ltm_augment.rs` was parsing *our own printed output of a tree we already
held in that exact shape*. Deleting them is "pass the `Expr0` instead of the `String`".

The consequence is stronger than the parse count. `db::ltm_ir` computes every occurrence's
`SiteId` on the `Expr2` tree; the wrap addresses occurrences by the child-index path it tracks as
it descends. Feeding the wrap the direct lowering makes those two path spaces **the same tree by
construction**. Before, they coincided only because print-then-parse happened to round-trip — a
property the corpus alignment sweep could *sample* but never *establish*. Given that five of the
six defects #980's review found were silent zeros arising from exactly this class of
"coincides in practice" assumption, moving it to by-construction is the point of the change.

Parse-site tally: **5 deleted, 1 relocated to a legitimate source-format boundary, 2 remain**
(with reasons, below). `ltm_augment.rs`: 5921 -> 5644 lines.

## Commits

| | |
|---|---|
| `8c7b4892` | lower LTM target equations to Expr0 without a reparse |
| `455001c9` | pin LTM per-element rows inside the wrap, not after it |
| `d6aadf8e` | delete the LTM occurrence fields nothing reads |
| `17a3ab27` | prove the LTM test occurrence axes against the IR |
| `2ac7c986` | carry LTM reducer bodies as ASTs, not printed text |
| `92615225` | decline loudly when an LTM row pin has no occurrence |
| `c4478627` | correct the LTM augmentation map in CLAUDE.md |
| `8b5df6b2` | pin LTM table-argument rows structurally, not loudly |
| `fc6c6ae0` | descend expression indices in the LTM pin-only walk |
| `1ea4f0f7` | decline an LTM table-argument index that reads at runtime |
| `afd1b92b` | gate the LTM table-index refusal on the enclosing freeze |
| `b429d4d8` | enumerate the LTM table-index verdicts, not just the found cases |
| `872e8f37` | resolve LTM table indices by the shared derivation |
| `4b295f52` | row the fifth LTM table-index variant too |
| `1375dad6` | look inside a compound LTM table index |
| `13aff4a5` | break an LTM index name collision the compiler's way |

## The crux: row pinning moved inside the wrap

`455001c9` is the load-bearing commit. Deleting the last two production Expr0-side classifiers
required relocating `PerElement` row pinning, and the obvious approach — a shape-preserving
pre-pass on the original AST — is **wrong**, which we established from the corpus rather than by
argument.

Row pinning emits two spellings of the same site shape: the occurrence the wrap left **live**
gets the row with bare element names; every occurrence it **froze** gets the same row qualified
(`region·boston`), so the frozen read compiles to a direct `LoadPrev` instead of an ambiguous bare
element name. That live-vs-frozen fact exists only in the *wrapped* tree, and it is not derivable
from the occurrence — two occurrences with identical `axes` can land on opposite sides.
`per_element_index_nested_occurrence.txt` already pins both spellings inside one equation. On a
model whose element names are ambiguous across dimensions — C-LEARN's regions — the bare spelling
**fails to compile**, so a pre-pass's divergence would be a silent zero, not cosmetic.

So the pin runs *inside* the wrap, which is the one traversal that knows both the occurrence (by
path) and whether it is about to freeze. This **reverses a documented decision** made in #980
(the "drive LTM live-ref selection from the occurrence IR" commit, squashed into `24759a7e`),
whose rustdoc rejected wrap-time occurrence *tagging* — that needs a tag-only descent
or a `Loc`-keyed side map, i.e. a second representation with its own drift surface. A pin-only
descent writes no tag and carries no map; it reads the one IR by path and performs the actual
lowering. Different thing, admissible for that reason.

Pinned narrowly rather than incidentally: mutating `child_frozen = frozen || will_wrap` down to
`frozen` — the exact information loss a pre-pass suffers — fails **precisely one** golden, the one
named in advance. Ignoring live/frozen entirely fails eight.

## One classifier family, and the surviving mirror is gated

`classify_expr0_per_element_axes` and `expr0_iterated_axis_lines_up` are out of production. Per-axis
equivalence is exact, not approximate: the retired classifier consulted the same
`ltm_agg::classify_axis_access` the IR uses, gate for gate.

A `#[cfg(test)]` Expr0 mirror survives, moved beside its only consumer in
`ltm_augment_wrap_test_support.rs`. Full deletion is not achievable: three wrap unit tests cannot
be db-backed fixtures — two feed deliberately unparseable and empty text (they pin the loud
parse-failure contract), and the third is `SUM(w[from]) + from`, which the engine **rejects as a
model** (`ArrayReferenceNeedsExplicitSubscripts`) yet still needs an occurrence stream.

Per the plan's own rule — *mirrors are acceptable when drift is loudly gated; where one must stay,
the gate is the deliverable* — `17a3ab27` extends the corpus gate from comparing `shape` to
comparing **`axes`** as well, so it now proves every field of the reconstruction the production
wrap reads (`SiteId` path, `shape`, `axes`, `in_reducer`). An `IndexExpr0 -> IndexExpr2` lift was
rejected: it would add a mirror faithful only to the forms `classify_axis_access` happens to
inspect, in order to delete a mirror.

## `OccurrenceSite`: four fields deleted, one kept

`target_element`, `routing`, `reducer_keys`, `already_lagged` had no production reader. The
strongest case is `routing`, which was not merely unread but a **duplicated derivation**:
`model_ltm_reference_sites` held two copies of the #793 agg-narrowing rule, and deleting the field
deletes one of them. Duplicate logic inside the IR built to remove duplicate logic.

`in_reducer` stays — its reader is the corpus gate's reducer-context assertion, and deleting it
would silently narrow that gate. Its rustdoc says so explicitly, and records why it is **not**
consumed by the #779 bare-feeder decline: `reducer_collapses_to_scalar` includes `SIZE` and
excludes `RANK` while `builtin_routes_through_agg` does the opposite, so wiring it up would flip a
bare source inside `RANK(...)` from scored to loudly declined. Filed as **#982**.

Deleting `already_lagged` also removed a parameter still threaded through every recursive call of
`walk_all_in_expr` computing only itself — dead code the compiler cannot see, because it is
genuinely "used" at each hop.

## A regression introduced and fixed inside this branch

`455001c9` introduced a silent zero, and the story is disclosed rather than tidied because the way
it was found matters.

A `PerElement` source occurrence inside a `LOOKUP` **table argument** came out un-pinned:
`lookup(pop[region, young], ...)`. That dimension-name subscript cannot resolve in a scalar
fragment, so the fragment is dropped, the variable keeps a layout slot with no bytecode, and the
score reads a constant 0. The cause is a genuine tension: `db::ltm_ir` records no occurrence under
a table argument (*"static data, not a causal edge"*) and it is **right** — but the pin such a
reference needs is about **compilability, not attribution**. The pre-`455001c9` pass pinned it
anyway because it re-classified the wrapped tree and so needed no occurrence. The IR-driven
pinning correctly refused to guess, and then said nothing, which was the defect.

`92615225` made it loud. **That was the wrong fix**, on a premise that turned out false: the
claim was that a table head can only be a lookup-only variable (no value slot, never a causal
source) or the WITH-LOOKUP self-reference. In fact `compiler/codegen.rs::extract_table_info`
resolves a table argument by ident against *any* table-carrying variable and has an explicit
`Expr::StaticSubscript` arm deliberately supporting `LOOKUP(g[<single element>], x)`; and
`var_is_lookup_only` requires an empty/sentinel equation, so a table-carrying variable *with* an
equation is value-bearing and can be a causal source. A review pass then produced the case that
shows the blast radius:

```
target[Region] = effect[Region,young] + LOOKUP(effect[Region,old], TIME)
```

That edge is scoreable from the first term alone — but because the same target reads `effect` in a
table argument, the whole partial returned `UnfreezablePartial` and **every per-element score for
the edge was dropped**.

`8b5df6b2` replaces the decline with a structural pin (`pin_dimension_name_indices`): an index
naming one of the source's own declared dimensions becomes this element's coordinate, and a literal
element index is qualified by its axis. It consults **no occurrence** — enforced by its signature,
which takes only indices and a context — infers no shape, makes no live/frozen decision, and
cannot make a table argument live-selectable. The loud net is kept behind it: an index naming a
*different* dimension still sets `missing_occurrence`, because resolving that needs the mapping
classifier the IR owns. `fc6c6ae0` completes it — a table reference may carry a *dynamic* element
index (`LOOKUP(pop[Region, pop[Region,young]], x)`), which `extract_table_info` computes an offset
for on purpose, so the reference inside the index needs the same pin. The `IndexExpr0::Expr` arm
the closure's own comment already claimed to have.

**Then two more review passes each found a defect in the previous fix**, and the end state is worth
more than the individual fixes. `fc6c6ae0` made the shape *compile* without making it *correct*:
the table argument is the only one of the three pin-only descents **not inside a freeze** — the
wrap holds it verbatim precisely because a `PREVIOUS` of a table has no value slot — so a runtime
index left live there stays live in the emitted partial, and a sibling occurrence's partial then
varies with it. `1ea4f0f7` declined runtime indices; `afd1b92b` then found *that* over-declined,
because a `LOOKUP` can itself sit inside a `PREVIOUS`/`INIT` or a whole-frozen reducer, where the
index is already lagged and refusing drops a valid score.

The resolution is a three-way split, which is the real shape of the problem:

| verdict | question it answers | behavior |
|---|---|---|
| pin | a **lowering** | resolve the index to this element's coordinate |
| `Unspellable` | about the **fragment** (compilability) | loud everywhere — no freeze can make a dimension-name subscript compile |
| `RuntimeRead` | about the **freeze** (ceteris paribus) | loud in a bare table argument, kept inside a freeze |

Both earlier defects were the same mistake: one predicate trying to answer two independent
questions. A lowering has two obligations — it must compile **and** it must preserve ceteris
paribus — and the pin-only descent satisfies the second only because the enclosing node freezes
everything. The table argument is the one site where that premise fails, which is exactly where
both defects landed. That premise is now in `pin_dimension_name_indices`' rustdoc in those terms.

## The verdict enumeration

`b429d4d8` adds `per_element_pin_index_verdict_enumeration`: **13 index kinds × {bare table
argument, inside a freeze} = 26 cells**, each stating the expected verdict. Every cell matched the
implementation on the first run, so it is test-only.

It exists because both of the findings above were **cells** — a shape in the wrong bucket — not
logic errors, and each was supplied by a reviewer's counterexample. An enumeration states a verdict
for every shape rather than the ones someone thought of.

The table's shape is itself the result: the four static rows and four unspellable rows are
**identical** across both columns, and **only the three runtime-read rows differ** — so the
compilability-vs-ceteris-paribus split is a visible property of the table rather than a claim in a
comment. Three of the unspellable rows were not previously tested at all (an own dimension at a
position the target doesn't project, transposed own dimensions, and an over-arity index).

It is a gate, not documentation: four mutations of the verdict logic fail it **independently** of
the seven per-shape tests, so an edit that moves a cell goes red even without breaking a named
shape test. Non-vacuity is asserted **per cell** — every case re-checks that the IR records nothing
for the table argument — so no cell can drift into testing the occurrence walker instead of the rule.

**Two cells are marked uncertain rather than asserted.** A **range** and a **wildcard** table index
are unreachable from a compilable model: codegen's `extract_table_info` rejects them with
`BadTable`, so the *target's own* equation fails to compile before its link scores exist. The rule
currently accepts them when frozen. If they ever became reachable, the failure would surface as a
fragment-compile Warning rather than a warned skip — loud either way, different channel. They were
left marked rather than guessed into `Unspellable`, because no reachable shape distinguishes the two
choices and a speculative production change was not worth the blast radius. One further caveat: the
**numeric literal** cell asserts the rule's verdict (leave a static selector alone), not downstream
compilability — a numeric index into a *named* dimension does not resolve, but that is a property of
the target's own equation, which carries the same index.

## Verification

- **C-LEARN numeric gate**, green on every production HEAD of the LOOKUP arc: **236.39s** (`fc6c6ae0`), **236.73s** (`1ea4f0f7`), **236.72s** (`afd1b92b`) — three runs within 0.4s of each other. Also green at 242.30s on `455001c9`, the commit that could falsify the pinning design, on the model whose element names are ambiguous across dimensions. Used as a negative control as well as a smoke test: had any real C-LEARN partial contained a runtime table index, the refusal in `1ea4f0f7` would have dropped its score and the pinned-loop assertion would have gone to zero. It didn't, so the tightening is narrow in practice.
- `cargo test -p simlin-engine --lib`: **5280 passed / 0 failed / 3 ignored**; integration **637 passed / 0 failed**.
- **All 16 char goldens byte-identical at every commit. Zero golden regenerations on this branch.**
- `scripts/lint-project.sh`, `cargo clippy --all-targets`, and `cargo check -p simlin-engine --lib --no-default-features` all clean; full pre-commit hook green on all 9 commits.
- Every fix mutation-verified; probes restored byte-for-byte with the editor, never `git checkout`.

## What remains on #965

Criteria 1-4 are met. Criterion 5 is substantially met with one documented exception, so this does
not close the issue:

- **`subscript_idents_at_element`'s re-parse (site 2147) is byte-load-bearing.** The wrap constructs
  `PREVIOUS` uppercase; four goldens carry lowercase `previous(` *only* because print -> re-parse
  lowercases function names, while `arrayed_target_slot_scores.txt` carries the uppercase spelling
  on a path where no re-parse happens. Both spellings are pinned in the corpus today. Deleting the
  site therefore requires normalizing the generators to one spelling and recapturing those goldens —
  a case-only change, mechanically checkable, and the right end state since it makes print -> parse a
  genuine no-op. Deferred deliberately rather than done at the end of a long branch; the decision is
  made and the four conditions (case-fold-identical diff, one uniform spelling, semantic-inertness
  proof, recorded as the single intentional golden change) are recorded.
- **The two `AggNode` feeders (sites 1916/1981) are blocked on a design question, not effort.**
  `Expr0`'s `PartialEq` includes `Loc`, and `AggNode` is salsa-cached with its `equation_text`
  documented as *"the canonical (`Loc`-insensitive) key the node was deduped on"*. Storing an `Expr0`
  beside it makes the cache entry position-**sensitive** where the key is deliberately
  position-**in**sensitive — an incremental-invalidation change. Decide explicitly whether to store
  it loc-stripped.
- Generation-side guard forms stay `format!` templates by choice. Converting them to AST builders
  would re-parenthesise and re-space every emitted equation, churning all 16 goldens to produce
  equivalent text — trading the corpus's entire value for an aesthetic.

## Filed from this work

- **#983** — `emit_source_to_agg_link_scores` re-parses *and re-lowers* an agg's equation text through the whole pipeline just to reach `classify_reducer` for a kind and a name. A deeper round trip than any of the eight addressed here. Two call sites (the second in `db/ltm/loops.rs` needs the whole lowered AST), and both fallible steps `return` early, so its failure mode is a silent zero.
- **#982** — the `reducer_collapses_to_scalar` vs `builtin_routes_through_agg` divergence on `SIZE`/`RANK`.
- **#986** — `classify_axis_access` resolves a subscript index dimension-name-first, opposite to the compiler's element-first precedence. Disclosed above; not fixed here because it feeds four downstream consumers and can move goldens.
- **#984** — **a pre-existing defect of the same class as the one fixed here, disclosed rather than fixed.** The wrap's `LOOKUP` arm returns the table argument untouched for *every* caller, so `LOOKUP(g[idx], x)` leaves `idx` live in **any** link-score partial, not only the `PerElement` ones. It surfaced only *through* this branch's code, because this path is the one that now reaches compilability. Not fixed here deliberately: it predates the branch, the real fix (give the wrap's own index pass the table argument's indices rather than holding the whole argument verbatim) touches every link score and can move goldens, and "hold the table argument verbatim" is a deliberate decision that deserves its own diff, review passes, and numeric gate rather than riding along in a regression fix.

## The table-argument rule: six findings, and what closed the class

The `LOOKUP` table-argument pin absorbed six review findings. Reported in full because the shape of
the iteration is the useful part, and because a reviewer should know which residual remains.

The findings, in order: not pinned at all (silent zero); a runtime index left live (misattribution);
over-declining inside a freeze; `@N` and a mapped dimension name misfiled as runtime reads; a missing
`StarRange` row (self-found); a constant-expression index (`1 + 1`) misfiled; and subscript
name-collision precedence. Every one after the second was a **cell** — a shape in the wrong bucket —
not a flaw in the design.

Three structural changes closed the class, rather than the individual cells:

1. **The catch-all is gone.** Unrecognized shapes were being silently filed as `RuntimeRead`. The
   match over `Expr0` is now exhaustive with no catch-all, so a new variant is a **compile error**
   instead of a silent refusal.
2. **Rows are derived from the type, not from what reviewers found.** `IndexExpr0` has exactly five
   variants; the table now names all five, which makes "no gaps" checkable. Subscript arity beyond
   the over-arity row is a *documented non-row*: `codegen::extract_table_info` accepts a table
   argument only when the subscript selects exactly one element, so an under-arity subscript is
   `BadTable` before link scores exist, and rowing it would state a verdict about a shape no
   compilable model can produce.
3. **Precedence is anchored to the compiler.** See the thesis note below.

**Residual risk, stated plainly:** this rule may still loudly refuse a statically-resolvable index no
row covers. The consequence is a **warned skip and a dropped score, never a wrong number.** The
silent direction — a runtime read left live, which misattributes — is guarded by the `RuntimeRead`
verdict plus the freeze gate and is pinned in both directions by mutation. Two deliberate boundaries
are stated rather than hidden: a 0-arity builtin index stays loud (telling `TIME` from `PI` inside
`App` would be a fourth copy of the builtin classification that `builtins`/`compiler::invariance`
own), and three unreachable rows (range, wildcard, star-range) are marked uncertain for one traced
reason rather than three hand-waves.

**Verdict enumeration: 3 tables, 52 cells.** Main 18 rows × 2 freeze columns, mapped 6 row-runs × 2
(3 rows across both declaration directions), collision 2 × 2. Ten mutation probes, all caught, and
**seven of the ten are caught by rows that already existed** — the table doing the job a per-shape
test cannot. Every emitted cell also asserts the live site kept its bare row, so a fixture that is
not a genuine `PerElement` instantiation cannot pass. The collision table carries a **control row**
(the same axis's non-colliding element) so a fix that merely special-cased the colliding name fails
one of the two.

## Single-sourcing eliminates drift; it does not establish truth

This branch's thesis is that one derivation should serve every consumer. One finding sharpened it,
and the correction matters more than the fix.

Subscript index precedence was resolved dimension-name-first, on the stated grounds of *mirroring
`classify_axis_access`*. But XMILE lets a dimension declare an element whose name is also a
dimension name, and `compiler/subscript.rs`'s `normalize_subscripts3` is **element-first** — its own
comment says "takes priority". So the analysis-layer classifier disagrees with the compiler, and
"mirror the shared classifier" is what propagated the defect into the new rule.

A shared derivation that disagrees with the compiler is a single source of a **consistent error**.
The authority for what an equation *means* is the compiler; the correct instruction is not "match the
shared derivation" but "match the compiler, and share *that*." Fixed here by ordering the rule the
compiler's way. **`classify_axis_access` itself is left wrong and filed as #986** — it feeds reducer
hoisting, the reference-shape IR, element-edge emission and link-score naming, so reordering it can
move goldens and needs its own diff and verification. Today its dimension-first reading is
*accidentally* safe (it finds no mapping, declines, and falls to the conservative path); #986 records
that the collision-plus-mapping case is a genuine wrong-answer hazard there, not mere imprecision.

## What C-LEARN does and does not gate here

Worth being precise, because a green numeric gate is easy to overclaim. `clearn_pinned_climate_loop_is_scored`
asserts Discovery mode, three partition entries, and that the deterministic series is finite,
not identically zero, and negative where non-zero. It asserts **no magnitudes and no warning count**.

So for a *tightening* it is a strong control — a refusal that dropped a real score zeroes the series
and trips the non-zero assertion. For a *widening* it is nearly blind, and three of the four fixes
here widen what gets scored.

What actually makes it inert is structural, not luck: C-LEARN's 199 subscripted lookup-table calls
(`Historical forestry LOOKUP[COP](…)` and siblings) all use **literal element** indices — the one
index form this arc never changed — and those heads are table-only holders, excluded from every
runlist, so they cannot be causal sources and the rule is never reached for them. Five C-LEARN runs
across the arc landed in a 2.3s band (235.78 / 236.39 / 236.72 / 236.73 / 238.03s), all green.

The honest conclusion: **C-LEARN does not exercise the shapes these four fixes changed.** Their
coverage is the three new end-to-end tests (`@N`, mapped dimension, name collision), each of which
compiles a real model and asserts zero assembly warnings plus a finite non-zero score. C-LEARN's role
here is a consistency check that nothing *else* moved.

## Method note

Worth recording honestly, because it is the reusable part.

Stage `455001c9` created **three** pin-only descents (a pre-existing `PREVIOUS`/`INIT`, a
whole-frozen reducer, and a `LOOKUP` table argument). Two were probed; the third was not, and that
is where the regression lived. It surfaced only because a review pass left a probe test behind. A
later probe of the `PREVIOUS`/`INIT` descent then found that deleting it left **all 5274 lib and all
634 integration tests green** — zero coverage on a second descent. The rule: derive the probe count
from the branch's branch structure, not from when probing starts to feel sufficient.

Four test-scaffolding divergences surfaced across this work, every one by mutation rather than by
review:

1. the whole-reducer descent had no coverage;
2. the `cfg(test)` occurrence builder synthesized only `Pinned`/`Dynamic` and never `Iterated`, so those tests leaned on the classifier rather than the occurrence stream they claimed to build;
3. it had no `MismatchedIterated` arm, so a non-aligning dimension name came out `Dynamic` — invisible because every consumer treats the two identically;
4. `expr_reference_idents` documented "returns an empty set when the text does not parse", but that result *is* the freeze set — an empty one emits a partial that freezes nothing, i.e. the target's own equation, scoring a constant `|Δz/Δz| = 1`. The #311 hazard, behind a docstring calling the fallback benign.

Review passes used `--base origin/roundtrips-track-a` so each read this branch's commits rather
than re-reading #980's 14; the #981 finding in `equation.rs` belongs to #980's stack and is
disclosed there.

**Review is a sampling process, and this branch measured it rather than assuming it.** Six passes:

| pass | result |
|---|---|
| 1 | 1 × P2 — the `LOOKUP` pin regression |
| 2 | clean |
| 3 | 1 × P2 — runtime index left live (in the tree pass 2 had just read clean) |
| 4 | 1 × P2 — the previous fix over-declined inside freezes |
| 5 | clean |
| 6 | clean |

Green here required **two consecutive completed clean passes**, and that bar is justified by
measurement, not argument: **pass 2 read the tree clean and pass 3 found a P2 in it.** A single
clean pass in this area has been demonstrated worthless once. (The same thing happened on #980,
where pass 1 read the typed-equation-storage commit clean and pass 3 found a P2 in that same file.) An aborted or
quota-blocked pass counted as neither clean nor dirty — one pass was verified as a genuine
completion via `turn.completed` and exit code rather than by grepping for a quota string, since the
obvious pattern `quota` false-matched this branch's own test name `unquotable_source_name`.


bpowers:
Review pass 7 (over the verdict enumeration itself) found two P2s, both in `pin_dimension_name_indices`. The PR body's claim that the enumeration "came back all-green" is accurate as to what was run and **incomplete as to what was true** — being corrected:

- `IndexExpr0::DimPosition` (`@N`) falls into the catch-all and is classified `RuntimeRead`, so a bare table argument declines and drops every score for the edge — though the compiler resolves `@N` to a static offset. The enumeration's 13 index kinds did not include `DimPosition`.
- A **mapped** dimension name is classified `Unspellable` solely because it differs from the source's axis name, so `effect[State, young] + LOOKUP(effect[State, old], time)` under a positional `State -> Region` mapping drops the whole edge. The retired mapped-aware classifier handled this.

Root cause is shared: the rule hand-rolls per-axis name resolution that already exists in single-source form (`ltm_agg::classify_axis_access`, `DimensionsContext::mapped_element_correspondence`). Being fixed by consulting those rather than by adding arms, with the enumeration extended to cover `@N`, mapped pairs in both declaration directions, element-mapped pairs, and subdimensions.

Worth recording the methodological point: mutation testing proved the table **constrains** the code — four verdict mutations fail it independently of the per-shape tests — but it cannot show the table asserts the **right** verdicts. A cell matching the implementation because both are wrong is invisible to mutation by construction. That gap is exactly what this pass closed, and it is why the enumeration was reviewed rather than trusted.

Not ready to merge until two consecutive clean passes land, at least one after the enumeration extension.

Issue #716, opened 2026-06-04
ltm: scalar module output feeding an arrayed reader emits a single scalar constant-0 link score, dropping all arrayed loops through modules


bpowers:
## Summary

A SCALAR module output feeding an ARRAYED reader gets its `input -> module` (really `module-output -> arrayed-reader`) link score emitted as a single **scalar** `$⁚ltm⁚link_score⁚{m}→{growth}` variable that evaluates to a constant **0**. There are NO per-target-element `$⁚ltm⁚link_score⁚{m}→{growth}[{elem}]` variables. As a result, every arrayed feedback loop that runs through a module is dropped by LTM discovery (the loop product is 0, so the loop never forms) and mis-scored in exhaustive mode. This is an **arrayed-link-score *emission* gap for the module-output-as-source case**, not a settled equilibrium.

## Empirical evidence

Probed on an arrayed fixture: arrayed stock `s[Region]`; a scalar multi-output module `m` (pos/neg producer); per-element reader `growth[Region] = f(m·pos)`; loop closes `s[e] -> m -> growth[e] -> s[e]`.

- The element-level causal graph correctly carries the subscripted edges (`s[nyc] -> m`, `s[boston] -> m`; exit reader `growth[nyc]`).
- But the `m -> growth` link score is emitted as a single scalar `$⁚ltm⁚link_score⁚m→growth` whose value is constant **0**. There are NO `$⁚ltm⁚link_score⁚m→growth[nyc]` per-element variables.
- The module internals are demonstrably alive (`m·$⁚ltm⁚composite⁚input_val = -1`, internal `input_val→pos = +1`, `input_val→neg = -1`), so the `0` is an emission gap, not an equilibrium.
- Consequence: discovery returns `loops: []` for the arrayed loop `s[e] -> m -> growth[e] -> s[e]`. The per-exit-port recompute (the GH #698 family, PR #705) can never apply to arrayed loops through a module because they never form.

## Root cause direction

[...]

PR #994, opened 2026-07-30
engine: root-cause and fix the silent LTM fragment-compile failures

bpowers:
On C-LEARN, 1,603 generated LTM fragments failed to compile. Each kept a layout slot with no bytecode and read a constant 0, so a link or loop score derived from it was silently degraded. The warning said only *that* a fragment failed, never why, so the class had never been diagnosed.

This branch root-causes them and fixes what can be fixed correctly.

**LTM fragment-compile failures 1,603 → 449. Silent zeros (a score present but computed around a hole) → 0.**

## Why it was undiagnosable

Four distinct error paths collapsed into one `Option::None`: two legs of the LTM equation compiler, and `compile_phase_to_per_var_bytecodes` discarding its emit error with `.ok()?`. `d71a4f6a` makes the reason ride the warning and adds `examples/ltm_fragment_failures.rs`, which buckets failures by compiler error and cross-tabulates them against structure. It runs on any model via `LTM_FAIL_MODEL`.

With reasons available, the 1,554 remaining failures resolve into four generators — grouped by generating *root* rather than by fragment, so a defect is counted once rather than once per fragment it poisoned.

## What is fixed

**`485c0170` — stop freezing a dimension-name subscript index.** The ceteris-paribus wrap froze a bare dimension name used as a subscript index. Per-element substitution then turned it into a qualified element, which constant-folds, and codegen rejects `PREVIOUS(<literal>)`. A guard saying exactly that a dimension name must be left verbatim already existed but was gated on a condition that stopped being true once the wrap gained a resolved axis. One token. Beyond the 49 compile failures it clears, three *arrayed* link scores were compiling while their capture helpers were not — silently reading zero. Those are restored.

**`c6b2c76c` — pin a LOOKUP table's index, and decline what cannot be pinned.** A per-element partial is a scalar fragment, so every reference must select one element. The pin pass built its candidates from the dependency set, and a graphical-function holder is deliberately not in it. The dimensions were never unreachable — `referenced_tables` rides the same struct the same AST pass already returns. Widening the *pin candidates* with it is the fix; the wrap's dependency set is deliberately not widened, for an attribution reason.

That alone made 21 parents compile while a helper beneath them stayed dead, so a completeness guard now declines an edge loudly when an unresolvable index survives into the finished partial. It inspects the emitted partial rather than predicting from the pin table — predicting over-declined by 63 edges, caught by an existing test. The guard subsumes a second defect class entirely, and runs at all four sites that emit this object.

**`8a9c17e4` — attribute a codegen rejection to its variable.** A fragment failing in codegen produced *no diagnostic at all*; only lowering errors were reported. Six ordinary hand-written apply-to-all shapes fail this way today **with LTM disabled** and report nothing about why. Severity was measured, not argued: a sweep across the `test/` corpus found added rows land only on projects that already fail to compile.

## Corrections to premises that were wrong

Two claims in the codebase were false and had tests that appeared to pin them.

**`541c05a8`** — `mapped_element_correspondence` declines every explicit element map, justified by the claim that execution resolves mapped references positionally, ignoring the map. That is false as a universal. The discriminator is which dimension name the subscript spells; declaration direction is irrelevant. Map-following is **correct**, settled against real Vensim output: C-LEARN's `im_6_ff_co2[COP]` has seven elements each uniquely identifying their source region in the checked-in `Ref.vdf`, and it is the declared map every time. The gate nonetheless stays — deleting it drops a true causal edge. Recorded because nothing else can catch it: every caller is analysis code, `DimensionsContext` is private with no re-export, and `simulates_clearn` passes with the gate removed.

**`964cc276`** — a module-port reference is spelled as one fully-quoted identifier, and a module-hood lookup misses on it. Real, and **inert**: canonicalization strips the quotes and it resolves anyway. Recorded so nobody re-derives it as a bug, which is exactly what happened here.

## Verification

Every fix is test-first, with the pre-fix failure captured. Each piece was reviewed adversarially by an agent with no stake in it; findings were driven to zero before commit. Reviews caught, by mutation rather than by reading: a coverage claim that was false, a load-bearing justification naming a coincidence rather than the mechanism, a guard reachable on three unguarded sibling sites, and four vacuous assertions — including one whose fixture could not have detected a mis-pin *in principle*, because the score saturated.

Full suite green throughout (5,475 lib + 647 integration + 4 + 3), including `simulates_clearn`, which gates C-LEARN against real Vensim output.

## Scope disclosed rather than implied

Nothing here fixes classes A or B; they are tracked. Every C-LEARN occurrence of the pinned shape is the identity case, which is a measured property of one corpus and not of the language. The implicit-helper phase compiler still swallows its codegen error.

## What is not fixed, and where it is tracked

No closing keywords: this branch fully resolves no existing issue. What it leaves is measured and attributed rather than vague.

| remaining | roots | tracked |
|---|---|---|
| module -> arrayed target: `scalarize()` relabels shape without rewriting the body | 340 | #716 (this is the mechanism behind it; commented with the two code sites) |
| array-valued operand must be a view over storage, so `PREVIOUS()` of an array fails | 109 | #995 |
| the 135 declined edges that may be a pin gap rather than a compiler gap | -- | #996 (**check before #995**; could shrink it substantially) |
| the element-map spelling fix that would let class D be scored rather than declined | 70 | #997 |
| three ranking surfaces dominated by groups of size one | -- | #998 |
| `collect_all_diagnostics` not reproducible run-to-run | -- | #999 |
| the implicit-helper phase compiler still swallowing its codegen error | -- | #1000 |

#984 is **not** closed. It is commented with what moved and with a correction: its surrounding analysis held that a graphical-function holder's dimensions were unreachable, and they were not -- the constraint is absence from the dependency *graph*, which is a different thing. Its runtime-index half is untouched.

Suggested order: #996 first (cheap, and may make #995 much smaller), then #716, then #995.

### DIFF
--- a/notebooks/build_clearn_ltm_audit.py
+++ b/notebooks/build_clearn_ltm_audit.py
@@ -1,0 +1,1030 @@
+#!/usr/bin/env python3
+"""Build the C-LEARN Loops-That-Matter audit notebook.
+
+Constructs notebooks/clearn_ltm_audit.ipynb cell-by-cell via nbformat, then
+executes it (populating outputs) with nbclient. The executed .ipynb and its
+HTML render are generated artifacts (gitignored); this script is the source of
+truth, so the notebook can be regenerated against any engine build.
+
+Every number in the notebook is COMPUTED by a cell -- nothing is pasted in --
+so a regeneration against a newer engine either reproduces the findings or
+visibly contradicts them.
+
+Sibling generator: build_notebook.py builds the older `clearn_ltm_experience`
+notebook, written when GH #653 (pinned loops unscoreable at C-LEARN's scale)
+was still open; its central technique -- composing loop scores by hand from
+link scores -- is a workaround for a bug that is now fixed.
+
+Run from the pysimlin venv, synced with the `notebooks` extra. Invoke the venv
+interpreter DIRECTLY: `uv run` re-syncs the project as an editable install and
+replaces any wheel you installed.
+
+    cd src/pysimlin && uv sync --extra dev --extra notebooks
+    src/pysimlin/.venv/bin/python notebooks/build_clearn_ltm_audit.py
+"""
+
+from pathlib import Path
+
+import nbformat as nbf
+
+NOTEBOOKS_DIR = Path(__file__).resolve().parent
+NOTEBOOK_PATH = NOTEBOOKS_DIR / "clearn_ltm_audit.ipynb"
+
+nb = nbf.v4.new_notebook()
+nb.metadata["kernelspec"] = {
+    "display_name": "Python 3",
+    "language": "python",
+    "name": "python3",
+}
+
+cells: list = []
+
+
+def md(source: str) -> None:
+    cells.append(nbf.v4.new_markdown_cell(source.strip()))
+
+
+def code(source: str) -> None:
+    cells.append(nbf.v4.new_code_cell(source.strip()))
+
+
+# ---------------------------------------------------------------------------
+# 0. Provenance
+# ---------------------------------------------------------------------------
+
+md(
+    """
+# C-LEARN under Loops That Matter
+
+**An experience report: driving Simlin's LTM implementation through `pysimlin` on a
+real, large, arrayed climate-policy model.**
+
+C-LEARN v77 is a Vensim model of the global carbon cycle and climate coupled to an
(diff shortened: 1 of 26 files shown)


bpowers:
@codex review

Issue #995, opened 2026-07-30
engine: an array-valued operand must be a view over storage, so PREVIOUS() of an array fails to compile (affects ordinary A2A models, not just LTM)


bpowers:
## Summary

Codegen requires an array-valued operand of a vector builtin to be a **view over storage** (`StaticSubscript | TempArray | Var | Subscript`; `compiler/codegen.rs::walk_expr_as_view`, ~line 484-578). Any *computed* array must be materialised into a temp by an `AssignTemp` earlier in the same fragment. `PREVIOUS(<arrayed ref>)` is a computed array, and there is no materialiser for it — so it fails to compile.

**This is not an LTM problem.** These shapes fail at HEAD in ordinary hand-written apply-to-all models with LTM disabled:

| equation | result |
|---|---|
| `out[d] = VECTOR SORT ORDER(vals[d], 1)` | compiles |
| `out[d] = VECTOR ELM MAP(vals[e1], offsets[d])` | compiles |
| `out[d] = RANK(vals[*], 1)` | compiles |
| `out[d] = VECTOR SORT ORDER(PREVIOUS(vals[d]), 1)` | **fails** |
| `out[d] = VECTOR ELM MAP(PREVIOUS(vals[e1]), offsets[d])` | **fails** |
| `out[d] = VECTOR ELM MAP(vals[e1], PREVIOUS(offsets[d]))` | **fails** |
| `out[d] = VECTOR SORT ORDER(INIT(vals[d]), 1)` | **fails** |
| `out[d] = VECTOR SORT ORDER(vals[d] * 2, 1)` | **fails** |
| `out[d] = VECTOR SELECT(PREVIOUS(offsets[d]), vals[d], 0, 1, 0)` | **fails** |
| `out[d] = RANK(vals[*] * 2, 1)` | **fails** (separate arm, see below) |

Before #994 these reported **no diagnostic at all**; that PR makes a codegen rejection name its variable, so they now surface properly.

## Where the fix belongs

**Not in codegen.** Temp ids and `temp_sizes` are derived from the lowered `Vec<Expr>` (`extract_temp_sizes`, consumed by `db::assemble::compile_phase_to_per_var_bytecodes`), so allocation must happen at lowering.

[...]

Issue #915, opened 2026-07-09
test: nothing runs the simlin-engine `#[ignore]`d tier; 7 tests fail there today (one untracked)


bpowers:
## Problem

Nothing in pre-commit or CI ever runs the `simlin-engine` `#[ignore]`d test tier, and 7 tests in it fail today. An `#[ignore]`d *failing* test is a weak marker: it never runs, so nobody learns when a feature starts working, and nobody learns when one silently regresses.

Reproduced on a clean checkout of `conveyor-engine` HEAD (`c1cc86d5`, no local modifications, verified via `git stash`):

```
cargo test --release -p simlin-engine --features file_io --test integration -- --ignored
...
failures:
    simulate::simulates_delayfixed2_mdl
    simulate::simulates_delayfixed2_xmile
    simulate::simulates_delayfixed_mdl
    simulate::simulates_delayfixed_xmile
    simulate::simulates_except_xmile
    simulate::simulates_getdata_mdl
    simulate::simulates_getdata_xmile
test result: FAILED. 21 passed; 8 failed
```

(The 8th failure, `simulate::clearn_ltm_var_count_guardrail`, is a stale pinned count being handled separately and is out of scope here.)

All seven live in `src/simlin-engine/tests/integration/simulate.rs`.

## Triage of the seven failures

Six of the seven are genuine, already-tracked feature gaps:

| Tests | Tracked by |
|---|---|
| `simulates_delayfixed_xmile`, `simulates_delayfixed_mdl`, `simulates_delayfixed2_xmile`, `simulates_delayfixed2_mdl` | #346 (DELAY FIXED ring-buffer semantics) -- OPEN |
| `simulates_getdata_xmile`, `simulates_getdata_mdl` | #347 (GET DATA BETWEEN TIMES + implicit `.dat` loading) -- OPEN |

The seventh, **`simulates_except_xmile`, has no live tracking issue**. It was one of four tests listed in #345 ("engine: EXCEPT + cross-dimension mapping blocks 4 simulation tests"), which was **closed as completed on 2026-03-12**. The close was correct for three of the four -- `simulates_except`, `simulates_except2`, and `simulates_except_basic_mdl` are un-`#[ignore]`d and passing today (verified at `src/simlin-engine/tests/integration/simulate.rs:1516`, `:1521`, `:1534`). But `simulates_except_xmile` (`:1480`) was left `#[ignore]`d and still fails.

[...]

PR #1001, opened 2026-07-31
engine: bind per-element implicit instances to their element (C-LEARN dead-on-cycle edges 35 -> 0)

### LINKED ISSUE #716: ltm: scalar module output feeding an arrayed reader emits a single scalar constant-0 link score, dropping all arrayed loops through modules

bpowers:
An implicit instance minted by per-element expansion belongs to ONE element. Nothing in the LTM layer knew that, and three places paid for it.

## Why this, and not #995

Ranking the LTM residual by *fragment count* is the wrong denominator. A failing link-score fragment keeps its layout slot with no bytecode and reads a constant 0, and `IndexedSearch::load_step_scores` **drops a zero-scored edge from the discovery graph** — so a dead score on a feed-forward edge costs one edge's attribution, while a dead score inside a cycle makes every loop through it undiscoverable.

`examples/ltm_edge_coverage.rs` (added here) joins cycle-membership × score-state per edge on any model. On C-LEARN it showed **50 of 298 loop-carrying edges with no usable score, all of them one shape**.

This also reinterprets an earlier note that "0 of 153 discovered loops traverse a failing edge". That is not evidence the failures were harmless — a zero-scored edge is excluded from the search graph *by construction*, so it is the defect's signature.

## Results on C-LEARN

| | before | after |
|---|---|---|
| **dead-on-cycle edges** | 35 | **0** |
| feed-forward dead edges | 345 | 40 |
| failing fragments | 980 | 640 |
| element-graph edges | 22,224 | 15,826 |
| loop-carrying edges | 298 | 206 |
| score coverage on loop-carrying edges | 83% | **93%** |

Loop-carrying edges fell because 92 of them were *manufactured* by the broadcast. On a 2-element fixture the enumerator goes from 6 loops (4 phantom, all scores dead) to 2, both scored.

## The three fixes

**A constant subscript index now resolves POSITIONALLY** (`dimensions::resolve_axis_index_position`). `Expr1::constify_dimensions` turns a qualified `dimension·element` reference — the form `substitute_dimension_refs` synthesizes for every per-element expansion — into that element's 1-based index within the dimension it *names*, and `normalize_subscripts3` lowers it to a raw offset into the *subscripted* variable's own axis. Resolving it by name would describe a row the simulation never reads. Where two dimensions list the same names in different orders those readings diverge, and the test checks the graph against the VM rather than against an assumption. Both production describers move together. The same rule closes a sibling gap where `pop[2]` into a named dimension classified as `DynamicIndex`; one existing assertion changed accordingly, because it pinned a describer that disagreed with the runtime.

**A `module·port` read now records a `ClassifiedSite`**, so the edge carries its `target_element`. `emit_edges_for_reference` already narrows to a single target node when that is `Some` — the emitter was right and simply never received it, so a scalar instance broadcast across every element of its arrayed parent. Only `module → variable` edges gain an entry; the other two module families have no AST reference and keep the fallback.

**Module and helper sources feeding an arrayed reader get one link score per target element** (`try_implicit_scalar_to_arrayed_link_scores`). Every sibling emitter opens with `source_vars.get(from)?`, which an implicit instance can never satisfy — hence its own arm. The old path was not merely producing a zero: each instance but the first received *another element's* partial, holding the wrong instance live and freezing its own, so the compile failure was masking a wrong answer.

Plus a prerequisite fix in the second commit: an arrayed implicit helper now gets a dimension-aware LTM dep stub. `ImplicitVarMeta::dimensions` exists for exactly this and was consulted for LTM parse helpers but not for the model's own implicit vars. Pre-existing — the same fixture produces identical failures at the base commit.

## Verification

5,494 lib + 647 integration green; full pre-commit hook green. The `#[ignore]`d tier still has exactly its 7 pre-existing failures (#915), confirmed against a pristine worktree rather than assumed. `simulates_clearn_wasm` and `clearn_pinned_climate_loop_is_scored` pass.

One existing test flipped deliberately: `analysis.rs` carried a test *predicting* this change and asking for a reviewed update when #716 closed. It now asserts two per-element module loops and that neither crosses elements.

Two adversarial review passes ran against this branch, with mutation testing. They caught a wrong fix direction (the positional discovery above — my original plan was name-based and would have shipped edges the simulation never reads), a merged rustdoc block, a NaN guard contradicting its own doc, and four unpinned guards. All addressed; the surviving mutations are disclosed in the test module rather than left silent.

## Residual, disclosed rather than implied

- **15 loop-carrying edges still emit no score.** 14 are a loud `UnfreezablePartial` decline, not a missing emitter: the partial would freeze an array *slice* inside `PREVIOUS`, which has no codegen path. That is #995's contract, so **#995 does have loop-level impact — via the decline path, not the failure path.** This corrects the "entirely feed-forward" reading that came from classifying compile failures alone.
- Putting the element on the `to` side inherits a cost from `try_scalar_to_arrayed_link_scores`: a sub-model **pathway** equation resolves link-score names at variable level and cannot spell a per-element name, so it falls back to a name the fragment compiler stubs to 0. The fallback is deliberate; being silent about it was not, and it now warns (`emit_unresolved_pathway_link_warning`).
- `simlin_analyze_get_links` reports `score: None` rather than a fake `Some([0,0,…])` for these edges — same pre-existing class as scalar→arrayed.
- An `Ast::Arrayed` default-expression slot still broadcasts, because `target_element: Option<String>` cannot express "every element no explicit slot claimed".
- Two of the three `ModuleOutput` record sites are uncovered by tests and say so in the module doc.

## Next: the array freeze (#995), with its design

The remaining loop-carrying gap is one root cause. The clean fix is #995's option B — synthesize an **arrayed** freeze helper `h[D] = PREVIOUS(src[…, D])`, whose per-element body is `PREVIOUS(<static slot>)` and therefore compiles today, and reference `h` as a whole array. Because `h` is a variable, it is a view over storage by construction, which is exactly what codegen's array-operand contract requires. No new opcode, no VM change, no wasm change. The dep-stub fix in the second commit here is a prerequisite for it.

The blocker is placement, and it is worth stating precisely: helper synthesis happens at parse time in `builtins_visitor::make_temp_arg`, which has a `DimensionsContext` (dimension name → `Dimension`) and a variable *name* set, but **no variable → declared-dims map** — so it cannot name the axis behind a bare `*`. Two ways forward: thread a bounded variable→dims lookup into the visitor (weighing the parse-memo incrementality the `model_scope_models` narrowing exists to protect), or handle only axes that are already *nameable*. The second is not a hack — C-LEARN's blocked shape spells its axes as `*:set_targets`, so it would close all 14 edges with no architectural change, and a bare `*` would keep declining loudly.

Note this is not LTM-specific: `picked[Row] = VECTOR SELECT(sel[Row,*], PREVIOUS(matrix[Row,*]), …)` fails to compile today as an ordinary user equation.

Fixes #716


bpowers:
Worked through all five findings — every one was real, and four were in the coverage harness this branch's numbers are read from, which makes them worth more than their size. Fixed in 2fab04f.

**Duplicate pathway warnings** — real, and the worst of the five, because the warning I added to make a silent zero *loud* could itself have buried every other diagnostic. Now gated on a `(port, from, to)` set. Measured on a two-pathway fixture with a shared unresolvable edge: without the fix the shared edge warns 4× and a single-pathway edge 2×; with it, 2× and 2×. The test asserts *equality* rather than `== 1`, which isolates pathway multiplicity from a separate pre-existing ×2 (a sub-model's LTM diagnostics are collected once directly and once through the parent's subtree — the single-pathway edge doubling too is what shows that isn't multiplicity).

**Compile status ≠ liveness** — real, and the most valuable. The harness now compiles the LTM-enabled project, runs it, and records which link-score columns are non-zero at some saved step.

One correction to the suggestion, though: I kept the two measures as **separate axes** rather than replacing the static one. An identically-zero column is what `load_step_scores` drops, but it is *not by itself a defect* — an edge with no influence is correctly scored zero, and 2,279 feed-forward C-LEARN edges are in exactly that state. Merging the axes would relabel every genuinely-inert edge as broken. So the defect axis stays the headline, and the runtime axis contributes the population that actually matters for the hidden-stub case you describe: edges on a cycle that **compile yet are zero at every step** — 74 on C-LEARN, and inspecting them shows they're `INITIAL(...)`-captured helpers whose sensitivity is genuinely zero. That count is now printed rather than assumed.

**#716 vs #995 misclassification** — real. `PREVIOUS requires a variable reference` is emitted both by a module partial freezing an un-rewritten arrayed body and by any array-valued freeze. Discriminated on the edge's source: a synthetic instance carries the reserved `$⁚` prefix, an ordinary variable doesn't, and only the first is #716's shape.

**`failed.len()` overstating link-score failures** — real, and off by 396 (it counted loop scores, composites, pathways and helpers). Both numbers now printed and labelled: 244 link scores, 640 fragments of all kinds.

**Element-level granularity** — real. I did *not* move the join to element granularity, deliberately: mapping a link-score name to element endpoints means duplicating the expansion `parse_link_offsets` performs, and a second derivation to keep in step is worse than a stated limitation. Instead `partially dead` / `partially live` are their own states and are never folded into the dead-on-cycle totals, which makes those a sound **lower bound**; the limitation and the reasoning are on the module doc. On C-LEARN the ambiguous rows don't fire on the defect axis at all.

Re-measured the baseline at the base commit **with the reworked harness** to make sure the branch's headline numbers survive it: still 35 dead-on-cycle and 345 feed-forward dead there, against 0 and 40 here.

5,495 lib + 647 integration green; full pre-commit green; `#[ignore]`d tier unchanged at its 7 pre-existing failures (#915).

@codex review

bpowers:
All four were real. Fixed in 656a8f2.

**Per-element module port selection** — the important one, and a genuine defect in this branch's own emitter. You're right that `module_output_ref_in_document_order` answers a per-*edge* question and I was reusing one answer for every element. Reproduced it before fixing, and it's worse than mis-naming — here is `x[b]`'s emitted score for `x[a] = m·pos` / `x[b] = m·neg`:

```
... OR (("m·pos" - PREVIOUS("m·pos")) = 0) then 0
else SAFEDIV(((PREVIOUS(m·neg)) - PREVIOUS(x[region·b])), ...) * SIGN(("m·pos" - PREVIOUS("m·pos")))
```

The port that slot actually reads is **frozen** in the numerator, while the guard and the sign come from a port the element never reads. That's a plausible-looking wrong number, not a missing one — precisely the class this branch exists to remove, one level down. Now resolved from that slot's own `OccurrenceRef::ModuleOutput`, falling back to the edge-wide answer for a non-module source and for an `ApplyToAll` target (neither records a per-slot module occurrence). Mutation-tested: forcing the fallback reds the new test.

**Arrayed link scores scanned only at the base slot** — real and it moved the numbers. `Results::offsets` gives the base, but an A2A score occupies one slot per element and `parse_link_offsets` expands to `base + idx`. Now scanning the full dimension-derived width: link scores measured non-zero went **1,149 → 1,308**, so 159 scores were being called dead because their base element happened to be zero.

**Infinite scores** — real. `load_step_scores` maps only NaN to zero and then keeps anything with non-zero magnitude, so `±inf` is live to discovery while my `is_finite()` check rejected it. Now using the same NaN-to-zero/non-zero predicate verbatim.

**Model-name canonicalization** — real. `sync.models` is keyed canonically while I indexed with the raw spelling, so an imported `MAIN` or `Main Model` would panic before producing any report, and the case-sensitive `== "main"` search missed it too. Canonicalized for both, with a named panic instead of a bare index if it still misses. The canonical spelling is also what `Diagnostic::model` carries, so the per-model filters now compare like with like.

On-cycle figures are unchanged by the harness corrections (117 non-zero / 74 compiled-but-zero); the feed-forward live count moved 631 → 638.

5,496 lib + 647 integration green; full pre-commit green; `#[ignore]`d tier unchanged at its 7 pre-existing failures (#915).

@codex review

bpowers:
Both real. Fixed in b99382a.

**Per-slot liveness** — you're right, and the evidence you point at is the right evidence: the full-width scan I added last round *observed* every slot and then threw away which one was live. So an A2A score with one live feed-forward slot marked its whole variable-level edge live, including a cyclic slot that was zero. That **deflates** `compiled_but_zero_on_cycle` — the population I was offering as the hidden-stub detector — so the error ran against my own claim rather than for it. My earlier "lower bound" argument covered the defect axis and did not cover this.

Rather than re-derive the expansion in the harness, I exposed the production one: `ltm_finding::link_score_offsets` is `parse_link_offsets` made public, and the join now keys on exactly the element edges discovery searches. A second implementation of the from-side projection is precisely the drift this repo warns about, and a measuring instrument that disagrees with the thing it measures is worse than no instrument. `analysis::build_link_expansion_context` was already public for the same reason, so the inputs match too.

The two axes now join at deliberately different levels, stated on the module: **defect** per variable (a compile failure is a property of a fragment), **runtime** per element edge (that's what discovery scores at). The granularity caveat is gone rather than narrowed.

What it shows on C-LEARN, measured with the same instrument run at the base commit — **cyclic element edges**:

| | main | branch |
|---|---|---|
| non-zero at some step | 572 | 523 |
| zero at every step (compiles) | 621 | 439 |
| no score offset at all | 1,281 | 539 |
| **total cyclic element edges** | **2,474** | **1,501** |

The absolute non-zero count falls slightly because the denominator falls further — 973 of those cyclic element edges were phantoms this branch removes. Coverage of cyclic element edges goes 23% → 35%, and unscored cyclic element edges 1,902 → 978. The variable-level defect axis is unchanged (35 → 0 dead-on-cycle, 15 no-score).

**Stale #716 claims** — real, and there were three, not one: the engine guide, `ltm_finding`'s recompute rustdoc, and `docs/design/ltm--loops-that-matter.md`. The design note also named `analyze_model_arrayed_module_loop_blocked_by_scalar_output_gap`, a test this branch renamed, so that reference was already broken. All three now describe the subscript stripping as live parity defense and point at `analyze_model_arrayed_module_loop_is_discovered_per_element`.

5,496 lib + 647 integration green; full pre-commit green; `#[ignore]`d tier unchanged at its 7 pre-existing failures (#915).

@codex review

Issue #1002, opened 2026-08-01
engine: sub-model implicit-var identity is nondeterministic under module-input parse contexts (HashMap-order index misalignment)


bpowers:
## Summary

A sub-model containing a stdlib module call (e.g. `sm = SMTH1(port, 3)` over a bound input `port`) compiles **nondeterministically**: across 8 identical runs of the same fixture, `compile_project_incremental` returned Ok on 2 and the unattributed

```
failed to compile fragments for variables: $⁚sm⁚0⁚arg1, $⁚sm⁚0⁚smth1, ...
```

error on the other 6. Whether the project compiles at all depends on the process's HashMap seed.

## Root cause

A HashMap-order-dependent index misalignment between the empty-context implicit-var registry and the with-module-inputs parse of the same parent:

- `model_implicit_var_info` keys implicit vars by (parent, `index_in_parent`), computed under one parse context.
- When the parent is instantiated with bound inputs, the module-input names widen `model_module_ident_context`, which changes the parent's parse and hence the implicit-var set/order.
- `ImplicitVarMeta.index_in_parent` computed under one context is then resolved under the other, so the identity of a sub-model implicit var is not stable across contexts, and which helper a given index resolves to depends on hash iteration order.

## Observable consequences (both verified with fixtures)

1. An ordinary sub-model-with-bound-inputs whose implicit helper fails compiles or fails depending on hash seed (the 2/8-vs-6/8 fixture above).
2. A phantom module-input dst binding (parent binds `sub.phantom` where `phantom` is undeclared — itself warned as `BadModuleInputDst`) makes `collect_all_diagnostics` report implicit-helper Error rows on runs where the project nevertheless compiles.

## Why it matters

Same nondeterminism class as #595 / #633 / #859 / #999 (unordered collection order reaching an observable), but here the observable decides **compile success**, not just which diagnostic is worded how or which variable a warning names. Any test, CI gate, or user workflow touching sub-models with stdlib calls over bound inputs is seed-flaky.

## Components

- `src/simlin-engine/src/db/query.rs` — `model_implicit_var_info`, `model_module_ident_context`
- `src/simlin-engine/src/db/fragment_compile.rs` — `lower_implicit_var` / `index_in_parent` resolution
- `src/simlin-engine/src/db/assemble.rs` — `build_module_inputs`

## Reproduce

[...]

Issue #913, opened 2026-07-09
engine: three more print_eqn/parser asymmetries (unary `!`, `!=`, and right-nested `^` re-parsing to a different AST)


bpowers:
## Problem

The invariant at stake: **`ast::print_eqn`'s output must be text `Expr0::new` accepts.** The engine re-parses `print_eqn` output constantly -- LTM partial-equation synthesis (`ltm_augment.rs`, ~15 call sites), `patch.rs:983`'s equation normalization, `conveyor_compile.rs:1951`, `builtins_visitor.rs` implicit-module argument text, and the MDL writer's `equation_to_mdl`. When `print_eqn` emits text the parser rejects, LTM emits a warning and drops the variable (its link score reads a constant 0), and the MDL writer takes its raw-text fallback (#912).

Fixing #912 / #863 closed the stacked-unary and unary-exponent-operand cases in the *parser*, and the `If`-under-an-operator case in the *printer*. Three more instances of the same class remain. All three empirically confirmed against `Expr0::new(s, LexerType::Equation)`.

### 1. `print_eqn` emits `!a` for a unary `not`, which the equation lexer cannot tokenize

`src/simlin-engine/src/ast/mod.rs` `PrintVisitor`'s `Op1` arm renders `UnaryOp::Not => "!"`, producing `!a`. But `src/simlin-engine/src/lexer/mod.rs` never lexes a bare `!` at all (only `<>` becomes `Neq`; there is no `!` character rule). `parse_eq("!a")` fails. There is even a comment in `src/simlin-engine/src/parser/tests.rs` saying "The lexer doesn't support '!' as a character - only the keyword `not` produces `Token::Not`". So `print_eqn(Op1(Not, x))` is unconditionally unparseable.

Suggested fix: print `not {inner}` (with a space). Note `test_print_eqn` in `ast/mod.rs` currently *pins* the broken `"!a"` output and would need updating.

### 2. `print_eqn` emits `a != b` for `BinaryOp::Neq`, which the lexer also rejects

Same file, `PrintVisitor`'s `Op2` arm: `BinaryOp::Neq => "!="`. `parse_eq("a != b")` fails; the lexer only produces `Neq` from `<>`.

Suggested fix: print `<>` (the form the lexer accepts, and the one the MDL writer already emits).

### 3. `print_eqn` renders a right-nested `^` without parens, but the XMILE parser is left-associative -- so the printed text re-parses to a DIFFERENT AST

[...]

Issue #372, opened 2026-03-01
engine: decouple module-aware parse context from per-variable parse cache


bpowers:
## Problem

`parse_source_variable` currently has a tension between correctness and incrementality:

- Making it globally module-aware fixes module-expansion-sensitive parsing cases (for example `PREVIOUS(x)` where `x` is a user-written stdlib-call auxiliary that should be treated as module-backed).
- But global module-aware parsing also broadens invalidation and broke per-variable cache stability guarantees in incremental compilation tests.

In this branch we reverted the global behavior and limited module-aware parsing to codegen-specific paths to restore cache stability while preserving correctness where needed.

## Why this matters

This is an architectural boundary issue in the incremental pipeline. Without a dedicated refactor, we risk repeated regressions between:

1. Correct module-aware expansion behavior, and
2. Stable, minimal per-variable invalidation/reparse behavior.

## Current workaround

- Keep `parse_source_variable` on the narrow per-variable path (no global module context).
- Thread module-aware parsing context only through specific codegen/implicit-info paths that require it.

This reduces risk now, but it is not a clean long-term design.

## Proposed direction

Introduce an explicit model-level input for module parsing context (for example a tracked `module_idents` set) and thread it into parsing/codegen in a way that preserves salsa cache granularity.

Design goal: correctness-sensitive module context should be explicit and local to the computations that require it, not an implicit global dependency that destabilizes unrelated variable caches.

## Acceptance criteria

- Add coverage for module-aware correctness cases (including `PREVIOUS` over module-backed identifiers) through incremental paths.
- Preserve existing per-variable cache stability tests (unchanged variable should not be reparsed due to unrelated edits).
- No regression in incremental compile behavior for implicit/module-expanded variables.
- Document the final parse-context dependency model in `docs/design/incremental-compilation.md` (or equivalent).

## Related

- `docs/tech-debt.md` item: "parse_source_variable Missing module_idents Context"
- Recent regression observed while making `parse_source_variable` globally module-aware and subsequently scoping behavior back to codegen-specific paths.

PR #1003, opened 2026-08-01
engine: LTM array-freeze materialization, diagnostics attribution/determinism, and group-aware ranking

### LINKED ISSUE #998: ltm: three ranking surfaces are dominated by groups of size one (dominant_periods, link ranking, lone pins)
## Summary

One root cause in three places. LTM's relative measures are **shares within a group**, and every surface that ranks them **globally** is systematically dominated by groups of size one — where the share is `±1` by construction and carries no information.

Found while driving LTM through pysimlin on C-LEARN. #994 documented all three on the Python surface as hazards; none is fixed.

### 1. `Analysis.dominant_periods` / `Run.dominant_periods` — partition-blind

`layout/metadata.rs::calculate_dominant_periods` takes a flat `&[FeedbackLoop]` with no partition field and ranks by `abs(importance)` across the whole model. But that series is a loop's share of **its own cycle partition**, so cross-partition values are not comparable. A loop alone in its partition reads exactly `1.0` at every active step.

On C-LEARN this makes the surface worthless: it reports the isolated trace-gas decay loops (`n2o -> n2o_uptake`, `hfc[hfc125] -> hfc_uptake[hfc125]`) as the model's dominant loops for the entire 1850–2100 run, and **never names a single climate loop**. Those same twelve loops are ranked 141st–153rd of 153 by `Analysis.loops` in the *same object*, whose "competitive-first" ordering deliberately demotes them. The two surfaces contradict each other.

Verified: every singleton-partition loop has `|score|` identically `1.0` while active.

Not a tuning problem — a flat cross-partition ranking is not well-defined. The surface must either report per partition or say which partition it means. The call site (`analysis.rs:427`) already has both `found_loops` and `partitions` in scope; `DominantPeriod` would need a partition field, which reaches the FFI struct, pysimlin, TS and the MCP `LoopDominanceSummary`.

### 2. `Link.average_relative_score()` — per-target normalisation ranked globally

The relative link score is normalised per target into `[-1, 1]`, which fixed raw scores' incomparability (#652) and inherited a new degeneracy: a link into a target with exactly **one** scored input reads `±1` regardless of whether the target matters. On C-LEARN, 143 of 949 scored links have a single-input target, and **58 of the global top 100** by this metric are such links. Restricting to multi-input targets helps but does not fix it — a two-input target whose second input is inert still reads 1.0.

A genuinely global ranking has to weight each target by that target's own importance, which the relative score deliberately divides out.

### 3. Lone pinned loops

A pinned loop occupies its own single-slot partition, so its relative score degenerates to `±1` (already documented). The documented workaround — read the raw `loop_score` series via `get_series` — is **element 0 only** on an arrayed loop, while the bare `get_relative_loop_score` and `Loop.behavior_time_series` report an argmax-abs aggregate across slots. Verified on `test/arrayed_population_ltm/arrayed_population.stmx`: for one loop the aggregate tracks Boston while the raw read tracks NYC, with nothing signalling the mismatch. Per-element **raw** loop scores are unreachable today.

## Suggested direction

Make within-group normalisation structural rather than advisory: ranking APIs should either take the group as a parameter or return grouped results, so a cross-group comparison is not expressible. That closes all three rather than patching each.

## Related

#670 is a different `dominant_periods` defect (flicker on data-driven models from instantaneous shares inheriting input noise) and is not this. #652 introduced the relative link score. #672 covers arrayed pins' aggregate confusing share-based analysis.

### LINKED ISSUE #999: engine: collect_all_diagnostics is not reproducible run-to-run (unit pass names a different variable each run)
## Summary

Two runs of the **same binary** over the `test/` corpus produce diagnostic sets differing by ~184 lines. All the differences are unit warnings, differing both in which variable is named and in row count.

Example: `scirev8.mdl` names `ada_average_difficulty_of_anomalies` on one run and `ardp_anomalies_relative_to_dominant_paradigm` on the next.

Found incidentally while measuring the blast radius of #994's codegen-diagnostic change — the sweep had to be run twice to notice that the baseline itself moved. Not caused by that change.

## Why it matters

Not currently user-visible in a damaging way: libsimlin's `collect_models_with_unit_warnings` already works around it by keying on model name rather than on the reported variable. But:

- any future "diagnostics are stable across an edit" gate on the unit pass will flap;
- it made an unrelated measurement harder to trust, which is how it was found;
- non-determinism in a diagnostic set is the same class as #595 (init runlist ordering) and #633 (layout per-seed), both of which turned out to be `HashMap`/`HashSet` iteration order reaching an observable.

The likely cause is the same: an unordered collection somewhere in unit inference or its conflict reporting, whose iteration order decides which member of a conflicting set gets named. `units_infer`'s `find_constraint_mismatches` and the conflict-clarification path are the places to look first.

## Reproduce

Run whole-corpus diagnostic collection twice in the same binary (or two invocations) and diff the output. Every `test/` model that produces a unit warning with more than one candidate variable in its conflict set is a potential differ.

Related: #595, #633, #613.

### LINKED ISSUE #1000: engine: compile_implicit_var_phase_bytecodes still swallows its codegen error (unattributed helper failures)
## Summary

#994 made a codegen rejection during per-variable fragment compilation attributable — it now emits a diagnostic naming the variable and the construct codegen refused, where previously it produced nothing at all.

`compile_implicit_var_phase_bytecodes` (`src/simlin-engine/src/db/fragment_compile.rs:1022`) was **not** included and still discards its codegen error the same way. So a failure in an implicit helper still surfaces only as the unattributed assembly message:

```
failed to compile fragments for variables: $⁚s⁚0⁚arg0
```

with nothing saying which construct was refused. That is the #913 shape.

This was disclosed as scope in #994's own comment rather than left implied, and is filed so it is not lost. The fix is the same shape as the one already landed: route through `compile_phase_to_per_var_bytecodes_reporting` and accumulate the reason.

Two things to be careful of, both learned from the landed change:

1. **Severity.** The landed one uses `Error` because the model does not simulate; verify that holds for the implicit-helper path too rather than inheriting it. The blast radius there was *measured* (a sweep across the `test/` corpus confirmed added rows land only on projects that already fail to compile, and no compiling project gains an error) — do the same here rather than assuming symmetry.
2. **The variable name must be interpolated into the message**, not left on the `variable` field alone. `errors.rs`'s `Assembly` arm never renders that field and the CLI prints only the message, so a field-only version names the variable nowhere a user looks.

Related: #913.

bpowers:
Six items from the post-#1001 backlog, executed in order, each followed by a fresh adversarial-review round (the review-fix commits are interleaved on the branch).

Fixes #1000
Fixes #999
Fixes #998

## 1-3: the #995 LTM burndown (three commits + reviews)

`PREVIOUS(<array slice>)` has no codegen path, so every ceteris-paribus partial that needed to freeze an array slice -- or hold a whole arrayed dep frozen inside a vector builtin's view position -- either failed fragment compile (silently, before #994) or was declined. Three changes close the LTM half:

- **Array-slice freezes materialize as arrayed helper vars** (`ltm_augment_array_freeze.rs`): each frozen slice becomes a `$⁚ltm⁚freeze⁚…` helper (`Equation::Arrayed`, per-row `PREVIOUS(arr[axis·elem])` arms compiling to `LoadPrev`), referenced with wildcard subscripts so A2A scores read the whole frozen row rather than a silently-broadcast element. Helpers dedup by content, ride the salsa `ShapedLinkScore` value, and sort into their own runlist tier. Dimension-name pins now happen at AST level *before* materialization, which is what let the C-LEARN population of these fragments actually compile.
- **Rank-like partials decline loudly** (`RankLikePartial`): an array-producing builtin (VECTOR SORT ORDER / RANK / ALLOCATE* / ELM MAP) cannot live in a per-element scalar partial, so those partials are declined with a warning instead of emitting uncompilable fragments -- checked on the substituted AST so hoisted aggregates are not wrongly declined.
- **View-position freezes** (the bare-`*` half): codegen positions that must stay views over storage (select/elm_map/allocate args 0-1, sort_order/rank arg 0) get whole-dep freeze helpers referenced with original indices, preserving ELM MAP full-storage-base semantics.

C-LEARN trajectory: **640 -> 0** failing LTM fragments; loop-carrying no-score edges **15 -> 1**; score coverage on loop-carrying edges 83% -> 99.5%; zero all-dead-score edges. The one residual loop-carrying edge (`co2eq_emissions_from_hfc -> total_co2eq_emissions_from_hfc`) is an un-hoisted per-element reducer -- a #792-class root cause, loudly declined, deliberately out of scope here.

**#995 is deliberately NOT closed by this PR**: it also covers the ordinary-model (non-LTM) half -- materializing `PREVIOUS(<slice>)` written by a modeler (options A/D, a Pass-1 materialization) -- which remains open.

## 4: #1000 -- implicit-helper compile failures are attributed

`compile_implicit_var_phase_bytecodes` now reports *why* a SMOOTH/DELAY/TREND capture helper failed to compile, and `compile_implicit_var_fragment` accumulates an Error diagnostic naming the helper, its parent ("synthesized while parsing 'out'"), and the refused construct -- closing the #913-shaped unattributed-batch-failure gap for implicit variables the way #994 closed it for explicit ones. The review round surfaced a pre-existing nondeterminism in sub-model implicit-var identity, filed as #1002 (out of scope; cross-referenced #372).

## 5: #999 -- diagnostics are a function of the model

Every HashMap-iteration order that reached the diagnostics collection is sorted: unit-inference constraint generation, both units_check element loops plus the declared-units per-element loop, the db/units SMTH-argument and conveyor-parameter loops, the per-variable and implicit-helper probe loops, and the models map. 931-row corpus check: byte-identical across repeated runs (was ~184 lines of churn). Every sort is pinned by a determinism fixture that was mutation-tested (reverting any one sort fails its fixture 3 of 3 runs); the review round caught a vacuous fixture (`"3 {widgets}"` is an equation *comment*) and a sort the commit message claimed but did not contain, both fixed.

## 6: #998 -- ranking within the normalization group

One root cause in three places: LTM's relative measures are shares WITHIN a group (a cycle partition for loops, a target's scored inputs for links), and every surface ranking them globally was dominated by groups of size one (share ±1 by construction).

- **Dominant periods are computed per cycle partition**, each period tagged with the partition it describes (`DominantPeriod.partition`, threaded through the FFI/pysimlin/MCP/TS surfaces and pysimlin's `Run` greedy). The isolated single-loop partitions no longer smother the competitive partition's timeline. The review round found the pattern surviving inside the partition-`None` subset -- unpartitioned (module-internal) loops were pooled -- now each gets its own solo group, mirroring discovery's `NormGroup::Solo`.
- **Links expose their normalization-group size** (`SimlinLink.scored_input_count`, additive ABI 28->32; pysimlin `Link.scored_input_count` + the grouped `links_by_target()` helper; TS `scoredInputCount`), making the ±1-by-construction singleton case detectable.
- **Per-element raw loop scores are reachable** (`simlin_analyze_get_loop_score` / pysimlin `Sim.get_loop_score`), the informative series for a lone pin -- the old `get_series` workaround silently read element 0 only on an arrayed loop.


bpowers:
@codex review

bpowers:
@codex review

PR #1004, opened 2026-08-01
pysimlin: rewrite README around loop-dominance analysis

bpowers:
The pysimlin README is the PyPI front page and the package's only prose documentation, but it steered the exact user it targets -- someone loading a real model to analyze feedback-loop dominance -- wrong in two ways: it never mentioned `Model.analyze()`, the only loop surface that works on models past the exhaustive-enumeration gate, and its link-ranking example used `average_score()`, which the docstrings themselves warn is unusable for ranking (see #652).

This rewrite keeps the existing comprehensive single-page structure and section order, and reworks the content around the analysis workflow:

- **Quick Start** builds a logistic model and ends with the loop payoff: `run.loops` and `run.dominant_periods` showing dominance handing off from the reinforcing to the balancing loop at the S-curve inflection (t=67), including an explanation of why loop id `u1` reports runtime polarity `R`.
- **New "Loop Discovery for Large Models" section** demonstrates `Model.analyze()` on World3 (`test/metasd/WRLD3-03/wrld3-03.mdl`) with output from a real run: 200 loops in a single 15-stock cycle partition, discovered in under a second. Blocks are skip-marked with a prose pointer to the model file.
- **Link ranking corrected** to `average_relative_score()` with a short explanation of the per-target normalization.
- The two ~100-line inline examples were verbatim copies of `examples/*.py` that CI already runs standalone; they are now links. The API reference sections continue with the quick-start model instead of rebuilding it three times.
- **New "Using pysimlin from AI Agents" section**: `explain()`, `check()`, transactional edits (with an `expect-error` block proving a bad equation raises and leaves the model unchanged), and a pointer to `@simlin/mcp`.

Writing the step-by-step simulation example surfaced a real API footgun, fixed in the first commit: `Sim.get_run()` returned a lazily-populated `Run` that failed with "Sim has been closed" once the `with model.simulate() as sim:` block exited -- the natural usage the docstrings themselves show. `get_run()` now materializes every surface before returning (a snapshot taken at call time); `Model.run()` deliberately keeps its lazy path since the Sim it creates is never closed. Covered by two new red-first lifetime tests.

Every executable README block runs under `tests/test_readme_examples.py`; skip-marked blocks (World3, VDF, file loading, matplotlib) were hand-run against real files, and all quantitative claims in the prose come from those runs.

### DIFF
--- a/docs/dev/python.md
+++ b/docs/dev/python.md
@@ -8,7 +8,7 @@
 
 ## Thread Safety
 
-- **All wrapper classes** (`Project`, `Model`, `Sim`) have a per-instance `threading.Lock` (`self._lock`) that protects `_ptr` and cached state.
+- **All wrapper classes** (`Project`, `Model`, `Sim`) have a per-instance lock (`self._lock`) that protects `_ptr` and cached state. `Project` and `Model` use `threading.Lock`; `Sim` uses `threading.RLock` so `Sim.get_run()` can hold the lock across the whole Run snapshot (which re-enters the individually locked accessors on the same thread) -- without that outer hold, a concurrent `run_to`/`reset`/`close` could land between the snapshot's reads and tear it.
 - **Module-level `_finalizer_refs`** (a `WeakValueDictionary`) is protected by `_refs_lock` in `_ffi.py`.
 - When adding new methods to wrapper classes, always acquire `self._lock` before touching `_ptr` or mutable state.
 - **Lock ordering**: `Model` methods must release `self._lock` before calling `Project` methods (which acquire the project's lock) to prevent deadlocks. Use double-checked locking for caches: check cache with lock, compute without lock, write cache with lock.
--- a/src/pysimlin/CLAUDE.md
+++ b/src/pysimlin/CLAUDE.md
@@ -21,7 +21,7 @@ For Python-specific standards (thread safety, lock ordering), see [docs/dev/pyth
 
 ## Thread Safety
 
-All wrapper classes have per-instance `threading.Lock`. Lock ordering: release `Model._lock` before calling `Project` methods. See [docs/dev/python.md](/docs/dev/python.md).
+All wrapper classes have a per-instance lock (`Sim`'s is an `RLock` so `get_run()` can hold it across the whole Run snapshot). Lock ordering: release `Model._lock` before calling `Project` methods. See [docs/dev/python.md](/docs/dev/python.md).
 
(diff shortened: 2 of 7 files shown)


bpowers:
Went through the five Codex findings one by one; dispositions in f62c9e2e:

**1. Lock the entire snapshot operation — real, fixed.** `_materialize()` issued many individually-locked reads, so a concurrent `run_to`/`reset`/`close` could land between them. The worst failure mode is silent: a reset mid-materialize makes the remaining per-variable reads raise, those errors are suppressed by design in `_build_results_dataframe`, and the snapshot comes back as a frame with missing columns. `Sim._lock` is now an `RLock` held across the whole snapshot in `get_run()`. A deterministic fault-injection test pins the exact interleaving (a timing-based hammer test never tripped under the GIL in 500 iterations, so it alone would have been a vacuous guard; both are included). `docs/dev/python.md` records why Sim deviates from the plain-Lock pattern.

**2. Copy overrides into the snapshot — real, fixed.** Both `Sim` and `Run` aliased the caller-owned dict; both now copy at construction, with a red-first test covering the `simulate()` and `run()` paths.

**3. Limit importance comparisons to one partition — real, fixed.** The bullet overclaimed ahead of the qualification three paragraphs later; it now scopes comparability to a single cycle partition inline.

**4. Document the compilation exception from invalid edits — mechanism claim is wrong, but the doc hedge is right; adjusted.** Verified empirically: every invalid-edit shape tested (unknown dependency, syntax error, empty equation, self-reference, circular dependency, bad function call) raises `SimlinRuntimeError`, not `SimlinCompilationError` — `simlin_project_apply_patch` reports the first *specific* diagnostic code (e.g. `UNKNOWN_DEPENDENCY`=29), never the aggregate `VARIABLES_HAVE_ERRORS`=30 that `check_out_error`'s compilation branch requires. That said, the class selection is an accident of diagnostic ordering rather than a designed contract (`_ffi.py` itself documents both classes for apply-patch), so the README now names the stable `SimlinError` base. Filed #1005 to make the class deterministic by design.

**5. Keep the linked examples exercised in CI — incorrect, no change.** `tests/test_examples.py` runs every `examples/*.py` as a subprocess (parametrized over a glob, so it covers all five scripts, not just the two in `pysimlin-tests.sh`), and the CI pysimlin job runs `pytest tests/`. The README's "Both run in CI on every commit" is accurate — the examples' CI execution never depended on the inline README copies.

Issue #595, opened 2026-05-20
engine: init runlist ordering is nondeterministic for an init-time algebraic cycle through a synthetic-module stock output


bpowers:
## Summary

The init-phase dependency runlist can be **nondeterministic** (HashMap-iteration-order dependent) when a model has a genuine init-time algebraic cycle that passes through a **synthetic-module stock output**. This is the init-phase analogue of the dt-phase ordering bug fixed in commit `eeace429` ("engine: chain-break stock submodel-output deps in the dt runlist"), which only repaired the dt phase and explicitly left the init phase carrying the edge.

This is a latent **correctness** bug: the model compiles (cycle detection reports `has_cycle == false`), but the chosen init order is HashMap-iteration-order dependent and can place a module before its own input, producing nondeterministic / incorrect initial values.

## Mechanism

A synthetic module is treated as a **sink** by the init-phase cycle relation -- `init_walk_successors` (`src/simlin-engine/src/db_dep_graph.rs`, the `if info.is_module { return Vec::new(); }` early-return) returns no successors for a module. But the *ordering* data follows the real edge `module -> input` (`module·output` references are collapsed to the bare module name by `normalize_deps`, and the module carries its input `src` as a direct dep).

In the **dt phase** this same shape was resolved by chain-breaking the stock-submodel-output dependency: a stock's dt value does not depend on its inputs within the same step, so the spurious `reader -> module` edge can be dropped. See commit `eeace429` and the legacy gate `model.rs::module_output_deps`.

In the **INIT phase** that chain-break is **NOT valid**: a stock's INIT value *is* a real init dependency. `init_walk_successors`'s own rustdoc (db_dep_graph.rs ~143-153) documents this asymmetry deliberately -- "a Stock is NOT an init-phase sink", "no stock filter and no stock sink (a stock-targeted init dep is a real init dependency, kept)". So the edge cannot simply be removed the way it was in the dt phase.

The result:
- Cycle **detection** never sees a cycle, because `init_walk_successors` makes the module a sink -> `has_cycle == false` -> the model compiles.
- But the init **ordering** data (the `initial_deps` field consumed by the `topo_sort_str` closure, db_dep_graph.rs ~2149, called for the initials runlist at ~2311) still has the `module -> input` edge plus, in the cyclic shape, a `reader -> module` edge that closes an ordering cycle invisible to detection.
[...]

PR #1006, opened 2026-08-01
engine: make implicit-helper identity a name, not a hash-ordered position

### LINKED ISSUE #1002: engine: sub-model implicit-var identity is nondeterministic under module-input parse contexts (HashMap-order index misalignment)

bpowers:
Fixes #1002

A sub-model that instantiates a stdlib module over a bound input compiled or
failed depending on the process's HashMap seed. On a 30-line fixture, roughly
one run in five returned the unattributed `failed to compile fragments for
variables: $⁚sm⁚0⁚arg1, $⁚sm⁚0⁚smth1`.

## Root cause

Not the one the issue proposed. `BuiltinVisitor` accumulated its synthesized
SMOOTH/DELAY/TREND/PREVIOUS helpers in a `HashMap` and every producer of
`ParsedVariableResult::implicit_vars` emitted it with `.values()`. Rust draws a
fresh `RandomState` key per map, so two parses of *one* variable -- in the same
process, differing only in their `ModuleIdentContext` -- reported the helpers in
two different ORDERS. `ImplicitVarMeta` identified a helper by its POSITION in
that vector, and a position is meaningful only against the exact vector it was
taken from: `model_implicit_var_info` records it under the no-extra-idents
context while assembly resolves it under the context widened by the instance's
module-input names. When the two orders disagreed the position named an aux
where a module was expected, or the reverse; both fail to lower, so the
sub-model lost every helper fragment.

## The fix

Two changes, in three production edits, kept together because they close
different properties.

**Identity.** `ImplicitVarMeta::name` replaces `index_in_parent`. A helper
resolves to itself or to nothing, never to another helper. `find_in` carries an
`index_hint` checked against the name before it is accepted, so the common case
stays O(1) (see Performance).

**Order.** `BuiltinVisitor::vars` is an `IndexMap`, and a new
`elements_in_stable_order` walks `Ast::Arrayed`'s per-element slot map -- itself
a `HashMap` -- at both expansion sites. Without this the two salsa-cached
vectors carrying that order (`ParsedVariableResult::implicit_vars`,
`VariableDeps::implicit_vars`) still differ run to run, salsa stops backdating,
and the compiled artifact stops being reproducible -- the #595 class, invisible
in numeric results and in the compile outcome alike.

Also deletes `LtmImplicitVarMeta::index_in_parent`, which nothing read.

## What gates what

Measured, not predicted: each edit was reverted alone, faithfully, and the suite
re-run three times. The matrix is in the test module header.

| reverted | reds |
|---|---|
| `IndexMap` | all three order tests |
| `elements_in_stable_order`, per-element site | `per_element_…` alone |
| `elements_in_stable_order`, shared-visitor site | `undimensioned_arrayed_…` alone |
| identity | `an_implicit_helper_declines_…` alone |
| the hint's name check | `an_implicit_helper_declines_…` |

The identity fix having one gating test is a property of the problem, not thin
coverage: with a stable order a positional index resolves correctly whenever
the two contexts synthesize the same helper set, so only a diverging-set fixture
can tell the two schemes apart.

## Performance

`find_in` replaces an index lookup with a name match, which is O(k^2) per parent
variable, and k is not small on ordinary models -- an apply-to-all `SMTH1` over
an N-element dimension mints ~2N helpers on ONE variable. Measured with two
release binaries (worktree at the parent commit vs. the branch), interleaved, on
`x[Dim] = SMTH1(y[Dim], 3)`:

| N | main | scan only | with hint |
|---|---|---|---|
| 200 | 0.08s | 0.14s | 0.08s |
| 400 | 0.29s | 0.53s | 0.31s |
| 600 | 0.64s | 1.20s | 0.65s |
| 800 | 1.12s | 2.13s | 1.16s |

C-LEARN is unmoved throughout (0.39s), which is why a C-LEARN-only check missed
the regression and why the synthetic is in the record.

## Verification

Full pre-commit gate green on each commit. 5553 lib + 647 integration. The
`#[ignore]`d tier has exactly its seven pre-existing failures, name for name
against #915. The original reproduction is 0/12.

Three adversarial review rounds; the first found a live coverage gap this
originally shipped with (the shared-visitor call site had no test, and a
name-only probe was structurally blind to it because that arm keeps names
stable and moves only the binding), a tautological anti-vacuity guard, and two
wrong cells in the mutation matrix. All fixed, each with the demonstration in
the commit message.

## Not closed by this PR

- **#1002's second reported observable** (a phantom module-input binding
  randomising implicit-helper diagnostics) does not reproduce, and the reason is
  structural: `collect_all_diagnostics` probes helpers with the empty input set
  AND derives them from `model_implicit_var_info`, which uses the same context,
  so no second context exists on that path. Kept as a guard test with that
  reasoning recorded, so it starts working if the probe is ever changed to use
  real instantiation input sets.
- **The cross-context helper-SET divergence.** When the widened context
  synthesizes genuinely different helpers (`PREVIOUS` over a bound input whose
  port lacks `access="input"`), a model's layout and its runlists disagree about
  which helpers exist and the project fails to compile. Deterministic,
  pre-existing (verified byte-identical at the parent commit), and #372's
  territory. This PR only makes the mis-resolution loud instead of silent.
- **Repeated helper names with different bodies.** An `Arrayed` element with its
  own init equation, or a variable with `compat.active_initial`, makes the
  initial parse read a different equation while the helper-naming counter
  restarts at zero -- so two bodies can claim one name, which
  `model_implicit_var_info` collapses and `compute_layout` gives one slot. Found
  while writing the tripwire, pre-existing, and disclosed on
  `repeated_implicit_helper_names_carry_identical_helpers` rather than fixed:
  the fix is a phase discriminator in the synthesized name.


bpowers:
All three findings were real. Verified each against a fixture before changing anything; two of them falsified claims this PR was making.

**P1 — phase-collided helper names (confirmed, and it was silent).** An aux with equation `SMTH1(driver * 2, 3)` and ACTIVE INITIAL `SMTH1(driver * 100, 7)` **compiled**, keeping only the initial pass's `$⁚v⁚0⁚arg0`, so the dt phase smoothed the wrong expression. Both passes name from a counter that restarts at zero, `model_implicit_var_info` is name-keyed, and `compute_layout` gives one slot per name.

Fixed by merging helpers across the two phases instead of appending, with `dedup_vars_by_ident`'s rule: a byte-identical repeat collapses (the `Arrayed` arm re-parses every slot, so that is the common case — it also halves the redundant entries), a genuine collision is a loud `DuplicateVariable`.

Refusal rather than repair, deliberately. Phase-qualifying every synthesized helper name changes every helper's identity and its layout slot; that is its own change. Loud rejection is the choice `dedup_vars_by_ident` already makes for the within-one-pass twin of this collision. Blast radius measured rather than assumed: full engine suite green including `simulates_clearn`, heavy `#[ignore]`d tier unchanged at its 7 pre-existing failures — and ACTIVE INITIAL appears in both C-LEARN and WRLD3, so the refused shape genuinely does not occur in the corpus.

One note on the fixture, since it nearly fooled me: my first version used a *different* variable in each equation, and that failed to compile even before the fix — for an unrelated reason (the discarded helper's dependency was missing from the survivor's dep set). It would have passed against the defect it was meant to catch. The committed fixture has both equations reference the same variable.

**P2 — names are not context-stable (confirmed; the PR's claim was wrong).** Reproduced exactly as described: for `PREVIOUS(port, 0) + SMTH1(port + 1, 3)` over a bound `port`, `$⁚sm⁚0⁚arg0` is the SMTH argument in the no-extra-idents parse and the PREVIOUS capture in the widened one. So "resolves to itself or to nothing, never to another helper" is false, and it is corrected everywhere it was stated.

I did not adopt the suggested structural identity, and here is the reasoning rather than a dismissal. The collision *requires* the two parses to synthesize different helper sequences — which is the same condition that makes the model's layout disagree with its runlists, so the project fails to compile before any mis-resolved fragment can run. I verified that: the fixture returns `Err(failed to compile fragments…)`. Both halves are now pinned in one test, so if that compile ever starts succeeding, the mis-resolution stops being harmless and something reds. Making the names themselves context-stable is #372's explicit model-level parse context, not a local change here.

**P3 (from the other reviewer) — `enumerate_module_instances`' rustdoc left on the new test-only accessor.** Fixed; the doc is back on the function it describes.

Also worth recording against an earlier round: a reviewer reported a perf regression on `find_in`, then retracted it after an interleaved A/B. The retraction was wrong. Two release binaries in worktrees, interleaved, `x[Dim] = SMTH1(y[Dim], 3)`: at N=800 the scan-only version took 2.13s against main's 1.12s. That is why `find_in` carries a name-checked `index_hint` (1.16s), and why dropping the hint's name check reds the identity test.

@codex review

bpowers:
Follow-up review round landed as `60d14d25`. Two of its findings were already closed by `7ba78dc9` (pushed after the reviewed commit) but had left the docs asserting the opposite; one was a genuine blind spot in a test probe.

**The `index_hint` doc claimed something false and then explained why it was false.** It said a wrong hint "cannot change the answer", then two paragraphs later noted that a name is *not* unique within a parse — which is exactly the missing precondition. `7ba78dc9`'s cross-phase merge makes names unique, so the claim is now true; the doc now states the precondition, what establishes it, and the test that checks it, and separates that within-parse property from the cross-parse residual, which is a different thing and stays. It also cited a test by a name I'd since renamed.

**The `dedup_vars_by_ident` hole was the same defect**, and is closed by the same merge — its guarantee ran inside one `instantiate_implicit_modules` call while the dt/init append happened above it, so a colliding pair was never seen by it. The reviewer's fixture is now a second arm of the regression test, with the doc recording that only the same-dependency arm gates the fix (theirs fails without it too, for an unrelated dep-set reason) so nobody later deletes the load-bearing one.

**Adopted: `implicit_helper_signatures` had a named blind spot.** `Variable::Module` has no equation, so `{:?}` degenerated to the bare name for exactly the helper kind carrying the wiring — a mutation re-binding which module reads which argument helper would have been invisible. Modules now render model name and references. The doc records that the compiled artifact is still strictly stronger (differs on 21 of 23 repeats under the same mutation), so the next person can reach for it deliberately rather than assuming this probe is the ceiling.

**On the performance disagreement, now settled — and the resolution is worth recording for anyone doing A/B work in this repo.** The earlier retraction came from sharing one `CARGO_TARGET_DIR` across git worktrees: cargo produced one artifact path and never rebuilt on switching, so every "interleaved main vs branch" measurement was one binary compared against itself. A clean run with three distinct binaries reproduces the regression (+53%/+60%/+63% at N=200/400/800, C-LEARN unmoved) and confirms the `index_hint` removes it.

Better still, it settles the algorithmic claim without a clock: instrumenting the scan, N=800 examines **17,289,600 entries without the hint and 14,400 with it** — a 1200x reduction — at a **100% hit rate** on C-LEARN, WRLD3, scirev7 and the synthetic. That is now the doc's primary evidence, with wall-clock kept as the secondary number it is.

Full pre-commit gate green on each commit; heavy `#[ignore]`d tier still at its 7 pre-existing failures, name for name.

@codex review

bpowers:
All three findings were real, and one of them was a regression the previous commit introduced. Fixed in `28a26218`.

**P2 — the cross-phase merge was O(k²), the same shape it had just removed elsewhere. Confirmed by measurement.** Scanning `implicit_vars` per incoming helper costs a canonicalize-and-intern per comparison, and `k` is ~2N for an apply-to-all `SMTH1` over an N-element dimension. Worse than the local quadratic: `model::ModelStage0` threads *one* vector across every variable of a model, so each variable was paying for every helper minted before it.

Measured against the previous commit, separate target dirs, interleaved:

| N | pre-merge | with merge | with index |
|---|---|---|---|
| 400 | 0.30s | 0.42s | 0.30s |
| 800 | 1.18s | 1.51s | 1.14s |

A name→index map, seeded **empty** and grown as helpers are pushed, removes it. Seeded empty deliberately: only helpers of the same *parent* can collide (a synthesized name embeds its parent's ident, and two parents sharing a canonical name is already a `DuplicateVariable` error), so seeding from the shared vector would have reproduced the model-wide quadratic the map exists to remove.

**P1 — the collision test covered one of three routes. This was the most valuable of the three findings.** `parse_equation` has exactly three arms that can make the initial pass read a different equation than the dt pass, and I had tested one. Enumerating them was not academic: with the merge reverted, **three of the four rows now present compile and silently drop a helper body** — `scalar_same_dep`, `apply_to_all`, and `arrayed_per_element_init`. A regression confined to the `ApplyToAll` or `Arrayed` route would have left the old test green, and two live silent-corruption routes were uncovered.

The doc now records which row gates what, because `scalar_disjoint_dep` fails without the merge too (unrelated dep-set reason) and must not be mistaken for one that does the work. The loop reports every failing row at once rather than stopping at the first, since the routes are independent.

**P1 — stale positional contracts.** `lower_implicit_var`'s rustdoc still described the chain as `parent → parsed.implicit_vars[index]` with `None` meaning an absent index, and `assemble.rs` carried the same prose twice more. All now describe resolution by name, so nobody reintroduces a positional assumption by following the documented path.

Full pre-commit gate green; heavy `#[ignore]`d tier still at its 7 pre-existing failures.

One note on methodology, since it bit an earlier round of this review badly enough to invert a conclusion: worktree A/B in this repo needs a **separate `CARGO_TARGET_DIR` per worktree**. A shared one makes cargo emit a single artifact path and skip the rebuild on switching, so the comparison silently measures one binary against itself. The numbers above were taken with distinct target dirs and md5-verified distinct binaries.

@codex review

bpowers:
Two verifications worth recording, both reproduced independently after a reviewer raised them. No code changes; `28a26218` stands.

**The branch carries no net regression against main.** All the performance numbers in this thread so far have been commit-to-commit (1.18 → 1.51 → 1.14), which shows each step is sound but structurally cannot see cumulative drift — four commits each landing "back to about where we were" can still sit above `main`. Measured directly, fresh worktree at `d5db88a3` with its own target dir, md5-verified distinct binaries, 3 interleaved rounds:

| model | main `d5db88a3` | head `28a26218` |
|---|---|---|
| N=200 | 0.08s | 0.08s |
| N=400 | 0.30s | 0.30s |
| N=800 | 1.12s | 1.12s |
| C-LEARN | 0.39s | 0.38s |

**The new refusal rejects nothing in the corpus — checked the strong way.** "The engine suite is green" was the weaker evidence for this, because the suite only *compiles* a subset of `test/`, and a newly-added compile rejection is exactly the change where an un-compiled corpus model could break with nothing noticing. So: every model file under `test/` — all 506 `.stmx`/`.xmile`/`.itmx`/`.mdl` — compiled with both binaries, outcomes diffed.

```
compile-outcome differences main -> 28a26218:   0 of 506
models emitting a duplicate_variable diagnostic: 0 on main, 0 on head
```

**On landing that census as a test: deliberately not doing it, and the reason is #915.** Nothing in pre-commit or CI runs the `#[ignore]`d tier — that is what #915 tracks — and it currently carries 7 unnoticed failures. A ratchet placed there would read as protection while never executing, which is precisely the failure mode this branch has spent four rounds correcting in its own tests. The honest version of that gate wants #915 fixed first, which is not this PR's scope. The evidence is recorded here instead.

One methodology note that inverted a conclusion earlier in this review and is worth the next person not rediscovering: **worktree A/B in this repo needs a separate `CARGO_TARGET_DIR` per worktree.** A shared one makes cargo emit a single artifact path and skip the rebuild on switching, so the comparison silently measures one binary against itself — and the tight agreement that produces looks exactly like evidence of no regression.

bpowers:
Both real, both fixed in `19acb6bc`. No behaviour change — every one of these is a string.

Correcting a false claim at the site where it was argued is not the same as correcting it, and that is the mistake here. When I corrected "resolves to itself or to nothing, never to another helper" on `ImplicitVarMeta::name`, three copies of it outlived the correction — the test module header, an assertion message, and a comment — plus a fourth statement in the order test's failure text still describing identity as *positional*, which it stopped being two commits earlier. Grepped for the phrasings this time rather than fixing the one I remembered writing.

The assertion message was the worst of the four, because it fires while someone is debugging: it told them a helper "must resolve to itself or to nothing" at precisely the moment that is the thing in question. It now states what the check actually verifies — a fragment carries the name it was resolved by — and points at the test that owns the rest.

The header now separates what name-based identity buys (no compile outcome depends on the vector's order) from what it does not, in its own paragraph, naming the test that builds the collision and the failing compile that bounds it, and saying outright not to build mutation work on the stronger invariant.

The order test's failure text described the pre-fix mechanism. It now leads with why the order still matters — two salsa-cached values with derived `PartialEq`, so an unstable one defeats backdating — and keeps the positional behaviour as history, since that is where the seed-dependence came from and a reader hitting this failure benefits from knowing it.

Verified no copies remain: `grep` for all four phrasings across `src/simlin-engine/` returns nothing.

bpowers:
@codex review

PR #1011, opened 2026-08-04
engine: fix unit-analysis false positives and diagnostic quality

### LINKED ISSUE #613: engine: residual ~14 C-LEARN unit diagnostics need a Vensim-parity strictness decision (IF-branch / log-argument / coefficient cancellation)
## Description

After four root-cause fixes on branch `fix-clearn-unit-errors` reduced the C-LEARN unit-diagnostic flood from 481 spurious diagnostics to ~14 (fixes: Context self-alias registration, `units_infer` time-unit alias resolution, `units_check` tolerating unknown dependency units, and skipping macro-marked models during unit checking), C-LEARN still produces ~14 unit warnings that genuine Vensim does not flag as mismatches. These break down as:

- **6-7** `computed units 'gtonsc/degreesc/yr' don't match specified units` on permafrost/clathrate methane-flux variables (`flux_c_from_permafrost_release`, `ch4_emissions_from_permafrost_and_clathrate`) -- a temperature-sensitivity-coefficient unit that does not fully cancel.
- **6** `expected left and right argument units to match, but 'dmnl' and '1/ppm' don't` on the `ph` variable (a log/concentration formula).
- **1** `IF branches have different units: then '1/gtonsco2', else '1/yr'` on `global_annual_rate_of_co2eq_2050_to_2100`.
- **1** umbrella main-model `unit inference [...]: unit_mismatch` warning (which underlying conflict it reports is nondeterministic -- see #474 for the order-dependence root cause).

## The open question

It is not yet established whether these are:

1. **Simlin being legitimately stricter than Vensim** -- e.g. Vensim may not check IF-branch unit consistency, or may not require the argument of a `LOG`/concentration formula to be dimensionless. If so, we may want to relax those specific checks for Vensim parity (an IF-branch-units check and a log-argument-dimensionlessness check are the two candidate relaxations).
2. **Genuine dimensional inconsistencies in the C-LEARN model** that Vensim happens to tolerate silently. If so, the diagnostics are correct and the resolution is to document them as known model-authoring imprecision (not an engine change).

Resolving this requires per-variable investigation of the C-LEARN equations against Vensim's actual unit-checking semantics for each of the three distinct message classes (coefficient non-cancellation, log-argument dimensionlessness, IF-branch consistency).

## Why it matters

- Unit checking is meant to be a verification safety net; residual noise on a flagship real-world model (C-LEARN) erodes trust in it and can mask genuine errors.
- If Simlin is over-strict relative to Vensim, every imported Vensim model with these idioms inherits spurious diagnostics.
- If these are genuine model errors, surfacing them is correct behavior and the work is documentation, not code.

## Components affected

- `src/simlin-engine/src/units_check.rs` -- the IF-branch consistency check and the log/binary-operator argument-units checks
- `src/simlin-engine/src/units_infer.rs` -- coefficient unit cancellation during inference

## Reproduce

```
cargo test -p simlin-engine --test clearn_unit_errors -- --ignored --nocapture dump_clearn_unit_diagnostics
```

## Context

Discovered while fixing the C-LEARN unit-error flood on branch `fix-clearn-unit-errors` (481 -> 14 diagnostics via four root-cause fixes). These ~14 are explicitly out of scope for that branch -- they require a Vensim-parity semantics decision per message class, not a mechanical fix. Distinct from #454 (resource_unit/resource_units alias), #455 (year-vs-implicit-time stock unification), #456 (duplicate_unit per stdlib instance), #457 (stdlib module output units not propagated), and #474 (the order-dependence that makes the umbrella diagnostic nondeterministic). Related: #35 (general unit-checking tracker).

bpowers:
Fixes #613

## Why

Sweeping the full test corpus (metasd, test-models, xmutil; 153 models) through `simlin simulate` surfaced **307 unit diagnostics, most of them our fault**: false positives from wrong builtin unit semantics, identical messages repeated once per array element, and messages that named neither the expected units nor the offending source. After this branch the sweep produces **44 diagnostics, each verified to trace to genuine unit sloppiness in the model itself** -- and **C-LEARN drops from 14 warnings to 0, matching real Vensim exactly** (which reports no mismatches on it). WORLD3, Theil_2011, thyroid-2008-d, the teacup XMILE samples, and the arithmetics fixtures also now check clean.

This resolves #613's open question in an unexpected direction: the C-LEARN residual was neither Vensim leniency nor model error -- all three message classes traced to checker bugs (the `^`-exponent handling behind the `ph` log-formula warnings and the permafrost coefficient non-cancellation, and IF-branch literal handling behind the rest).

## Semantics fixes (mirrored in `units_check` and `units_infer`)

- **A bare numeric literal is unit-polymorphic** through MAX/MIN/PREVIOUS, IF-THEN-ELSE, SAFEDIV's fallback, and SSHAPE (`Units::first_explicit`). The old code returned the *first* argument's verdict, so `MAX(0, land)/dev_time` computed as `1/year` instead of `hectare/year` (WORLD3's only warnings) and `IF t<t0 THEN 0 ELSE flow` collapsed to dmnl (Homer's Covid19US).
- **SQRT halves unit exponents** (Vensim fn_sqrt: `SQRT(units*units) --> units`); a non-perfect-square argument warns in checking and degrades to unknown in inference (a squared metavariable is unsolvable).
- **`x^n` with a literal exponent** multiplies unit exponents; half-integer exponents (`x^0.5`, `x^-0.5`, `x^1.5`) root the doubled map; other exponent shapes degrade to polymorphic rather than wrongly keeping `x`'s units (Theil's `rmse` flagged a dimensionally *correct* model). One shared `units::power_units` decision serves both files so they cannot drift.

## Policy and UX fixes

- **Unit checking is opt-in by declaring units**: a model declaring units on no variable gets no consistency diagnostics either -- the arrayed element-consistency pass used to fire off sim_specs `time_units` alone, flooding purely numeric fixtures.
- **Identical (variable, code, details) rows dedup** before emission (keyed on what the user reads, not the Loc-bearing Display form): arrayed variables repeated one identical mismatch per element (C-LEARN's `ph` 6x, scirev 50x).
- **Mismatch messages name both sides**: "the equation computes to units 'X', but the variable's specified units are 'Y'".
- **Multi-word unit names parse** (XMILE 3.3.6: names are identifiers "stored with underscores but generally presented to users with spaces"; Stella writes the presentation form -- the canonical teacup model died with a bare `extra_token`). The join is byte-length-preserving so parse-error offsets stay aligned with the original string, and definition errors quote the offending units string.
- **The builtin unit table matches XMILE 3.3.6**: time-unit abbreviations (`s`/`min`/`hr`/`wk`/`mo`/`qtr`), quarters, sub-second units, the `per_X` derived units, and `unitless` as dimensionless -- merged with Vensim's default `22:` synonym groups as before. Model-defined units precede equation-bearing builtins so `per_week` resolves against a model's own `week = day` definition.

## Non-obvious decisions

- Vensim's docs (Units Synonyms, 20705.html) confirm it does **not** auto-equate singular/plural beyond its default synonym list, so the remaining corpus warnings (Lotka's `Hares`/`Hare`, pendulum's `Kilograms`/`Kilogram`, workforce, SIR) are true positives Vensim would flag too -- deliberately left in place.
- Vensim's behavior for a **non-literal or non-half-integer `^` exponent is unverified**, so those shapes degrade silently to polymorphic rather than warning.
- `UnitMap::exp` now clears on `x^0` (a `{meter: 0}` map printed as dmnl while comparing unequal to dmnl) and saturates the multiply (model files are untrusted input; an overflow panic would abort `panic=abort` hosts).
- The third commit addresses the findings of an adversarial multi-agent review of the first two (dedup key, per_* ordering, exp overflow, half-integers, SAFEDIV/SSHAPE, offset preservation, and corrected XMILE section citations).


bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

bpowers:
@codex review

Issue #799, opened 2026-06-19
diagram: finish design-token consolidation (shadow tokens, px->rem typography, remaining hardcoded colors, toolbar-spacer token)


bpowers:
## Problem

A frontend UX audit (branch `frontend-ux-audit-improvements`) began a design-token consolidation but only finished part of it. The completed work added tokens to `src/diagram/theme.css` (`--color-surface` #fff, `--color-border` #ddd, `--color-field-bg` #f5f5f5, `--shadow-sm/md/lg` matching the three repeated MUI elevation triplets, `--panel-top-inset`, and the filled-in `--spacing-4..-9` scale) and adopted them in the editor chrome (search bar; zoom/undo/snapshot cards; the three detail-panel cards) plus `Button` (contained), `SpeedDial`, `Snackbar`, `Paper`, and `Card`.

The audit rated the full sweep "large" and the **remainder is untracked**:

1. **Hardcoded colors not yet on tokens.** ~a dozen components still hardcode `#fff` backgrounds, grey text (`#666`/`#888`/`#999`/`#aaa`/`#333`/`#777`), and dividers (`#ddd`/`#eee`/`#ccc`) instead of the `theme.css` tokens.

2. **Missing elevation-shadow tokens for the distinct MUI elevations.** `AppBar`, `Menu`, `Dialog`, and `Drawer` use DISTINCT MUI elevation shadows (elevation 4/8/16/24) that have no token yet. Add `--shadow-4`/`--shadow-8`/etc. (or a renamed equivalent) and adopt them. (Note: `--shadow-sm/md/lg` from the done work only cover the three repeated elevation triplets, not these four distinct levels.)

3. **Raw-px text sizing inconsistent with the rem-based typography system.** Editor chrome and detail panels size text in raw px (11/12/13/14/16px in `Editor.module.css` and `ModuleDetails.module.css`), so it does not honor user zoom. Add a `caption` typography variant (0.75rem) and convert these raw-px sizes to rem so they scale with user zoom.

4. **`.toolbarSpacer` hardcoded height coupled to the dense Toolbar.** `src/app/Home.module.css` `.toolbarSpacer` is a hardcoded `48px` coupled to the dense `Toolbar` height. It should derive from a shared `--toolbar-dense-height` token (or render an actual spacer `Toolbar`) so the two cannot drift apart.

## Why it matters

[...]

PR #1012, opened 2026-08-09
tests: remove implementation-asserting tests repo-wide

bpowers:
## Why

A sweep of every test suite (simlin-engine, five other Rust crates, six TypeScript packages, pysimlin) for tests that assert implementation rather than behavior: re-assertions of source literals, tests of local copies of production logic, strict duplicates of stronger tests, and tests that cannot fail (zero assertions, swallowed exceptions, tautologies). Such tests must be edited whenever the source changes even when behavior is preserved, so they impose maintenance cost while constraining nothing.

Deleting them exposed a second layer: production code whose only callers were the deleted tests, and test-support surfaces compiled into release binaries. Net: **246 files, +2,421/−14,338 (about 660 tests deleted, 15 added or rewritten to be real)**.

## Highlights by area

- **simlin-engine**: 41 `_monolithic`/`_vm` twin pairs deleted -- both helpers compile through the same incremental path since the salsa unification, so each pair ran one path twice (the stale `_incremental` helper spellings are renamed too). The mdl literal-echo cluster, the `stack_effect` transcription family (38 of 100 opcodes -- a sample, not an enumeration), the circularly-self-verifying `db/differential_tests.rs`, nine duplicate corpus registrations in `simulate.rs`, and ~20 genuinely dead prod items including both 145-line `strip_loc` walks, `get_incoming_links`, and the `ltm_post` scalar pair whose rustdoc claimed an FFI consumer that does not exist.
- **diagram (TS)**: class-name assertions restated component source verbatim (the test build sets `localIdentName: '[local]'`); two whole files tested local copies of production logic -- one encoded a *wrong* copy of the drag-select predicate while passing green.
- **engine (TS)**: the layered backend suites (direct/worker/wasm) re-tested identical operations; zero-assertion try/catch tests; test-only `worker-server` getters removed after rewriting their tests to observe behavior.
- **simlin-serve**: six dead prod functions (`check_and_increment`, `refresh_meta`, `save_to_disk`, `diagnostics_set_changed`, `baseline_from_errors`, `datamodel_to_canonical_json`) were kept compiling solely by their own tests; three helper duplications consolidated (forward-slash x3, extension dispatcher x2, sidecar rule x2); ~2.4s of pure test sleeps removed.
- **libsimlin/mcp**: a test that greps `patch.rs` for a substring; tests that pass whether or not the patch applied; the test-only `errors.rs` re-export module deleted after migrating the uncovered `format_diagnostic` arms into a 7-row table in the engine.
- **pysimlin**: constructor/frozen/repr echoes and exception-swallowing tests; dead `LoopPolarity.from_runtime_scores` + threshold constant (hand-duplicated Rust logic, zero callers), `ErrorDetailDict`, `error_code_to_string`.

## Tests strengthened or added instead of deleted

- The 20 `unit_checking_test.rs` tests whose only assertion was "it compiles" now chain `assert_no_unit_diagnostics()` -- mutation-tested: breaking `model_time_units` reds 6 of them and nothing else in the file.
- The invariance oracle now asserts a non-empty invariant set per caller (a classifier hoisting nothing previously passed all seven corpus gates).
- pysimlin's swallowing sim tests replaced with exact-value tests; polarity enums now asserted against the C header constants (a real FFI contract, mutation-checked); the untested `explain()` Module arm covered.
- `IconButton`'s documented href/disabled invariant (diagram CLAUDE.md) gained its first test, mutation-checked against the source line.
- wasm backend: the deleted handle>0 pins were replaced by two behavioral tests (default-engine-is-VM, per-sim state isolation) that pin contracts previously covered nowhere.
- `EditOperation` conversion is now a 5-arm table (the `RemoveVariable` arm had no test anywhere in the repo).
- The MCP test double `TestFileSystemAccess` is now a type alias for production `FileSystemAccess` (it had silently diverged exactly where production is non-trivial), and the SD-AI relationships-regeneration path has its first test.
- `race.test.ts` now asserts the patches it claims are ordered actually landed.

## Real defects found and fixed by the sweep

- VDF byte readers (`read_u32`/`u16`/`f32`) violated their documented totality: `offset..offset+4` could panic on overflow in debug builds (load-bearing under libsimlin's panic=abort). Now `checked_add`.
- `docs/sdai-model.schema.json` was 139 lines stale -- caught the moment the schema test became a drift guard instead of a test that *mutates the checked-in file on every `cargo test` run*.
- `format_diagnostic` leaked the `"<unknown>"` placeholder into the structured `variable_name` field on the Unit arm (the Equation arm correctly passed `None` through).
- The pysimlin version-tag guard was vacuously skipped in every CI run (shallow clones fetch no tags); CI now uses `fetch-tags: true` and the guard is verified to pass against the real tags.
- Engine `test_common`, serve/mcp-core `test_support`, and the 1.3k-line `layout/eval_stats` no longer compile into release binaries (feature-gated; verified with `nm` on the built artifacts).
- Cross-crate: an unrelated single edit invalidating things aside, the `simlin-serve`/`simlin-mcp-core` `test-support` features must land together (serve's dev-dependency names mcp-core's feature).

## Known judgment calls / surfaced-not-fixed

- `design-tokens.test.ts` (app) deleted: it cannot detect the drift it names but breaks on token renames. Cheap to reverse if you want that #799-era guard back.
- The empty-view reducer contract ("reducers return NaN for empty views, ArraySum returns 0.0") is pinned by nothing: all four `view.size() == 0` guards in vm.rs can be deleted with the suite green, and the wasm lowering restates the contract with no parity test. Closing it needs hand-assembled bytecode (a zero-element dimension cannot be built through the pipeline), which would swamp this branch -- documented on the renamed `reduce_view_tests` module instead.
- Engine TS test files are neither linted nor type-checked (pre-existing, #899); dead imports were cleaned by hand in every touched file.


docs/design/ltm--loops-that-matter.md
```diff
@@ -788,8 +788,10 @@ surfaces handle it differently:
   from the engine primitive (bound as `Sim.get_loops_runtime` ->
   `reclassify_loops_from_results`, GH #679/#685, the all-slots Rust source of
   truth) and attaches the per-step relative-score series on top. There is no
-  separate Python slot-0 reclassification path. (`LoopPolarity.from_runtime_scores`
-  survives only as a standalone scalar-array convenience utility.)
+  separate Python slot-0 reclassification path. (Python's own
+  `LoopPolarity.from_runtime_scores` -- a standalone scalar-array port of the
+  Rust classifier that no production path called -- has been removed, so the
```
bpowers: good documentation is _NEVER_ a changelog.  it is evergreen writing about the current state of the source code. people can dig through git history to find the changelog.

PR #1013, opened 2026-08-09
doc: align version-guard, loop-helper, and reclassify docs with the code

bpowers:
Follow-up to #1012, addressing the three codex review findings that arrived as it merged. All three are the stale-claim class: documentation describing a state of the world the code no longer has.

- **`simlin-mcp` version guard**: the comment and skip message described CI as a tagless shallow clone where the guard "is a no-op" — the opposite of what `ci.yaml`'s `fetch-tags: true` checkout (added in #1012) does. An unexpected skip in CI would have been misread as documented-normal behavior. Both now state that CI fetches tags specifically so the guard runs, and that a skip is expected only on a local clone without them. (Verified live: with the fresh `pysimlin-v0.8.0` tag fetched, the guard passes against the 0.8.0 release.)
- **Engine module map**: `src/ltm/` still advertised the deleted top-level `detect_loops` helper; anyone following the documented surface would import a nonexistent API.
- **`reclassify_loops_from_results` rustdoc**: claimed pysimlin `Run.loops` "still reclassifies via its own Python `LoopPolarity.from_runtime_scores` mirror (slot-0 only)" and "exposes the all-slots engine path separately as `Run.loops_runtime`" — both false: `Run.loops` rides this helper via `Sim.get_loops_runtime`, there is no separate property, and the Python classifier no longer exists. The A2A-semantics note is now correctly a two-site comparison (this helper's all-slots concatenation vs discovery's strongest-path scalar) instead of three.


bpowers:
@codex review

Issue #610, opened 2026-05-21
engine: INT/MOD negative-number semantics diverge from Vensim (blocks tests/rounding); MOD uses rem_euclid not XMILE floored modulus


bpowers:
## Summary

The corpus fixture `test/test-models/tests/rounding/test_rounding.xmile` is commented out of `TEST_MODELS` in `src/simlin-engine/tests/simulate.rs` (around line 27). It is an all-scalar, single-stock RK-free model with a genuine Vensim DSS `output.tab` expected-output file and **no submodules** -- exactly the kind of fixture the corpus comparator handles -- so the only thing keeping it disabled is a numeric divergence on `INT` and `MOD` of negative numbers.

This was surfaced during the Phase 4 review of the wasm-backend work. The new wasm backend faithfully mirrors the bytecode VM's `floor`/`rem_euclid` semantics, so this is **correct wasm-vs-VM parity, not a wasm bug**. The divergence is purely engine-vs-Vensim and is **pre-existing** (not introduced by the wasm-backend work).

## The fixture and the divergence

`test_rounding.mdl` / `.xmile`:
```
StockA      = -10 + Time/10
test integer = INTEGER(StockA)     # INT
test modulo  = MODULO(StockA, 3)   # MOD, positive divisor 3
```

Vensim's expected `output.tab` (genuine Vensim DSS 7.3.4) vs current engine output:

| Time | StockA | Vensim `test integer` | engine INT | Vensim `test modulo` | engine MOD |
|------|--------|-----------------------|------------|----------------------|------------|
| 0    | -10    | -10                   | -10        | **-1**               | **2**      |
| 1    | -9.9   | **-9**                | **-10**    | **-0.9**             | **2.1**    |
| 11   | -8.9   | **-8**                | **-9**     | **-2.9**             | **2.1**    |

- `INT(-9.9)`: engine `-10` (floor, `a.floor()` in `vm.rs` `apply()`, `BuiltinId::Int`) vs Vensim `-9` (truncate toward zero).
- `-9.9 mod 3`: engine `+2.1` (`l.rem_euclid(r)` in `vm.rs` `eval_op2`, `Op2::Mod`) vs Vensim `-0.9` (sign follows dividend).

These are deliberate, pinned engine semantics today -- see `vm.rs` tests `apply_int_floors` (`INT(-3.2) == -4.0`) and the `Mod(-7, 3) == 2.0` assertions with the comment "rem_euclid always returns a non-negative result."

[...]

PR #1015, opened 2026-08-10
engine: add ROUND builtin with round-half-to-even semantics

bpowers:
Adds a `ROUND(x)` builtin that rounds to the nearest integer with exact .5 ties going to the even neighbor -- Python `round()` / IEEE roundTiesToEven -- wired through every layer of the engine plus the wasm backend.

## Why ties-to-even

The XMILE v1.0 spec defines no ROUND function (zero occurrences in the in-repo spec), so the semantics are Simlin's to choose. Ties-to-even matches Python and is statistically unbiased over uniformly distributed ties. Two external-tool caveats are documented at the `BuiltinFn::Round` variant rather than left implicit:

- **Stella** does define ROUND (tie rule unspecified in isee's docs), so a Stella-authored `.stmx` calling ROUND now imports and simulates instead of failing with `UnknownBuiltin`; whether Stella agrees at exact ties is unverified.
- **Vensim** defines no ROUND at all, so the MDL writer records an `ExportWarning` when an equation calls it (the transpose operator's exact contract): there is no Vensim-expressible round-half-to-even, so the call is emitted as-is and disclosed instead of degrading silently.

## Implementation

- **VM**: a single `f64::round_ties_even` in `apply()`.
- **wasm backend**: the single native `f64.nearest` instruction, which the wasm spec defines as IEEE roundToIntegralTiesToEven -- the two backends agree bit for bit by construction, signed-zero ties included.
- Wired through the parser arity table, all four AST stages, lowering, codegen, array-operand materialization, run-invariance classification (pure), LTM polarity (non-decreasing step function, like INT), units checking/inference (units-preserving), the patch reprinter, the pretty printer, and the MCP function-reference docs. Constant folding deliberately leaves ROUND unfolded, matching the existing stance for ABS/INT/MIN/MAX.

## Tests

One shared case table (`test_common::ROUND_TIES_TO_EVEN_CASES`, bit-pattern comparison contract) drives all three backend tests -- the end-to-end parse-&gt;compile-&gt;VM pipeline, the VM `apply()` unit test, and the wasm parity test (which executes the emitted blob under the wasm interpreter) -- so the backends cannot drift apart on rows one test carries and another omits. The exact .5 ties are the load-bearing rows: `f64::round` (ties away from zero) agrees on every non-tie input, so a suite without them would pass with the wrong function. Further coverage: elementwise arrays, ROUND under a reducer (temp materialization path), NaN/infinity propagation, case-insensitivity, arity errors, units preservation, MDL export warnings, and the patch print-reparse round trip.

## Review findings folded in

`/code-review` ran on the branch; all six findings were verified and fixed:
- corrected a doc claim this branch introduced (Vensim's INTEGER truncates toward zero -- confirmed against the checked-in Vensim DSS ground truth -- rather than flooring; the skill table now states the real INTEGER/INT divergence, and the underlying import-fidelity gap was found to be already tracked as #610, where the importer-specific framing was added as a comment)
- the MDL ROUND export warning described above
- the Stella citation on the variant comment
- consolidated three drifting per-backend test tables into the one shared table
- listed the new test module in the engine CLAUDE.md
- deleted the dead `project_io.proto` `BuiltinId` enum (referenced by no message and no code, with a comment claiming bytecode serialization that never happens) and regenerated bindings; an unreferenced enum contributes nothing to any serialized instance, so wire compatibility is unaffected

https://claude.ai/code/session_01XjDuCGQqFp59hGWk2vf77G


bpowers:
Addressed the review feedback in fcf8a808:

- **XMILE citation (codex P2 on `round_builtin_tests.rs`)**: confirmed — the "zero occurrences" claim was false, an artifact of a grep that silently skipped the spec's non-UTF8 HTML as binary. A proper search finds 39 occurrences of the word, all prose (display precision, "rounded rectangle", "background"), and no `ROUND(` function form in the builtin catalog — so the substantive claim (XMILE defines no ROUND builtin) stands, and the comment now cites the verification that actually supports it.

- **Vensim lowering (codex P2 on `writer.rs`)**: agreed the "not Vensim-expressible" premise was overstated and fixed the stated rationale. The writer still deliberately emits `ROUND(...)` + `ExportWarning` rather than lowering: an IF THEN ELSE/INTEGER/MODULO composition duplicates the argument expression (wrong outright for stochastic arguments, which would be re-drawn per copy) and rests on Vensim edge semantics this repo has either not verified or knows to diverge (INTEGER truncates where our INT floors, MODULO disagrees with our MOD — #610 — and Vensim's `=` behavior at an exact .5 comparison is unverified). A subtly-wrong silent lowering is strictly worse than a loud warning; this matches the transpose operator's existing contract.

- **Stella tie semantics (codex P1)**: intentionally not changing behavior. Python's ties-to-even is the product requirement here regardless of Stella's (undocumented) tie rule — matching Stella is explicitly a non-goal right now. The `BuiltinFn::Round` comment now records this as a decision rather than an open caveat, and notes that if Stella fidelity ever becomes a goal, its tie rule should be verified against ground-truth output first.

- **Polarity enumeration gap (claude P3)**: fixed — `Round` is now a row in `test_builtin_polarity_monotone_increasing`, so all seven arms of the non-decreasing builtin group are pinned and moving Round out of that arm reds the test.

@codex review

Issue #1018, opened 2026-08-10
engine: the per-variable dimension narrowing matches raw equation spellings against display names, so a case-differing dimension reference is dropped before canonical matching


bpowers:
## Problem

`parse_source_variable_impl` (`src/simlin-engine/src/db/query.rs:179-190`) narrows the project's
dimensions to the ones a variable actually references, so an unrelated dimension edit does not
invalidate every cached parse. The narrowing compares **raw as-written equation spellings against
display names**:

```rust
let expanded = expand_maps_to_chains(relevant_dim_names, all_source_dims);
project_datamodel_dims(db, project)
    .iter()
    .filter(|d| expanded.contains(&d.name))   // <- display name, exact string match
```

`relevant_dim_names` is seeded from the equation's dimension references **as written**.
`expand_maps_to_chains` canonicalizes for *traversal* and resolves mapping **targets** back to
display names through `canonical_to_display` (`src/simlin-engine/src/db/sync.rs:781-790`) — but the
**seed names pass through untouched**. So a reference spelled `dima` against a dimension declared
`DimA` never matches, is filtered out of `dims`, and never reaches `variable::get_dimensions`'
canonical matching at all.

The variable is then parsed against a `DimensionsContext` that is missing a dimension it references.

## Why it is not caught today

`get_dimensions`' own docs cite GH #541 (a synthesized apply-to-all carrying canonical `hfc_type`
resolving against a declared `HFC type`), and that case works — but only because the declared name
survives the pre-filter, or because the dimension arrives as a **mapping target**, which
`expand_maps_to_chains` *does* display-resolve. The canonical matching in `get_dimensions` is
therefore doing real work in some cases and is unreachable in others, depending on how the dimension
entered the set.

## Impact

A model whose equation spells a dimension in different case from its declaration is rejected, where
canonicalization elsewhere in the engine would accept it. Case- and separator-insensitive identifier
matching is what `canonicalize` exists for, so this is an inconsistency rather than a deliberate
strictness.

## Why it is not a drive-by fix

Making the seed display-resolved **changes which models compile**: a currently-rejected model would
start compiling. That is a semantics change needing corpus validation (does anything currently
relying on the rejection break? does a newly-admitted model then fail later, with a worse
diagnostic?), not something to fold into an unrelated change.

## Suggested direction

Resolve the seed the same way targets already are — canonicalize each seed name and map it back
through `canonical_to_display` before the `expanded.contains(&d.name)` filter — so both halves of
[...]

Issue #1017, opened 2026-08-10
ltm: a repeated dimension in an owner routes to element-level Johnson and enumerates 34,904 circuits that are all declined (150x on an 8-element fixture)


bpowers:
## Summary

An owner variable that repeats a dimension (`cube[D1,D1,D2]`) makes the variable-level cycle
classify as `CrossElementOrMixed`, which routes it to the element-level Johnson slow path. A repeated
dimension pairs every `(i,j)` with every `(k,l)`, so a 16-element model produces a near-complete
16-node SCC and Johnson enumerates **34,904 elementary circuits** — all of which are subsequently
**declined**.

Enumeration itself is only ~13 ms. **~93% of the cost is downstream**, in `model_ltm_variables`
processing those 34,904 circuits (rotation dedup, `assign_loop_ids` sort, partition assignment)
before declining them.

## Measured

Controlled A/B where the only difference is whether the owner repeats a dimension. Element counts
are identical in both fixtures:

| fixture | `model_ltm_variables` | vars / arms | slow-path circuits | largest SCC |
|---|---|---|---|---|
| `cube[D1,D1,D2]` / `pop[D1,D1]` (square) | **191–224 ms** | 93 / 93 | **34,904** | 16 |
| `cube[D1,D3,D2]` / `pop[D1,D3]` (rectangular) | **1.4 ms** | 109 / 109 | **0** | 0 |
| square, **discovery mode** | 1.1 ms | 84 / 84 | — | — |

**~150x apart, and the square variant emits *fewer* variables and arms than the rectangular one.**
That rules out equation generation as the cause — the obvious first hypothesis — and is why `perf`
shows no hot loop: the work is 34,904 circuits × a few small allocations and salsa query-edge
inserts, diffuse by construction. The observed signature is diffuse string interning plus salsa
edge inserts with nothing above 2.6%.

Exhaustive mode only. Discovery mode never calls the tiered enumeration, which is why the same
fixture is 1.1 ms there.

## Why the existing guard does not catch it

`MAX_LTM_SCC_NODES` gates on **SCC node count**. The blowup is in **circuit count**, and the two are
only loosely related — a dense n-node SCC has worst-case ~(n-1)! elementary circuits. 16 nodes sails
through the node gate, and 34,904 circuits evidently does not trip `MAX_LTM_CIRCUITS` either.

A node-count gate cannot bound this in general: the quantity that needs bounding is the one being
enumerated.

## Impact

[...]

Issue #1016, opened 2026-08-10
ltm: should a ceteris-paribus partial freeze TIME? (today it does not, and the target's autonomous time evolution is attributed to the link)


bpowers:
## The question

Should a ceteris-paribus partial hold **TIME** at its previous value?

Today it does not. `db/ltm/compile.rs` (three sites, each
`|| matches!(effective, "time" | "dt" | "initial_time" | "final_time")`) skips those idents during
dependency processing, and the ceteris-paribus wrap freezes *dependencies*. So for a target whose
equation reads `time()` — directly or through `STEP`/`RAMP`/`PULSE` — the partial advances the clock
while holding every model input back, and the target's **autonomous time evolution is attributed to
the link**.

## Why the reference implies TIME should be frozen

`docs/reference/ltm--loops-that-matter.md:231-233` defines the quantity as

> **Delta_x(z)** = `f(x_current, y_previous) - z_previous`: the partial change in z due to x alone,
> computed by re-evaluating f with the current value of x but the *previous* values of all other
> inputs (ceteris paribus)

The document never mentions TIME, so this is an inference from the formula rather than a quotation.
But the formula only closes if "all other inputs" covers everything `f` reads that varies. Take a
source `x` that does not appear in `f`: then `f(x_current, y_previous)` reduces to `f(y_previous)`,
and the anchor `z_previous` is `f` at the previous step's inputs, so `Delta_x(z) = 0` identically.
Exclude TIME and that identity fails — a source with no influence whatsoever scores nonzero, which
also contradicts the reference's own edge case at `:249` ("If x does not change, the link score is
0").

## Evidence that this is what is happening

Measured on C-LEARN v77 with LTM enabled, by zeroing the arms whose model dependencies are all
frozen and diffing all 30,123 result slots × 251 saved steps:

- **35 of 35** link-score variables whose values move carry a live time-dependent term in their
  partial. No exceptions.
- **26 of them sit at exactly ±1.0** — which for a fully-frozen arm can only mean `|N| = |Δz|`, i.e.
  the entire change in the target came from something the freeze did not reach. In a fully-frozen
  partial the only such thing is the clock.
- The competing explanation for unit magnitude does not apply: the single-live-input case
  (`docs/design/ltm--loops-that-matter.md:391`, "for a single-input black box `z = F(x)` all of `Δz`
  is attributable to `x`") requires a live input, and these arms have none.

Two independent derivations converge: the algebra above, and the runtime measurement.

## Blast radius

[...]

Issue #977, opened 2026-07-20
ltm: link-score generation materializes one full guard-form arm per element of an arrayed target (quadratic in element count)


bpowers:
## Summary

For a per-element (`Ast::Arrayed`) target, a link score is shaped over the **whole** target: one arm
per target element. An arm whose element equation does not reference the link's source with the
link's access shape has every source reference frozen by the ceteris-paribus wrap, so it is intended
to evaluate to zero — but it is still materialized as a full guard form wrapping a full copy of that
element's equation, and it is generated, printed, parsed, lowered, and executed every timestep like
any other arm.

With N distinct source elements over an N-slot target that is `N^2+N+1` arms for a **single causal
edge**, of which `N^2-N` are the intended-zero kind. (The title's older `(N+1)^2` figure is off by
the cross terms; the measured count is `N^2+N+1`.)

## Scale

Synthetic fixture — dimension `Wide` with N elements; `growth[Wide]` a per-element
`Equation::Arrayed` flow whose element `e` equation is `pop[e] * rate * 0.01`; `pop[Wide]` an array
stock fed by `growth`; `rate` a scalar aux. Release build, timing `db::model_ltm_variables`:

| N | `model_ltm_variables` | link_score vars | total arms |
|---|---|---|---|
| 50 | 37.3 ms | 51 | 2,551 |
| 100 | 140.9 ms | 101 | 10,101 |
| 200 | 568.0 ms | 201 | 40,201 |
| 400 | 2.304 s | 401 | 160,401 |

Clean 4x per doubling. The salsa queries the emission reads are linear and negligible: at N=400,
`model_ltm_reference_sites` 5.2 ms + `model_element_causal_edges` 1.16 ms +
`model_loop_circuits_tiered` 0.81 ms = ~7 ms against 2,304 ms. The cost is in the emission body.

Real model — **C-LEARN v77, LTM enabled** (discovery mode; 7,001 synthetic variables). Of the 12,364
emitted guard-form arms, 10,570 (85.6%) come from `build_arrayed_link_score_equation`, and **9,514
(77.0%) have no live source reference at all** — 7.63 MB of 9.63 MB of arm text (79.3%). LTM takes
the root flow program from 31,525 to 1,208,106 opcodes, and these arms are the bulk of it.

**Scalar models are unaffected.** WORLD3-03 has zero such arms; its own LTM cost is entirely
per-link guard scaffolding, which is a different problem.

## Why the obvious collapse is unsafe, and what the safe form is

[...]

PR #1021, opened 2026-08-10
engine: performance pass on compile, simulation, and the LTM path

### LINKED ISSUE #1019: ci: nothing executes the wasm-opt'd bundle under test -- both automated lanes skip wasm-opt, and only deploy/release run it
## The gap

`src/engine/build.sh` runs `wasm-opt -O3` **in place** over `core/$out_name`, so whatever loads that
blob afterwards loads the optimized one. That means the TypeScript suite executes the optimized
bundle only where `wasm-opt` actually ran.

It does not run in either automated lane:

| caller of `pnpm build` | `wasm-opt` |
|---|---|
| `.github/workflows/ci.yaml:219` | **disabled** (`DISABLE_WASM_OPT=1`, deliberate — the comment cites the multi-minute pass) |
| `scripts/pre-commit` | **disabled** (170 s per engine-touching commit; see below) |
| `scripts/deploy-web.sh:68` | runs |
| `scripts/deploy-web-staged.sh:69` | runs |
| `.github/workflows/serve-release.yml:99` | runs |
| `.github/workflows/ts-release.yml:74` | runs |

The four that run it **build and ship** the bundle; none of them executes the engine test suite
against it. So the optimized bundle is never exercised under test anywhere.

Until the pre-commit change this was covered incidentally, on developer machines only — which is
not coverage that can be relied on: it depended on every developer running the hook, on a machine
with `wasm-opt` installed, and it was invisible in CI either way.

## Why the pre-commit hook stopped doing it

Measured on an engine change that genuinely alters emitted code (a probe that only adds an unused
item is stripped by LTO, leaves the wasm byte-identical, and `build.sh`'s `cmp` guard then skips
`wasm-opt` entirely — so it measures a build that never does the expensive part):

- `src/engine/build.sh`: **188.7 s**, of which **`wasm-opt` is ~170 s (90%)**
- the whole pre-commit hook on such a commit: **~200 s**, of which `pnpm build` is ~90%
- for comparison, `cargo test` in the same hook: ~12 s

Cost of running the tests against the unoptimized blob instead: **none measurable** — engine TS
suite 1.26 / 1.29 s unoptimized vs 1.40 / 1.42 s optimized. The suite is harness-dominated, not
wasm-execution-dominated.

170 s per engine-touching commit is far too much to pay for coverage that only some machines
provide. But the coverage itself is worth having somewhere.

## What would close it

A lane that builds **with** `wasm-opt` and then runs the engine TypeScript suite against the
resulting blob. It does not need to be on the PR gate — nightly, or path-filtered to engine and
`build.sh` changes, is enough, because what it guards against is a `wasm-opt` miscompilation or a
version incompatibility, not a per-PR regression.

Sizing, from the same measurements: the optimizing build is ~170 s more than the plain one, and the
suite it would then run is ~1.3 s.

## Related constraint, so a future fix does not take the wrong shape

**Do not close this by flipping the default in `build.sh` or `package.json`.** `wasm-opt` is off in
two callers and on in four; a default flip routes through the four deploy/release callers and ships
a browser bundle **24% larger** (5.00 → 6.20 MB) to every user. The size argument is recorded in
`.cargo/config.toml`'s comment about keeping the wasm target at `opt-level="z"`. The enable/disable
decision belongs to each caller, not to the build script.

## Also measured, so it is not re-proposed

Lowering the Rust optimization level to make the wasm build faster is **backwards**: `opt-level=1`
emits a 61% larger module (7.90 → 12.71 MB raw) and `wasm-opt`'s cost scales with input size.

| profile | cargo | wasm-opt | total |
|---|---|---|---|
| `opt-level=z` + LTO (today) | 34.5 s | 170.4 s | **205 s** |
| `opt-level=1` + LTO | 48.4 s | 295.5 s | 344 s |
| `opt-level=1`, no LTO | 41.3 s | 266.6 s | 308 s |

`opt-level=z` is both smaller and faster end to end.

bpowers:
A performance pass over `simlin-engine`: compilation and simulation, with and without LTM, on the two hero models. Every change is behaviour-preserving except one, which is called out below.

## Results

Interleaved A/B against the merge base, both binaries built and run in the same session. Wall-clock on a contended box is corroboration only; retired instructions are the load-bearing channel (see the methodology note added to `docs/design/engine-performance.md`).

| | compile | run |
|---|---:|---:|
| C-LEARN | **−20.0%** | −7.7% |
| C-LEARN + LTM | **−14.2%** | **−51.5%** |
| WORLD3-03 + LTM | −4.4% | **−43.1%** |

Retired instructions for the C-LEARN LTM run: **−41.1%**. LTM compile-stage peak live bytes: **438.8 → 353.2 MiB (−19.5%)**.

**Interactive edit latency** — a single-equation edit on C-LEARN, paired over 40 variables:

| | before | after |
|---|---:|---:|
| median | 2.8 ms | 2.8 ms |
| p90 | **36.0 ms** | **3.0 ms** |
| max | 73.9 ms | **6.7 ms** |
| retired instructions | 11.03G | **5.27G (−52%)** |

The median was already fine; the tail was the problem. A cheap edit is now 40.6M instructions, dominated by salsa's own `maybe_changed_after` verification plus re-lexing the one edited equation.

**Developer loop**: `cargo test --workspace` −25% on a 4-core CI shape (31.0 → 23.2 s), and ~170 s off every engine-touching pre-commit run.

## What changed

**Compile.** Six changes, all artifact-identical: a provably-identity `canonicalize` removed from a per-dimension predicate; `variable_dimensions` derived without a second parse; two previously-untracked functions (`compile_implicit_var_fragment`, the cycle gate's fragment probe) made salsa queries; FxHash on the topological-sort probe maps; and the engine's own `·` separator answered without a Unicode case-table lookup.

**VM.** Superinstructions, all created by `ByteCode::fuse_three_address` on the Vm's private execution copy unless noted: `SetCond;If[;AssignCurr]` (100% adjacent by construction — codegen is their sole producer and emits them together), leaf stores and a fusible module-input leaf, `LoadPrevConst`, `ApplyTerConst`, `SubVarPrev`, `BinStackPrev`. Two reach codegen and therefore wasmgen: real builtin arity instead of padding `Apply` to three operands, and a lookup's constant element offset resolved at compile time.

**LTM.** The direct fragment compile is memoized — 5,985 of 7,125 LTM variables previously took an uncached path and were compiled twice per diagnostics pass, three times in MCP `edit_model`. Equation ASTs are `Arc`-shared between the shaped memo and the emission loop. And link-score arms whose ceteris-paribus partial is *provably* `PREVIOUS(target)` are omitted rather than materialized: 4,335 arms on C-LEARN, −19.2% of the flow program.

## The one behaviour change

Arms are omitted only when the partial is provably `PREVIOUS(target)` **and lag-aligned** — every read exactly one step back. Lag alignment is enforced, not assumed: an original `PREVIOUS` in the target's own equation, or a synthesized `PREVIOUS` nested inside another, both break it and both are rejected. Each has its own gate row, mutation-tested per clause (reverting either check fails exactly one row and neither covers the other). On both hero models no arm was ever misaligned, which is why every earlier measurement was clean and why enforcing this costs zero arms — but that is a property of those models, not of the change.

Given that, the transformation is bit-exact except for the sign of zero. Omitted arms lower to `AssignCurr(off, Const(0.0))` = `+0.0`, where a materialized arm ends in `... * SIGN(Δx)` and can produce `−0.0`. Whole-slab differential over 7,560,873 slot-steps: 51,358 differ, in **exactly one bit-pattern pair** (`−0.0 → +0.0`), numerically zero. Nothing in the engine branches on sign-of-zero — `vm.rs` pins `eval_op2(Eq, 0.0, -0.0) == 1.0` and `!is_truthy(-0.0)`, `float.rs` pins `approx_eq(0.0, -0.0)`, and relative loop scores sum `|score|`. One cosmetic surface: a `simlin-cli` column that printed `-0` now prints `0`.

**Why the obvious cheaper version is wrong**, since it is what a reviewer will reach for: testing "the link's source stayed frozen" (`live_ref == None`) is a *different* question and is unsound — it changes 187 result slots across 35 link-score variables, 151 of them by ≥ 1.0, with a worst case of 8,086.97 → 0. Those arms are not structurally zero; they are scoring the clock, because TIME is excluded from the ceteris-paribus freeze. See #1016. The predicate here asks the positive question instead, and is stable under fixing that defect.

## Evidence, and its boundary

- Full suite green on the integrated tree: 5,505 lib + 648 integration + 4 alloc + 3 doctests, 0 failures. Every commit went through the full pre-commit hook; none used `--no-verify`.
- `clearn_residual_exactness`, `simulates_clearn`, `oracle_clearn`, `clearn_ltm_var_count_guardrail` green throughout.
- Compile and VM changes are **artifact-identical** — compiled bytecode byte-for-byte unchanged — except the two codegen changes, whose golden movement was reviewed opcode-by-opcode.
- LTM value-neutrality is established **on C-LEARN and WORLD3-03 only**: per-slot maxima and full opcode identity respectively. Not on arbitrary models, and not per-step for C-LEARN — the checked-in digest is maxima, not a slab. It is runnable rather than asserted:
  `cargo test -p simlin-engine --release --test integration -- --ignored clearn_ltm_slot_maxima_digest`
- That digest covers **all 20,892 LTM slots** across 7,153 variables, including every element of the 1,088 arrayed ones. The accounting closes exactly: 21,011 slots described by LTM metadata, 119 of them never saved. It also pins at 0 the count of LTM result slots whose extent the metadata cannot describe, so a population it cannot see fails rather than being silently skipped.
- Its identity component is an FNV-1a over the canonically ordered `(name, element, mantissa, exponent)` stream, and **two separate tests constrain two separate properties**: a value swap (maxima exchanged, order fixed) pins that the fold is ordered, and a rebinding (same maxima, same order, different name or element index) pins that the identity bytes are actually read. The second exists because the first does not imply it — an ordered fold moves under a swap whether or not identity is in the stream, and a rebinding is the closer analogue of the offset-remapping regression this is meant to catch, since it leaves both the multiset and its order intact.
- The four `#[ignore]`d `simulates_delayfixed*` failures are pre-existing (#346), verified by reverting to base.

## New tests, because three gaps were found

- **No value-level LTM gate existed at any scale.** A change that rewrote 149 C-LEARN LTM slots to zero passed every named C-LEARN gate — `clearn_residual_exactness` never enables LTM, and the var-count guardrail moves on count and width, neither of which arm omission touches. Added: a sub-second default-suite gate with one arm per known failure mechanism, plus an `#[ignore]`d C-LEARN slot-maxima digest. The gate includes a **positive control** — forcing the predicate true drops 82 real scores to zero and reds the digest — so a passing run cannot mean a blind instrument.
- **The omission path had no characterization coverage**: of sixteen golden fixtures exactly one had an arrayed link score, pinned to the materialize path.
- **`STACK_CAPACITY` is 64 and the deepest stack in the corpus is 8**, so a wrong `stack_effect` has 5x headroom to hide in and neither a passing suite nor a matching fingerprint reaches it. Added a corpus audit asserting fusion never raises peak depth and neither stream underflows.
- Also pinned: which slots carry information across a simulation step. Two separate optimization attempts have depended on that invariant with nothing guarding it.

## Measured and declined

Recorded so they are not re-proposed:

- **Raising the engine's dev `opt-level`**: +144 s per engine-touching CI build, and it would push the Rust pre-commit pipeline past the TypeScript one — the hook is `max(A,B,C)` with 24 s of slack — making the developer's hook *slower* while saving time on a pipeline that is not the critical path. A loss on both sides.
- **A faster Rust profile for the wasm build**: 50–67% *slower* end to end. `opt-level=1` emits a 61% larger module and `wasm-opt`'s cost scales with input size, so `opt-level=z` is both smaller and faster.
- **Removing the duplicate LTM parse**: −7.3% allocations for **+82.2 MiB** peak, which gives back the entire memory win. The duplicate parse is now documented as deliberate, with the numbers.
- **Chunk-index swapping instead of copying**: the premise ("everything is rewritten each step") is false — the implicit `dt`/`initial_time`/`final_time` prefix is established once by `run_initials`.
- **Hoisting repeated LTM subexpressions into helper auxes**: worth ~zero once scored against the *post-fusion* stream, and a net loss for `ABS`/`SIGN`. A helper read and a fused opcode are both one dispatch.
- **An O(1) index for uniform-grid graphical functions**: predicted −2.7% / −3.1% of run instructions, measured **−0.63% on C-LEARN and +0.59% on WORLD3** — the wrong sign, at ~22 sigma, so a real regression rather than noise. The mechanism fired (per-call lookup cost −34.8% on C-LEARN) but made WORLD3 7.7% *worse*. The error is worth recording: **eligibility is not profitability**. 91% of corpus tables *could* take the fast path; profitability is set by table size, because the guess costs ~50 fixed instructions against ~12 per search probe and only pays above ~4 probes — and WORLD3's median table is 7 points. Reverted after three attempts, per the rule that a repeated regression is a scope signal rather than something to try harder at.

## Documentation

`docs/design/engine-performance.md` gains a measurement-methodology section, and it corrects a conflation that had been doing damage: the recorded "~4% noise floor" is a **cycles** floor and does not bind retired instructions, whose measured sd across six independent builds is 0.026%. That conflation had already carried a verdict — the #711 lazy-`IF` NO-GO cites "below the ~4% layout-noise measurement floor" for a 1.5% *instruction* share, which is ~58 sigma. The verdict stands for other reasons; only its stated reason is corrected, here and on the issue.

Evidence for the floors is a **null control**: the same binary as both sides of an interleaved A/B reports −0.003% on instructions and **−1.540% on cycles** — a "win" from measuring nothing.

## Costs

`Vm::new` is slower — 5.67 → 8.59 ms on C-LEARN+LTM — because the fusion pass scans more windows over a larger program. It is 3 ms against a 943 ms run.

## An open question this surfaced but did not answer

**1,369 of 7,000 C-LEARN LTM slots are ever non-zero — 81% are identically zero across the whole run** in discovery mode. Uninvestigated. It is either uninteresting (many causal edges genuinely score zero) or the next large lever, and leaving it unstated guarantees nobody finds out which.

## Discovered and filed rather than fixed here

#1016 (should a ceteris-paribus partial freeze TIME? — worth ~25 further points of the LTM flow program, but it moves ~41% of C-LEARN's link scores and needs dominance-ranking validation), #1017 (a repeated dimension enumerates 34,904 circuits that are all declined), #1018 (dimension narrowing matches raw spellings against display names). #977 was rewritten around this work and its stale 322-arm row corrected to 0 with the measurement attached. #715 and #915 gained corpus facts that make them concrete.

## What review caught

Two `/code-review` passes at high effort over the whole diff, nine findings, all resolved. The second pass verified the risky engine changes independently — every fused opcode's stack effect against the sequence it replaces, operand pop order including the non-commutative cases, jump guards and `pc_map` 1:1-ness, builtin arity across the VM, codegen and wasm arm-for-arm, and that both lag-alignment checks are load-bearing — and **found no correctness defect in them**.

Every finding was verified against the code before being acted on. **Four came back different from how they were reported**, and in three of those the difference changed the fix:

- **A wasm build-cache hole**: mechanism real, user-facing consequence **refuted** — both deploy scripts clean first, so no unoptimized bundle reaches users. Fixed anyway, because that safety is incidental to two call sites rather than a property of the cache.
- **The merge proptests passed vacuously** for the new lookup opcode — but the reported cause (catch-all arms in the oracles) was wrong: those arms were *unreachable*, because neither generator constructs the opcode. Fixing only the oracles would have left the property unverified while making it look covered. Verified by mutation: twelve properties green before, four red after.
- **Two new corpus tests were reported as reintroducing a serial mega-test floor.** Measured: they run at 0.145s and 0.150s against a wall pinned by a 3.181s test. Never the floor. Split anyway for diagnosis, with the measurement recorded so nobody reads a performance claim into it — and the split turned a corpus-wide `assert!(checked > 100)` into a per-model `assert!(checked > 0)`, closing a guard that could itself pass vacuously.
- **A hard `jq` dependency** on the primary build path: removed rather than accommodated, since `python3` is already required earlier on the same path.

The two that were straightforwardly right were the two that mattered most:

- **`variable_dimensions` diverged on valid models**, not only on ones that already fail to compile — two shapes went from 1 slot to N. A 483-model scan found zero shipped models affected, so artifact-identity held *by corpus luck rather than by construction*. Gating on the mechanism (an equation that lexes to zero tokens, which is the parser's own early return) rather than on the two shapes found immediately turned up a third — comment-only equations — that the natural fix would have answered wrongly.
- **The LTM predicate was unsound**: it treated `PREVIOUS` as step-invariant, so an arm whose target equation already contained one could be dropped despite scoring 0.985. A second shape was found that the review missed — and its numbers had been sitting in the tree for weeks, documented as an open semantics question while also being unread evidence that such an arm is not zero. Both shapes now have gate rows, mutation-tested per clause. The fix costs zero arms and let an earlier workaround be deleted.

A third round found five more, four of them the same pattern and two of them in the digest that is this branch's primary evidence: it sampled **one element per arrayed variable** (7,000 slots of 20,892) while documenting itself as covering every slot, and its unweighted sums were **permutation-invariant**, so values attached to the wrong slots would not have moved it. Widening it showed the gap was not theoretical — 1,772 slots carrying real scores were invisible, and the gate was detecting an eighth of the damage its own positive control inflicts (82 of 614). Re-running value-neutrality at the full coverage: every pinned number identical with the omission disabled, so the central claim is now supported roughly three times as well as when it was first made. The others: a stale "optimized" stamp could survive a failed optimization and make the next build treat a raw blob as optimized — **and the guard meant to catch that read the same stamp**, so the backstop shared the failure mode of the thing it backed up; the installer the optimized-WASM lane depends on was not in its own path filters; and two more copies of the stale fixed-arity comment, one of them in the file that *defines* the opcode, sitting immediately above the line that replaced it.

One pattern ran through all of it, worth naming because it is the round's most reliable defect source: **every one of these was prose asserting a property the code did not have** — a noise floor applied to the wrong channel, "bit-exact" without lag alignment, "only unparseable equations", "nine significant digits" on a digest that quantized absolutely. Prose is the one artifact nothing executes. The durable fix, applied repeatedly here, is to make the claim executable or make its violation a compile error.

## The named next round

Nothing here is started, and each carries its measurements so it does not begin from zero:

- **#1016** — the TIME semantics question. Worth roughly doubling the LTM arm-omission win, but it moves ~41% of C-LEARN's link scores and needs dominance-ranking validation.
- **#372** — context-stable helper naming: the remaining interactive-edit lever. Not the dependency graph and not the fragment compilers, which are the symptom; a structural edit re-keys an interned context handle and a new key cannot backdate. Not artifact-identical, so it wants a deliberately reviewed golden re-baseline.
- **C4, the parallel compile fan-out** — prototyped at 1.34x wall against a measured 1.44x Amdahl ceiling, deferred to its own PR. Its two hazards are recorded as MUSTs in the design doc; both are silent and one is a process abort.
- **The per-loop iteration access-plan cache** — ~3-5% of the C-LEARN run, the largest single run-side item found and not started.
- **#1017**, **#1018**, **#715** — each now carrying a concrete reproduction rather than a hypothesis.

Fixes #1019


bpowers:
All four verified against the code and fixed in `2afe6265`. Each was the same defect class — an assertion the code did not implement — which is the one this PR's own review passes kept turning up, so they were worth taking seriously rather than filing.

**Harness allocators.** Correct, and the sharper version is that *a dev-dependency does not select an allocator at all*. Only `clearn_profile` installed it; `backend_bench` and `ltm_mem_bench` kept `System` as their counting allocator's backing, and the four benches declared none. So the compile-path timings those harnesses report came from precisely the allocator the comment calls unrepresentative — over-crediting the changes it warns about. Fixed by making the claim true rather than narrowing it: all seven harnesses install it now, since the `compiler`/`simulation` benches are the ones whose subject is the allocation-bound compile path. The comment now also says a new harness must install it too, which is the part that would otherwise decay again.

**`SIMLIN_STATIC_LIB`.** Correct, and worse than stated: `_get_library_path`'s own docstring says the pin exists because *"guessing wrong here silently links a stale engine into the extension (GH #682)"* — so the script resolved the archive for its freshness check and then dropped it at the one call that consumes it. Now pinned to the same archive line 14 built. I also removed the `2>/dev/null || true`: a failed CFFI build left no extension, and the import error that followed said nothing about the cause. Verified end-to-end — the pre-commit pysimlin stage passes with failures no longer suppressed.

**`rust-toolchain.toml` in the path filters.** Correct. Added to both lists with the reason recorded inline: the compiler selects what WASM is emitted and whether binaryen can still read it, and the ordinary frontend lane runs `DISABLE_WASM_OPT=1`, so a toolchain bump could reach a release without either blob having been optimized once — which is the exact gap this workflow was added to close.

**The diagnostic probe's comment.** Correct. It still described `compile_implicit_var_fragment` as untracked and quoted a ~15 ms per-revision helper recompile, which is the opposite of what the query conversion establishes. Rewritten to describe per-helper memo reuse and what actually invalidates it, plus a note not to remove the walk on the assumption it is doing the compiling — it is the accumulator replay that needs it.

One thing found while fixing: `compile_ltm_fragment_for` reads its `expected` argument only from a debug assertion, so release builds emitted an unused-variable warning that the debug-profile clippy gate never sees. Silenced with the reason attached, since the argument is what forces a caller to hold the variable and is therefore the coupling being checked.

Full pre-commit gate green (dependency policy, doc links, project lint, Rust fmt/clippy/test, TypeScript + WASM build, pysimlin).

@codex review

bpowers:
Both review comments addressed, and both CI failures fixed at the root, in `b8b3af87`.

## P1 — non-finite arms: confirmed, disclosed and pinned; the decision is now #1022

Correct, and reproduced rather than reasoned about. Fixture: a flow-less stock holding `0` (a plain constant folds away and the target never goes non-finite), `nan_src = zed / zed`, `growth[boston] = nan_src * 0.02`.

| | boston slot |
|---|---|
| materialized | `[0, NaN, NaN, NaN, NaN, NaN]` |
| omitted | `[0, 0, 0, 0, 0, 0]` |

Your mechanism is exact, including the part that makes the guards useless here: `SAFEDIV`'s fallback is for a **zero** denominator, not a `NaN` one, so it returns `NaN/NaN`. Infinity collapses to the same case since `inf - inf` is `NaN`.

**On the remedy, though — "keep such arms materialized unless finiteness is guaranteed at runtime" is not implementable.** Non-finiteness is a runtime property, so declining to omit anything that *could* go non-finite means declining to omit at all, which is the entire optimization. A cheap runtime sentinel does not close it either: `0 * Δtarget * SIGN(Δsource)` takes sign-of-zero from the wrong delta, yields `NaN` at `TIME = INITIAL_TIME` where the materialized arm is exactly `0`, and costs 4–5 opcodes against the 1 an omitted slot lowers to. Anything short of exact equivalence is a second approximation stacked on the one under question.

So: **disclosed and pinned, and the semantics decision filed as #1022** rather than settled inside a performance PR. `a_nonfinite_target_arm_is_omitted_to_zero_not_nan` asserts the fixture premise (the target really is `NaN`, so it cannot pass vacuously), the omitted zeros, and a counterweight — a *live* arm over the same `NaN` target still scores `NaN`, which bounds the change to arms with no causal dependence on the source. Mutation-tested. Its rustdoc names it as the one gate row that pins the omission *changing* a value rather than preserving one, so it does not read as an anomaly and get "fixed".

Worth surfacing, because it cuts against the direction your comment and a previous attempt both leaned: **`ltm_post::denom_summand` already excludes a `NaN` summand from its partition denominator** specifically so the bad entry's own numerator stays `NaN` — "the honest per-loop 'undefined here' signal" (GH #542). That is an adjudicated decision that a `NaN` score carries meaning per entry, and it also means a `NaN` link score does *not* poison its partition — removing the strongest available argument for `0`. Two independent looks called `0` "arguably right" citing only `float.rs`; neither weighed #542. #1022 records the distinction that actually needs adjudicating: whether *undefined target* or *no causal dependence* dominates when both hold.

## P3 — stale `Apply` dispatch comment

Correct, fixed. It now says the pop count is `func.arity()` from the table shared by codegen, the VM and this backend, and states the consequence you identified: locals above the arity hold whatever the previous `Apply` left there, so they must not be read.

## CI

**The optimized-WASM lane caught its own install recipe on its first run.** `apt-get install binaryen` ships older than the flags `build.sh` passes, so the build died on `Unknown option '--enable-bulk-memory-opt'`. That recipe was copied from `ts-release.yml` — the npm publish workflow — which would have failed identically on the next release. Both now share `scripts/install-binaryen.sh`, which pins an upstream release, verifies that version accepts the flags `build.sh` passes, and fails at *install* rather than after `build.sh` has already staged an unoptimized blob. Verified end-to-end locally.

**The three RUSTSEC advisories are unrelated to this branch and pre-date it**: `bitmaps`, `im` and `sized-chunks`, declared unmaintained together, all reached through one edge (`simlin-serve → loro → loro-internal → im`), each with no safe upgrade — `im`'s own announcement points at the `imbl` fork, which is loro's migration to make. Ignored with rationale and a revisit condition, matching the three entries already present for this class; they go together because two are only present as the third's dependencies.

Full pre-commit gate green.

@codex review

bpowers:
All five addressed in `9d626314`. Four were correct as reported; the two digest findings were correct **and understated**, and one of the build findings was worse than described in a way the report did not reach.

## Digest coverage — confirmed, and the gap was doing real damage

`Results::offsets` is `HashMap<Ident<Canonical>, usize>`, one entry per variable, extent discarded when `CompiledSimulation` is built — exactly as you describe. Measured on C-LEARN: **7,153 LTM variables, 20,892 slots, 1,088 of them arrayed.** The digest was reading 7,000.

Widening it to every element:

| | before | after |
|---|---:|---:|
| slots covered | 7,000 | **20,892** |
| `nonzero_slots` | 1,369 | **3,141** |
| positive control: real scores zeroed | 82 | **614** |

So 1,772 slots carrying real scores were invisible, and the gate was detecting **an eighth** of the damage its own positive control inflicts. It also now pins a count of LTM-prefixed result slots whose extent the metadata does not describe at `0`, so a population it cannot see fails here rather than being absorbed silently.

**The consequence for the PR is a strengthening, not a retraction**: value-neutrality re-run at the new coverage is identical in every pinned number with the omission disabled. The previous result could only speak for the 7,000-slot sample; the central claim is now supported roughly three times as well.

## Permutation-invariance — confirmed, and demonstrated rather than asserted

Fixed by adding an FNV-1a over the canonically ordered `(name, element, mantissa, exponent)` stream. The plain sums are kept because they are interpretable — a drop in one localises a regression faster than a hash does.

`permuting_two_slots_moves_only_the_identity_digest` constructs the swap and pins **both halves**: that the sums really are unchanged (so the blindness being fixed is not a strawman) and that the identity digest really moves. It covers a swap between two variables and a swap between two elements of a single arrayed variable — the likelier remapping bug.

## The stale mode stamp — correct, and the guard shared its failure mode

Reproduced with a `wasm-opt` shim that is present but exits 1:

| step | blob | `.raw` | mode | |
|---|---|---|---|---|
| 1. good optimized build | 6829333 | 8327392 | `opt` | optimized |
| 2. Rust changes, wasm-opt dies after the copies | 8327384 | 8327384 | `opt` | **raw** |
| 3. next run, real wasm-opt | 8327384 | 8327384 | `opt` | **early-out, exit 0** |

Step 3 exits **0** with the raw output staged and stamped `opt` — a green build with a wrong artifact, which is worse than the bug the stamp was added for. Fixed by clearing the stamp before restaging, so the window is self-invalidating.

**Beyond the report: `REQUIRE_WASM_OPT` read that same stamp**, so the guard would have passed on precisely the artifact it exists to catch. It now also requires the blob to differ from its `.raw` — direct evidence the pass transformed something rather than a claim that it did. Mutation-tested against the forged state.

The `.raw` write has no equivalent window, and the reason is now recorded rather than left to be re-derived: the blob is copied *before* `.raw`, so a failure between them leaves `.raw` describing the previous output and the next `cmp` rebuilds.

## Installer in the path filters — correct

Added to both lists. Every other input was checked against "can a change to this file break something only *this* lane would catch", and nothing else qualified — `package.json` and the lockfiles are excluded deliberately because `ci.yaml`'s frontend job has no path filter and reds first.

Also added a lint rule (pre-commit phase 1) failing if any workflow's `push` and `pull_request` `paths` lists diverge. Hand-maintained twin lists that read as one filter are where "runs on merge but not on the PR" hides.

## Remaining fixed-arity docs — correct, and there were three

`EmitCtx::apply_locals` and the `emit_apply` rustdoc as you named, plus a third in **`bytecode.rs`** — `// Builtins always take 3 args (actual + padding)` sitting immediately above the line that replaced it, introduced by the same commit that added the correct one. A grep of `wasmgen/` would not have found it; the claim had propagated into the file that defines the opcode. Swept for the class afterwards: zero remaining. Comments in `vector.rs` calling the scratch locals "free" were deliberately left — they claim availability to clobber, not contents, and each emitter writes before reading.

Full pre-commit gate green.

@codex review

bpowers:
Correction to my previous comment, plus two numbers I got wrong. Pushed as `3029f138`.

**The permutation test I described did not test what I said it tested.** I wrote that it "constructs the swap and pins both halves — that the plain sums are blind to it, and that the identity digest is not". The first half was true; the second was not tested. Mutating the identity bytes out of the fold left the test **green**, because FNV-1a is an *ordered* fold: exchanging two slots' maxima reorders the value stream and moves the hash whether or not `name` and `elem` are in it. It demonstrated order-sensitivity and claimed identity binding.

Now split into two rows constraining different things:

| row | constrains | mutation that reds it |
|---|---|---|
| value swap — maxima exchanged, order fixed | the fold is ordered | (correctly green under identity removal) |
| **rebinding** — same maxima, same order, different name or element index | the identity bytes | removing `name`/`elem` — **reds** |

The rebinding row is also the closer analogue of the regression at issue: an offset or remapping change attaches correct values to the wrong links, which leaves both the multiset *and* its order intact — so the swap case would not have caught it either.

The C-LEARN pin itself was never wrong: its real caller sorts by `(name, element)`, so a swap does move it, and all seven numbers are unchanged. What was wrong was the argument that it would.

**Two numbers corrected.** The positive control zeroes **614** real scores at full coverage, not 82 — 82 was the pre-widening figure and I quoted it in the "before" column while describing post-widening behaviour. And my estimate that C-LEARN had ~24,908 LTM slots was built on a wrong subtrahend: I used 5,215 as the non-LTM slot count when it is **9,132** (5,215 is a variable count, not a slot count). The measured accounting closes exactly — 30,024 slots/step, 21,011 described by LTM metadata, 119 never saved, **20,892 covered**, `unknown_extent = 0`.

Worth stating plainly, since it is the third time this gate has been wrong about itself — quantization, then coverage and permutation-invariance, now the test asserting a property it did not exercise. The first two were in the artifact; this one was in the evidence, which is the harder kind to notice and the reason the property is now split across two rows rather than argued.

Full pre-commit gate green. 75 commits.

@codex review

bpowers:
The PyYAML finding is correct on both halves and is fixed. Reproduced before fixing, by shadowing the module: traceback to stderr, then `Project lint check passed`, exit 0.

It worked here only because PyYAML is a system package on this machine — which is the hazard in one line: **an undeclared import makes a check's coverage a property of the developer's machine.**

`scripts/check-workflow-paths.py` now parses the subset of YAML it needs with the standard library, and is strict in the direction that matters: an inline `paths:` value, an unparsed line, or a mapping key nested inside a paths list all raise, so a workflow it cannot read fails loudly rather than reporting "no difference".

That strictness is tested rather than asserted, and it needed to be — **the first version of the parser silently ignored a stray key inside a `paths` list while its docstring claimed it would raise.** I caught it by testing the claim rather than re-reading it, which is the same failure this whole review sequence has been about, written into the fix for it.

Four properties now verified: a clean tree passes; real drift is caught (dropping one entry from the `pull_request` list alone reds it, naming the entry); a check that cannot run fails loudly; and an unparseable workflow fails rather than passing vacuously.

**Your second point — that the failure branch only reads stdout — was the more valuable half, and it generalised.** Rule 3 had the identical shape: if `check-copyright.py` crashed, its traceback went to stderr, its stdout was empty, and the lint passed. Both rules now go through a `run_line_check` helper that treats a check which *fails to run* as an error in its own right. A rule that has silently stopped running looks exactly like a rule that found nothing.

Not taking the P3 on `docs/dev/deploy.md` in this PR.

Full pre-commit gate green.

PR #1055, opened 2026-09-04
engine: dedupe fragment-refusal variable names

### LINKED ISSUE #1047: engine: assemble_module's 'failed to compile fragments for variables' names a variable once per runlist it is in (module-argument helpers listed twice)
## Problem

`assemble_module` (`src/simlin-engine/src/db/assemble.rs`) builds its batch refusal by calling `program_fragments` three times -- once each for the Initials, Flows and Stocks phases -- with the **same** `&mut missing: Vec<String>`, then joins the vector as-is:

```rust
let mut missing: Vec<String> = Vec::new();
let initials = program_fragments(Phase::Initials, &dep_graph.runlist_initials, ..., &mut missing);
let flows    = program_fragments(Phase::Flows,    &dep_graph.runlist_flows,    ..., &mut missing);
let stocks   = program_fragments(Phase::Stocks,   &dep_graph.runlist_stocks,   ..., &mut missing);
if !missing.is_empty() {
    return Err(format!(
        "failed to compile fragments for variables: {}",
        missing.join(", ")
    ));
}
```

`program_fragments` pushes a runlist member's name whenever it has no bytecode for the phase being emitted (`None => missing.push(name.clone())`). A variable whose lowering was refused has no bytecode for **any** phase, so it is pushed once per runlist it is a member of. The message therefore names a variable as many times as it has runlist memberships, not once, and reads as if there were twice as many failing variables as there are.

Which variables that hits (from the runlist rules in `db/dep_graph.rs`):

- A hoisted-argument helper feeding a stdlib module instance (`SMTH1`/`DELAY`/... arguments, `$⁚<parent>⁚<n>⁚arg<k>`): it is in the initials runlist (a module instance's inputs are among its initial dependencies) **and** the flows runlist -- listed twice.
- Any ordinary aux that is both an initial dependency (of a stock, a module instance, or an `INIT` referent) and a per-step flow -- twice.
- A stock whose fragment was refused: initials + stocks runlists -- twice.
- A `PREVIOUS`-only capture helper is flows-only and an `INIT`-only capture helper is initials-only, so those are listed once. That is why the doubling looks intermittent from the outside: it tracks runlist membership, not the kind of failure.

## Reproduction

`probe_smth1_args.xmile` -- two `SMTH1` calls in one apply-to-all equation whose hoisted argument expressions codegen refuses (a computed array operand to `allocate_available`, the same refusal `diagnostic_tests::codegen_rejection_of_an_ordinary_variable_names_the_variable_and_its_reason` pins):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<xmile version="1.0" xmlns="http://docs.oasis-open.org/xmile/ns/XMILE/v1.0">
  <header><name>probe_smth1_args</name><vendor>probe</vendor><product version="1.0">probe</product></header>
  <sim_specs><start>0</start><stop>2</stop><dt>1</dt></sim_specs>
  <dimensions>
    <dim name="d"><elem name="e1"/><elem name="e2"/><elem name="e3"/></dim>
    <dim name="xp" size="4"/>
  </dimensions>
  <model>
    <variables>
      <aux name="request"><dimensions><dim name="d"/></dimensions><eqn>10</eqn></aux>
      <aux name="pp"><dimensions><dim name="d"/><dim name="xp"/></dimensions><eqn>1</eqn></aux>
      <aux name="pp_bump"><dimensions><dim name="d"/><dim name="xp"/></dimensions><eqn>0</eqn></aux>
      <aux name="supply"><eqn>35</eqn></aux>
      <aux name="out">
        <dimensions><dim name="d"/></dimensions>
        <eqn>SMTH1(allocate_available(request[d], pp[d,1] + pp_bump[d,1], supply), 3) + SMTH1(allocate_available(request[d], pp[d,1] + pp_bump[d,1] + 1, supply), 3)</eqn>
      </aux>
    </variables>
  </model>
</xmile>
```

`simlin simulate probe_smth1_args.xmile` prints six attributed `assembly error in model 'main': implicit variable '$⁚out⁚N⁚arg0⁚eK' ... failed to compile: codegen rejected ...` rows (one per helper per element -- correct), then:

```
error compiling model 'main': SimulationError{not_simulatable: failed to compile fragments for variables: $⁚out⁚0⁚arg0⁚e1, $⁚out⁚0⁚arg0⁚e2, $⁚out⁚0⁚arg0⁚e3, $⁚out⁚1⁚arg0⁚e1, $⁚out⁚1⁚arg0⁚e2, $⁚out⁚1⁚arg0⁚e3, $⁚out⁚0⁚arg0⁚e1, $⁚out⁚0⁚arg0⁚e2, $⁚out⁚0⁚arg0⁚e3, $⁚out⁚1⁚arg0⁚e1, $⁚out⁚1⁚arg0⁚e2, $⁚out⁚1⁚arg0⁚e3}
```

Six failing variables, twelve names: the initials-phase pass and the flows-phase pass each contributed the full set.

Control: the same two argument expressions under `PREVIOUS(..., 0) + INIT(...)` instead of `SMTH1(..., 3)` give `failed to compile fragments for variables: $⁚out⁚1⁚arg0, $⁚out⁚0⁚arg0` -- once each, because a `PREVIOUS` capture is flows-only and an `INIT` capture is initials-only.

## Why it matters

- This line is the one thing a user reads when the build fails, and the list is what they scan to find what to fix. Doubling it misstates how many variables are broken, and on C-LEARN-sized failures (#580 quoted ~150 helper names) it doubles an already very long line.
- Tests that cite the message (`fragment_determinism_tests.rs`, `array_tests.rs`, `diagnostic_tests.rs`) key on membership, so they are unaffected; but anything that counts the names, or a future test pinning the message text, would be pinning runlist membership by accident.

## Fix

Small and local: deduplicate `missing` before joining, preserving first-seen order (an `IndexSet`, or a `seen` set at the push site) -- first-seen order keeps the runlist's topological order, which is the more useful reading than a sort. Alternatively make `missing` an ordered set at the call site. One test with a refused module-argument helper (the probe above, or its `TestProject` equivalent) asserting each name appears exactly once in the message pins it; a second row with the `PREVIOUS`/`INIT` control shape lets the test state which runlist-membership arms it covers.

This is a message-shape fix only. It does not touch #581 (structured diagnostics from the assembly chain are discarded), #666 / #1000 (attribution of the helper's own failure -- the per-helper `assembly error` rows above show the codegen path now attributes), or #580 / #1002 (the underlying failures).

## Component

`src/simlin-engine/src/db/assemble.rs` -- `assemble_module`, `program_fragments`.

## Discovery context

Identified during the Phase 9 adversarial review of the compiler-unification-v2 branch (PR #1040) on the probe model `probe_helper_dedup_distinct`; the doubled listing was present on both the pre-branch and branch binaries. The message construction is the same on `main` (`git show main:src/simlin-engine/src/db/assemble.rs`: the one `missing` vector threaded through the three `program_fragments` calls and joined without dedup). The reproduction above was run against the branch's working-tree debug build of `simlin`; the mechanism is the per-phase push, not anything branch-specific.

BetterAndBetterII:
## Summary
`assemble_module` built its batch refusal by calling `program_fragments` three times with the same `&mut missing` vector, then joining it as-is. A refused variable has no bytecode for any phase, so it was pushed once per runlist membership. Module-argument helpers (e.g. SMTH1) sit on both the initials and flows runlists, so six failing names were printed as twelve.

Deduplicate `missing` before joining, preserving first-seen (runlist topological) order.

Fixes #1047

## Test plan
- [x] `cargo test -p simlin-engine --lib fragments_refusal_names_each_variable_once`
- [ ] Confirm SMTH1 argument helpers list 6 unique names (not 12)
- [ ] Confirm PREVIOUS/INIT control still lists each helper once

### DIFF
--- a/src/simlin-engine/src/db/assemble.rs
+++ b/src/simlin-engine/src/db/assemble.rs
@@ -1364,9 +1364,19 @@ pub fn assemble_module<'db>(
         &mut missing,
     );
     if !missing.is_empty() {
+        // A refused variable has no bytecode for any phase, so the three
+        // `program_fragments` passes would otherwise list it once per runlist
+        // it is a member of (a SMTH1 argument helper: initials + flows).
+        // First-seen order is the runlist topological order (GH #1047).
+        let mut seen = HashSet::new();
+        let unique: Vec<&str> = missing
+            .iter()
+            .filter(|name| seen.insert(name.as_str()))
+            .map(|name| name.as_str())
+            .collect();
         return Err(format!(
             "failed to compile fragments for variables: {}",
-            missing.join(", ")
+            unique.join(", ")
         ));
     }
 
--- a/src/simlin-engine/src/db/assemble_tests.rs
+++ b/src/simlin-engine/src/db/assemble_tests.rs
@@ -743,3 +743,112 @@ fn each_program_emits_in_runlist_order_then_the_ltm_tail() {
         "stocks: runlist order, then the LTM implicit tail"
     );
 }
+
+/// Names listed in `assemble_module`'s batch fragment-refusal message, in the
+/// order they appear -- including duplicates, if any. That is the whole point:
+/// a test that only checks membership would pass the doubled listing GH #1047
+/// is about.
+fn fragment_refusal_names(err: &crate::common::Error) -> Vec<&str> {
+    const PREFIX: &str = "failed to compile fragments for variables: ";
+    let details = err.details.as_deref().unwrap_or_else(|| {
+        panic!("expected NotSimulatable details carrying the fragment list, got {err}")
+    });
+    let rest = details
+        .split_once(PREFIX)
+        .unwrap_or_else(|| panic!("expected `{PREFIX}…` in the compile error, got: {details}"))
+        .1;
+    rest.split(", ").collect()
+}
+
+/// The computed `allocate_available` priority-profile operand codegen refuses
+/// (`Cannot push view for expression type`; pinned by
+/// `diagnostic_tests::codegen_rejection_of_an_ordinary_variable_names_the_variable_and_its_reason`).
+/// Wrapping it in `equation` is how a test chooses which runlists the refused
+/// helper sits on: a stdlib-module argument is initials+flows, a `PREVIOUS`
+/// capture is flows-only, an `INIT` capture is initials-only.
+fn refused_allocate_available(equation: &str) -> crate::common::Error {
+    crate::test_common::TestProject::new("fragment_refusal")
+        .named_dimension("d", &["e1", "e2", "e3"])
+        .indexed_dimension("xp", 4)
+        .array_const("request[d]", 10.0)
+        .array_const("pp[d,xp]", 1.0)
+        .array_const("pp_bump[d,xp]", 0.0)
+        .scalar_const("supply", 35.0)
+        .array_aux("out[d]", equation)
+        .compile_incremental()
+        .expect_err("the computed allocate_available profile is refused by codegen")
+}
+
+fn assert_each_fragment_name_once(names: &[&str], what: &str) {
+    assert!(
+        !names.is_empty(),
+        "{what}: expected a non-empty fragment-refusal list"
+    );
+    let unique: HashSet<&str> = names.iter().copied().collect();
+    assert_eq!(
+        names.len(),
+        unique.len(),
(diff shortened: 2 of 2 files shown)


bpowers:
thanks for the PR! I ended up solving this a similar but slightly different way

